PR #5, opened 2019-11-15
[IN PROGRESS] Added parsing and indexing markdowns to algolia index

drobnikj:
I need to set up right env vars on travis and added npm command to travis config after we released new doc.


drobnikj:
I parse documentation from HTML in the end. 
I will change this code to just parse markdowns to check if all files are valid.

PR #17, opened 2020-03-16
Apify docs to use Github Actions + moving here the API blueprint specification

mtrunkat:
I replaced Travis with Github Actions so that the log and everything else is visible here directly and you don't have to go to Travis and also to beware of problem where this is executed twice on Travis.com and Travis.org (fails).

Then I moved here an API blueprint docs and set up a deployment.


mtrunkat:
Master to production
Everything else currently to staging

PR #52, opened 2020-04-15
Feature/support renamed redirects

m-murasovs:
Added the 'paths' and 'menu' properties to .md files, extended the checks in parser.js to make sure 'paths' are present and the current one is included.


src/parser.js
```diff
@@ -19,12 +19,18 @@ const readAndParsePage = async (fullPath, shortPath) => {
...
+    if (!_.includes(metadata.paths, filenamePath)) throw new Error(`Value metadata.paths in ${fullPath} is missing the path "${filenamePath}"`);
 
     return Object.assign({
         content,
         contentHash: crypto.createHash('sha256').update(content).digest('base64'),
-        menuTitle: metadata.title,
-        path: shortPath.replace('/index.md', '').replace('.md', '').replace(/_/g, '-'),
+        menuTitle: metadata.menuTitle,
+        path: filenamePath,
+        redirectPaths: metadata.paths,
```
mtrunkat, line 33: I would exclude at this point `filenamePath` from `redirectPaths` to make sure that we won't create a redirect loop at the website.

src/parser.js
```diff
@@ -19,12 +19,18 @@ const readAndParsePage = async (fullPath, shortPath) => {
...
+    // Check if the path based on filename is in the metadata.paths array
+    const filenamePath = shortPath.replace('/index.md', '').replace('.md', '').replace(/_/g, '-');
+    if (!_.includes(metadata.paths, filenamePath)) throw new Error(`Value metadata.paths in ${fullPath} is missing the path "${filenamePath}"`);
 
     return Object.assign({
         content,
         contentHash: crypto.createHash('sha256').update(content).digest('base64'),
-        menuTitle: metadata.title,
-        path: shortPath.replace('/index.md', '').replace('.md', '').replace(/_/g, '-'),
+        menuTitle: metadata.menuTitle,
```
drobnikj, line 31: I would do it in way: If menuTitle in metadata exists then use it as menuTitle otherwise use title. It is unessesary to have menuTitle and title in all metadata same.

src/parser.js
drobnikj, line 33: You are creating a new field called redirectPaths in object here, but after assign whole metadata to this object you get duplicate values with different key see paths/redirectPaths:
```
"api.md": {
      "contentHash": "cR3wBRBzOo75+vrnaioP1nk9qPvDMVTXD809r6s36BI=",
      "menuTitle": "API",
      "path": "api",
      "redirectPaths": [
        "apo,",
        "api"
      ],
      "sourceUrl": "https://apify-docs.s3.amazonaws.com/master/pages/api.md",
      "title": "API",
      "description": "Documentation of Apify REST API that enables integration with external applications and systems.",
      "menuWeight": 10,
      "paths": [
        "apo,",
        "api"
      ]
    },
```

src/parser.js
jancurn: Yeah, that makes sense

jancurn:
Looks good! Just one question - so do I understand correctly that we can enter to `paths` even a URL with #fragements, which will be passed to the JSON used by the web server, which can then handle the redirect on client-side?

PR #55, opened 2020-05-05
Revamp/rename actor to actors

m-murasovs:
First PR for revamp, just updating 'actor' to 'actors'.
Also demonstrates that the redirects work - yay :)


vbartonicek:
Just make sure, that original URLs containing "actor" will be properly redirected ;-)

jancurn:
Good point @vbartonicek. Actually, in this case it would be worth adding an integration test for that...

PR #63, opened 2020-05-13
Revamp/remove console from run

m-murasovs:


m-murasovs:
Quick fix just to get us by until it's rewritten

Issue #21, opened 2020-03-22
Move Apiary API docs and scripts into a separate directory


jancurn:
It's quite a mess to have it in the top-level directory. Also, add this README to the new dir:

```
This directory contains a source file for Apiary
API reference (https://docs.apify.com/api).
It is automatically uploaded to Apiary during web deployment process.

## Local testing

1. Install Apiary gem `gem install apiaryio`
2. After that you can can open generated doc with command: `apiary preview --path="./api-v2-reference.apib"`

### Test

You should ALWAYS log in to Apiary, analyze the document and make sure there are no warnings!
```

m-murasovs:
Done in PR #81 
https://github.com/apify/apify-docs/pull/81

PR #81, opened 2020-08-07
Fix/api docs GitHub issue

m-murasovs:
Connected to issue https://github.com/apify/apify-docs/issues/21

Moved the upload scripts to the SRC folder and adjusted .github/workflows/build.yaml
Tested the new build script - Apiary gets updated

Added a bit to the RADME to show how API docs work


mtrunkat:
I am not really sure about this this move

```
upload_to_apiary.sh → docs/api_v2/upload_to_apiary.sh
```

as all bash scripts are located in root dir and `/docs` dir contains only the documentation source files and no scripts. The script that deploys API docs to Apiary is kind of equivalent to the one that deploys the rest of the docs to S3 - see https://take.ms/1UZSD . Perhaps we could move both these files to `./src` dir that contains all the build scripts but it should not a part of docs content in `./docs`. What do you think?

Issue #79, opened 2020-08-07
Update the README 


m-murasovs:
Include info about API docs/blueprint/external docs like actor-scraper

Mention that images should be run through tinypng.com for compression

PR #88, opened 2020-08-19
Update the readme

m-murasovs:
Issue: https://github.com/apify/apify-docs/issues/79

I went a bit extra, but basically I wanted to make sure all this stuff isn't just in my head but also available for others (in case I get kidnapped by aliens or something)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,17 @@ Before making changes, make sure you're familiar with the [Apify Writing Style G
 
 ### Highlighting
 
-For consistency, use **bold** for highlighting non-code words/phrases. For inline `code` examples, use **backticks** (\` \`).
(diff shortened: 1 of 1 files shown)


mtrunkat (approved):

mtrunkat:
Cool! Only one thing - Notion is not public and so the Writing Style Guide linked from this repo is not publicly available. Perhaps we could make it part of the docs if it's not possible to make that single page public.

m-murasovs:
> Cool! Only one thing - Notion is not public and so the Writing Style Guide linked from this repo is not publicly available. Perhaps we could make it part of the docs if it's not possible to make that single page public.

Aah, I forgot! I think I'll just take that bit out, as it doesn't seem like the style guide is really docs material

PR #90, opened 2020-08-19
Revamp/scraping101 bypassing protection

m-murasovs:
Kicking off the Web scraping 101 section.

Naming will soon be updated so we don't have the clashing 'Scraping' and 'Web scraping 101' folders. I'm going to rename the 'Scraping' folder to 'Apify scrapers'


metalwarrior665:
I think the article is really good, not too simple and not too complicated. I'm still working on my antiscraping "masterplan" but that won't reach public state for quite some time (and even doesn't have to).

metalwarrior665 (approved):

mtrunkat (approved):

PR #104, opened 2020-09-25
Revamp/proxy section

m-murasovs:
I tried to avoid one monster PR by doing several smaller increment branches but it wasn't working out. Apologies for the quantity :) 

Currently available on develop branch: https://docs.apify.com/proxy?version=develop (if something else is there, just re-run the deployment job under the Actions tab)

@mnmkng if you could check SDK-related stuff, that would be great :) 

@davidjohnbarton if you could take a look, that would be great :)


docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,668 @@
...
+### [](#single-request-with-a-random-ip-address) Single request with a random IP address
+
+The IP address is chosen from all available proxy groups.
+
+Select this option by setting the **username** parameter to **auto**.
+
+```marked-tabs
+<marked-tab header="Node.js (axios)" lang="javascript">
+const HttpsProxyAgent = require("https-proxy-agent");
+axios = require("axios");
```
mnmkng, line 40: missing `const`

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,668 @@
...
+
+const httpsAgent = new HttpsProxyAgent({
+    host: "proxy.apify.com",
+    port: "8000",
+    // Replace <YOUR_PROXY_PASSWORD> below with your password
+    // found at https://my.apify.com/proxy
+    auth: "auto:<YOUR_PROXY_PASSWORD>"
+});
+
+axios = axios.create({httpsAgent});
```
mnmkng, line 50: Not a good idea to reassign what you required. Better create a new variable.

mnmkng: Btw, it seems that the example is not linted with the `@apify/eslint-config` rules. `({httpsAgent})` => `({ httpsAgent })`

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,668 @@
...
+echo $response1;
+echo "\nShould be contain same clientIp as\n";
+echo $response2;
+?>
+</marked-tab>
+```
+
+### [](#two-requests-with-different-ip-addresses) Two requests with different IP addresses
+
+The IP addresses are chosen from the `SHADER` and `BUYPROXIES94952` proxy groups.
```
mnmkng, line 281: This seems to imply that you need to specify some proxy groups to get different IPs, but that's not the case.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,668 @@
...
+
+## [](#using-the-apify-sdk) Using the Apify SDK
+
+If you're developing an actor using the [Apify SDK](https://sdk.apify.com), you can use Apify proxy in:
+
+* [PuppeteerCrawler](https://sdk.apify.com/docs/api/puppeteer-crawler#docsNav) using the [createProxyConfiguration()](https://sdk.apify.com/docs/api/apify#apifycreateproxyconfigurationproxyconfigurationoptions) function.
+* [requestAsBrowser()](https://sdk.apify.com/docs/api/utils#utilsrequestasbrowseroptions) function by specifying proxy configuration in the options.
+* [launchPuppeteer()](https://sdk.apify.com/docs/typedefs/launch-puppeteer#docsNav) by specifying the configuration in the function's options.
+
+### [](#single-random-ip-address) Single random IP address
```
mnmkng, line 425: This will actually rotate IPs automatically, so the single IP might be a bit misleading. Yeah, it's a single IP for a single request, but that's always the case by definition.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,668 @@
...
+        console.log(response2.body.clientIp);
+    } catch (e) {
+        console.error(e);
+    }
+});
+</marked-tab>
+```
+
+
+### [](#single-ip-from-a-specific-group-using-puppeteercrawler) Single IP from a specific group using PuppeteerCrawler
```
mnmkng, line 563: Here it truly is a single IP until failure.

mnmkng (commented):
I admit I did not read it very thoroughly, but aside from some titles / comments that I think are a bit misleading, the examples are fine. 👍

docs/proxy/datacenter_proxy/examples.md
m-murasovs: Looks like the current docs linter doesn't pick it up.
We have an issue to improve docs linting, so I'll include this there

docs/proxy/datacenter_proxy/examples.md
m-murasovs: In the first example, does each opened browser keep the same IP address until the IP fails?

mnmkng: Which example is "the first example"?

mnmkng: Oh, wait, I scrolled up to the nearest header, but now I see that there are two different approaches.

In the first example, it will use a random IP address for each new browser. Browsers are typically killed after 100 requests. I would not mention it in the docs, that's an SDK thing. Only mentioning it for context.

In the second example, because of the `maxPoolSize: 1` config, a single IP will be used by all browsers until it fails. Then, all running browsers will get retired and a new IP will be selected, new browsers spawned and they will all use that new IP.

mnmkng:
This is becoming too huge and GitHub is really not the best place to read docs in markdown with codeblocks in marked-something.

Maybe we could deploy it and then iterate on the deployed changes? It would be so much easier with code-tabs and highlighting to read. 🤔

docs/proxy/datacenter_proxy.md
```diff
@@ -6,107 +6,89 @@ paths:
...
-            by default, the proxy uses all available
-            <br/>proxy servers from all countries.
-        </td>
-    </tr>
-    </tbody>
-</table>
+[Click here]({{@link web_scraping_101/anti_scraping_techniques.md#bypassing-ip-address-based-blocking}}) to learn more about IP address rotation and other ways of bypassing blocking.
 
-If you do not want to specify both `groups` and `session` parameters and therefore use **default** behavior for both, use the following username:
+## [](#session-persistence) Session persistence
```
mnmkng, line 82: What about changing this section from session persistence to `IP rotation` and rewriting it a bit? The idea is that a lot of people have no idea what a `session` is, whereas IP is a generally known term.

I remember, 3 months into my first engineering job, where I was building a Node.js server framework, I had no idea what a session was. 🤦  So I think it's fair to assume 99% of people are clueless about sessions.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+    port: "8000",
+    // Replace <YOUR_PROXY_PASSWORD> below with your password
+    // found at https://my.apify.com/proxy
+    auth: "auto:<YOUR_PROXY_PASSWORD>"
+});
+
+axiosHttpsAgent = axios.create({ httpsAgent });
+
+async function useProxy() {
+    const response = await axiosHttpsAgent.get("https://api.apify.com/v2/browser-info");
```
mnmkng, line 54: Maybe use http://proxy.apify.com/ instead of `browser-info` endpoint?

mnmkng, line 54: Node 14 has support for top level `await`, so let's simplify and skip the `useProxy` function.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+    console.log(response.data)
+};
+useProxy();
+</marked-tab>
+
+
+<marked-tab header="Node.js (got)" lang="javascript">
+const got = require("got");
+const HttpProxyAgent = require("http-proxy-agent");
+const HttpsProxyAgent = require("https-proxy-agent");
```
mnmkng, line 64: With `axios` we're only using the `HttpsProxyAgent`. Why is this different?

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+
+const httpsAgent = new HttpsProxyAgent({
+    host: "proxy.apify.com",
+    port: "8000",
+    // Replace <YOUR_PROXY_PASSWORD> below with your password
+    // found at https://my.apify.com/proxy
+    auth: "auto:<YOUR_PROXY_PASSWORD>"
+});
+
+axiosHttpsAgent = axios.create({ httpsAgent });
```
mnmkng, line 51: I think a better name would be `const axiosWithProxy = `

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+<marked-tab header="Node.js (got)" lang="javascript">
+const got = require("got");
+const HttpProxyAgent = require("http-proxy-agent");
+const HttpsProxyAgent = require("https-proxy-agent");
+
+// Replace <YOUR_PROXY_PASSWORD> below with your password
+// found at https://my.apify.com/proxy
+const proxyUrl = "http://auto:<YOUR_PROXY_PASSWORD>@proxy.apify.com:8000"
+
+async function useProxy() {
```
mnmkng, line 70: same here

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+// found at https://my.apify.com/proxy
+curl_setopt($curl, CURLOPT_PROXYUSERPWD, "auto:<YOUR_PROXY_PASSWORD>");
+$response = curl_exec($curl);
+curl_close($curl);
+if ($response) echo $response;
+?>
+</marked-tab>
+```
+
+### [](#two-requests-with-the-same-ip-address) Two requests with the same IP address
```
mnmkng, line 142: Multiple, instead of Two?

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+echo $response1;
+echo "\nShould be contain same clientIp as\n";
+echo $response2;
+?>
+</marked-tab>
+```
+
+### [](#two-requests-with-different-ip-addresses) Two requests with different IP addresses
+
+The IP addresses are chosen from the `SHADER` and `BUYPROXIES94952` proxy groups, however you don't have to specify proxy groups to use different IPs.
```
mnmkng, line 282: I still don't get this too much. Why conflate proxy groups and IP rotation into the same example? Maybe that's only my opinion, but I always thought that examples should provide the minimal workable solution to the problem at hand, without any extra features or unneeded clutter. If we want to show off proxy groups, we should add proxy group examples.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+echo $response1;
+echo "\nShould have different clientIp than\n";
+echo $response2;
+?>
+</marked-tab>
+```
+
+## [](#using-the-apify-sdk) Using the Apify SDK
+
+If you're developing an actor using the [Apify SDK](https://sdk.apify.com), you can use Apify Proxy in:
```
mnmkng, line 420: I think the Apify SDK examples should be top level examples.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+    const proxyConfiguration = await Apify.createProxyConfiguration();
+
+    const crawler = new Apify.PuppeteerCrawler({
+        requestList,
+        proxyConfiguration,
+        handlePageFunction: async ({ page, request, proxyInfo }) => {
+            return Apify.pushData({ title: await page.title() });
+        },
+        handleFailedRequestFunction: ({ request }) => {
+            console.error("Request failed", request.url, request.errorMessages);
```
mnmkng, line 449: The `handleFailedRequestFunction` is an optional function so better remove it from the examples.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+    const proxyConfiguration = await Apify.createProxyConfiguration();
+    try {
+        const { body } = await Apify.utils.requestAsBrowser({
+            url: "https://www.example.com",
+            proxyUrl: proxyConfiguration.newUrl(),
+        });
+        console.log(body); // returns HTML of returned page
+    } catch (e) {
+        console.error(e);
+    }
```
mnmkng, line 471: I don't think we need the error handling in the example.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+    );
+    const proxyConfiguration = await Apify.createProxyConfiguration({
+        // if you need to use more then one group
+        // simply add the additional ID to the array below
+        groups: ["SHADER", "BUYPROXIES94952"],
+    });
+
+    const crawler = new Apify.PuppeteerCrawler({
+        requestList,
+        proxyConfiguration,
```
mnmkng, line 496: This config will give you a random IP for each page, not a single IP, because it does not use sessions anywhere.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+            json: true
+        });
+        const response2 = await Apify.utils.requestAsBrowser({
+            url,
+            proxyUrl,
+            json: true
+        });
+        console.log(response1.body.clientIp);
+        console.log("should be different than");
+        console.log(response2.body.clientIp);
```
mnmkng, line 557: The title is Single IP, but here it says "should be different" (and it will)

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+    const proxyConfiguration = await Apify.createProxyConfiguration({
+        // if you need to use more then one group
+        // simply add the additional ID to the array below
+        groups: ["SHADER"],
+    });
+
+    const crawler = new Apify.PuppeteerCrawler({
+        requestList,
+        proxyConfiguration,
+        useSessionPool: true,
```
mnmkng, line 589: This will use Random IPs, not a single IP as the title suggests. I see there's an explanation in the paragraph below the title, but I still find it confusing. Single IP to me means static IP. The same IP for all requests.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+    });
+
+    await crawler.run();
+});
+</marked-tab>
+``` 
+
+Using [PuppeteerCrawler](https://sdk.apify.com/docs/api/puppeteer-crawler#docsNav), keep a single IP address selected from the `SHADER` proxy group until it fails (gets retired).
+
+The `maxPoolSize: 1` configuration means that a single IP will be used by all browsers until it fails. Then, all running browsers are retired, a new IP is selected and new browsers opened. The browsers all use the new IP.
```
mnmkng, line 605: I think this example and the example above are examples of two completely different proxy rotations and should not be nested under the same "Single IP" title

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,700 @@
...
+            console.error("Request failed", request.url, request.errorMessages);
+        },
+    });
+
+    await crawler.run();
+});
+</marked-tab>
+```
+
+### [](#use-a-single-ip-address-from-specific-groups-for-multiple-requests) Use a single IP address from specific groups for multiple requests
```
mnmkng, line 641: I think even the title says that this is becoming too complex.

Let's separate the Group examples and Proxy rotation examples. The groups are rather simple, so perhaps only 2 examples are needed, groups and no groups (auto). The IP rotation examples can then just use auto and not get polluted with groups.

docs/proxy/google_serp_proxy.md
```diff
@@ -1,47 +1,58 @@
...
 
-Search results from Germany: `http://www.google.de/search?q=<query>`
+* Shopping results from Great Britain: `http://www.google.co.uk/seach?tbm=shop&q=<query>`
 
-See [full list of available domain names](https://ipfs.io/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/List_of_Google_domains.html) for specific countries. When used, always remember to prepend the domain name with `www.` prefix.
+See a [full list](https://ipfs.io/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/List_of_Google_domains.html) of available domain names for specific countries. When using them, remember to prepend the domain name with the `www.` prefix.
 
+## [](#using-with-puppeteer) Using with Puppeteer
 
+Google SERP proxy cannot be used with [Puppeteer](https://pptr.dev/), due to HTTP Strict Transport Security [headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security).
```
mnmkng, line 58: Ummm, even I don't know what this means 😄 Does that mean that Chromium/Chrome would enforce a `https` connection to Google which is not supported by the proxy?

docs/proxy/google_serp_proxy/examples.md
```diff
@@ -0,0 +1,512 @@
...
+    auth: "groups-GOOGLE_SERP:<YOUR_PROXY_PASSWORD>"
+});
+
+axiosHttpsAgent = axios.create({ httpsAgent });
+
+// Encode your query as a URI parameter
+const query = `q=${encodeURIComponent('wikipedia')}`;
+
+async function useProxy() {
+    const response = await axiosHttpsAgent.get(`http://www.google.com/search?${query}`);
```
mnmkng, line 57: I would use `axios`'s `params` option instead of encoding the query manually and then constructing the URL from strings.

docs/proxy/google_serp_proxy/examples.md
```diff
@@ -0,0 +1,512 @@
...
+
+const httpsAgent = new HttpsProxyAgent({
+    host: "proxy.apify.com",
+    port: "8000",
+    // Replace <YOUR_PROXY_PASSWORD> below with your password
+    // found at https://my.apify.com/proxy
+    auth: "groups-GOOGLE_SERP:<YOUR_PROXY_PASSWORD>"
+});
+
+axiosHttpsAgent = axios.create({ httpsAgent });
```
mnmkng, line 51: just a reminder of `const axiosWithProxy =`

docs/proxy/google_serp_proxy/examples.md
mnmkng, line 57: A reminder to remove the `useProxy` func

docs/proxy/google_serp_proxy/examples.md
```diff
@@ -0,0 +1,512 @@
...
+
+// Replace <YOUR_PROXY_PASSWORD> below with your password
+// found at https://my.apify.com/proxy
+const proxyUrl = "http://groups-GOOGLE_SERP:<YOUR_PROXY_PASSWORD>@proxy.apify.com:8000"
+
+// Encode your query as a URI parameter
+const query = `q=${encodeURIComponent('wikipedia')}`;
+
+async function useProxy() {
+    const response = await got(`http://www.google.com/search?${query}`, {
```
mnmkng, line 77: `got` also has some `params` or `query` option. We should use it.

docs/proxy/google_serp_proxy/examples.md
```diff
@@ -0,0 +1,512 @@
...
+    port: "8000",
+    // Replace <YOUR_PROXY_PASSWORD> below with your password
+    // found at https://my.apify.com/proxy
+    auth: "groups-GOOGLE_SERP:<YOUR_PROXY_PASSWORD>"
+});
+
+axiosHttpsAgent = axios.create({ httpsAgent });
+
+// Encode your query as a URI parameter
+const query = `q=${encodeURIComponent('wikipedia')}tbm=${'shop'}`;
```
mnmkng, line 174: use the `params` for `tbm` too

docs/proxy/google_serp_proxy/examples.md
```diff
@@ -0,0 +1,512 @@
...
+echo $response;
+?>
+</marked-tab>
+```
+
+## [](#using-the-apify-sdk) Using the Apify SDK
+
+If you're developing an actor using the [Apify SDK](https://sdk.apify.com), you can use Apify Proxy in:
+
+* [PuppeteerCrawler](https://sdk.apify.com/docs/api/puppeteer-crawler#docsNav) using the [createProxyConfiguration()](https://sdk.apify.com/docs/api/apify#apifycreateproxyconfigurationproxyconfigurationoptions) function.
```
mnmkng, line 272: Wait, so does it work or not? Above it says that SERP does not work with Puppeteer.

docs/proxy/google_serp_proxy/examples.md
```diff
@@ -0,0 +1,512 @@
...
+        handlePageFunction: async({ page, request, proxyInfo }) => {
+            await Apify.utils.puppeteer.injectJQuery(page);
+            const searchResults = await page.evaluate(() => {
+                return $('#search div.g').map(function () {
+                    return {
+                        title: $($(this).find('h3')[0]).text(),
+                        description: $(this).find('.s div .st').text(),
+                    }
+                }).toArray();
+            });
```
mnmkng, line 312: I would get rid of the parsing entirely. Google changes selectors quite often, plus IMO it adds unneeded complexity to the example.

docs/proxy/google_serp_proxy/examples.md
mnmkng, line 272: `CheerioCrawler` is much better suited for working with Google SERP. I'd replace the example with that. It does not make any sense to use Puppeteer with SERP, because SERP only returns proxied HTML anyway, it does not actually connect you to Google.

docs/proxy/google_serp_proxy/examples.md
```diff
@@ -0,0 +1,512 @@
...
+            console.log(searchResults);
+        },
+        handleFailedRequestFunction: async({ request }) => {
+            console.error(request.errorMessages);
```
mnmkng, line 316: reminder to not use those functions

docs/proxy/google_serp_proxy/examples.md
```diff
@@ -0,0 +1,512 @@
...
+        },
+    });
+
+    await crawler.run();
+});
+</marked-tab>
+
+
+<marked-tab header="launchPuppeteer()" lang="javascript">
+const Apify = require('apify');
```
mnmkng, line 326: Again, probably not worth the example. Too complex, too slow and for no gain.

docs/proxy/google_serp_proxy/examples.md
```diff
@@ -0,0 +1,512 @@
...
+     const { body } = await Apify.utils.requestAsBrowser({
+         url: 'http://www.google.com/search?q=wikipedia',
+         proxyUrl,
+     });
+     const $ = cheerio.load(body);
+     const searchResults = $('#search div.g').map(function() {
+         return {
+             title: $($(this).find('h3')[0]).text(),
+             description: $(this).find('.s div .st').text(),
+         }
```
mnmkng, line 383: I'd skip the parsing here too.

docs/proxy/google_serp_proxy/examples.md
```diff
@@ -0,0 +1,512 @@
...
+</marked-tab>
+```
+
+### [](#get-and-parse-a-list-of-shopping-results) Get and parse a list of shopping results
+
+Get a list of shopping results for the query **Apple iPhone XS 64GB** from Great Britain (**google.co.uk**) and parse them using [cheerio](https://cheerio.js.org/).
+
+
+```marked-tabs
+<marked-tab header="PuppeteerCrawler" lang="javascript">
```
mnmkng, line 396: Same here, use `CheerioCrawler` instead.

docs/proxy/google_serp_proxy/examples.md
```diff
@@ -0,0 +1,512 @@
...
+            console.error(request.errorMessages);
+        },
+    });
+
+    await crawler.run();
+});
+</marked-tab>
+
+
+<marked-tab header="launchPuppeteer()" lang="javascript">
```
mnmkng, line 441: Delete this.

docs/proxy/google_serp_proxy/examples.md
```diff
@@ -0,0 +1,512 @@
...
+            url: `http://www.google.co.uk/search?tbm=shop&q=${query}`,
+            proxyUrl,
+        });
+        const $ = cheerio.load(body);
+        const searchResults = $('[class$="list-result"] > div > div:nth-child(2) ').map(function() {
+            const title = $(this).find('a[jsaction="spop.c"]')[0];
+            return {
+                title: $(title).text(),
+                price: $(this).find('div:nth-child(2)').text(),
+            }
```
mnmkng, line 503: don't parse

docs/proxy/residential_proxy.md
```diff
@@ -6,66 +6,33 @@ paths:
     - proxy/residential-proxy
 ---
 
-# [](#residential-proxy)Residential proxy
+# [](#residential-proxy) Residential proxy
 
-Sometimes datacenter proxy servers are not a viable option for certain solutions and for these cases Apify Proxy includes an option to use Residential Proxy. This proxy solution allows the user access to a much larger pool of proxy servers than with datacenter proxy servers and therefore it is usually a better option for situations where a large number of proxy servers is required.
+Residential proxies are connected to real Internet Service Provider-based proxies located in homes and offices. Unlike [datacenter proxies]({{@link proxy/datacenter_proxy.md}}), this makes them impossible to block.
```
mnmkng, line 11: I'd say "very hard to block", not impossible.

docs/proxy/residential_proxy.md
```diff
@@ -6,66 +6,33 @@ paths:
...
-    </tr>
-    </tbody>
-</table>
+The `username` field enables you to pass various [parameters]({{@link proxy/connection_settings.md#username-parameters}}), such as groups, session and country, for your proxy connection.
 
-This is how the username would look for the most complex variation: Session set and IP selected from the United States
+When using residential proxies, the username should specify `groups-RESIDENTIAL`.
 
-    groups-RESIDENTIAL,session-my_session_1,country-US
+## [](#session-persistence) Session persistence
```
mnmkng, line 29: reminder of IP rotation vs Session persistence comment

docs/proxy/residential_proxy/examples.md
```diff
@@ -0,0 +1,620 @@
...
+description: Learn how to connect to Apify's residential proxies from your application with Node.js (axios and got), Python 2 and 3 and PHP using code examples.
+paths:
+    - proxy/residential-proxy/nodejs-examples
+    - proxy/residential-proxy/python-examples
+    - proxy/residential-proxy/php-examples
+    - proxy/residential-proxy/examples
+---
+<!-- Watch out! This file is hard to read because our own "marked-tabs" language fences aren't syntax-highlighted -->
+
+# Connect to residential proxies
```
mnmkng, line 12: Do we really need another huge batch of examples for residentials, when they are no different than any other proxy group in terms of connection?

I understood the idea about SERP, where we provided Google specific examples, but this is exactly the same as datacenter stuff.

I think here it should just say that `RESIDENTIAL` is just like any other proxy group and link to `How to set a proxy group` examples and `How to set proxy country` examples.

docs/proxy/residential_proxy/tips_and_tricks.md
```diff
@@ -5,29 +5,29 @@ paths:
...
-Since Residential proxy is priced by data traffic used, it's very easy to realy quickly use up all prepaid traffic, especialy when accessing websites with large files loaded on every page.
+Residential proxy is priced by data traffic used. Thus, it's easy to quickly use up all your prepaid traffic. In particular, when accessing websites with large files loaded on every page.
 
-There are two main ways to reduce the traffic use of your solutions:
+There are two main ways to reduce your traffic use.
 
-*   [Block requests which are not required for your solution](https://help.apify.com/en/articles/2423246-block-requests-in-puppeteer)
-*   [Cache responses that you require but do not wish to load again and again](https://help.apify.com/en/articles/2424032-cache-responses-in-puppeteerr)
+*   [Block requests which are not required for your solution](https://help.apify.com/en/articles/2423246-block-requests-in-puppeteer).
+*   [Cache responses that you require but do not wish to load again and again](https://help.apify.com/en/articles/2424032-cache-responses-in-puppeteerr).
```
mnmkng, line 19: Both of those may or may not be effective at this point and could be bad advice. Both guides use a thing called request interception. The trouble is, request interception disables browser cache in Puppeteer. So in the end, the traffic may actually end up being worse. I think we should do some tests before using those guides. We don't recommend using request interception in the SDK docs and use https://sdk.apify.com/docs/api/puppeteer#puppeteerblockrequests instead.

mnmkng (commented):
I think this is a huge and good piece of work so don't take the number of comments as a sign of a bad job. It's actually a job well done. I basically have 2 high level comments that reflect all the specific comments:

- the examples should be dead simple, no need to add stuff that's not absolutely needed for the task at hand
- the examples could be better structured. Instead of mixing groups and sessions and countries and what not, explain each idea separately.

Also, as a side note. I was not able to load this on "staging" with `?version=develop`. I assume it's because it's not really in the `develop` branch. `?version=revamp/proxy-section` did not work either.

docs/proxy/datacenter_proxy.md
m-murasovs: @mnmkng There is already an `IP rotation` section above this, do you mean I should join them together?

Or maybe there's a clearer way to explain this?

Also, there's a bit in `proxy.md` that explains sessions a little - I'll add a link to it.

docs/proxy/datacenter_proxy/examples.md
m-murasovs: The browser info endpoint was used in the original examples, plus it's pretty cool - it shows you the IP address from which the request is coming, so if you're trying to use two different IPs, you can see if it works or not.

Or is there a reason why we shouldn't use it?

m-murasovs: Will everyone know that it only works in [modules](https://dev.to/mikeesto/top-level-await-in-node-2jad)?

We'd need to add extra disclaimers for everyone who's mad they're getting the 'await only works in async functions' error

docs/proxy/datacenter_proxy/examples.md
m-murasovs: I had to ask @petrpatek for help with this one and it looks like he's more thorough than me :) 
Since the `http` and `https` agents work the same, I'll just keep the `https` one

docs/proxy/datacenter_proxy/examples.md
m-murasovs: Yeah, that's a good point, I'm going to add separate examples for using proxy groups

m-murasovs: Actually, by the looks of it, the 'multiple requests with different IPs' example looks the same as 'single request with a random IP' one. The only difference is that the `useProxy()` function is called twice. 

So, actually this can be refactored to be a lot more economical

docs/proxy/datacenter_proxy.md
mnmkng: Yeah, sorry I did not realize the above section was named IP rotation. What I meant was:

There's no clear connection between "Session persistence" and "IP rotation". Under the IP rotation section, there is generic information about IP rotation and what it is, and in the Session persistence, we talk about persisting sessions. But there is very little "glue" that would tie those two together. I think the fact that's the most important: sessions represent IPs and you can therefore manage IPs by managing sessions, should have a more prominent place and should be explained very clearly.

docs/proxy/datacenter_proxy/examples.md
mnmkng: Yeah, it's very useful in the double request examples. But in single request examples the proxy.apify.com is better, because it immediately gives you information about your connection, so you have fast feedback.

I also apologize, should've sent this one: `http://proxy.apify.com/?format=json` because that's the better option for use with `axios`

mnmkng: Hmm, I missed that. Bummer. Ok, let's use the `useProxy` function. This could take a while before it becomes widely known.

docs/proxy/google_serp_proxy.md
m-murasovs: Damn, I have no idea where I got this from 😕 Doesn't even make sense, since we have examples on how to use PuppeteerCrawler with SERP. 

I think this one's a total faux pas, please forgive 😄

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -40,15 +38,15 @@ const Apify = require("apify");
...
+        "my-list", ["http://proxy.apify.com/?format=json"]
     );
     const proxyConfiguration = await Apify.createProxyConfiguration();
 
     const crawler = new Apify.PuppeteerCrawler({
         requestList,
         proxyConfiguration,
         handlePageFunction: async ({ page, request, proxyInfo }) => {
-            return Apify.pushData({ title: await page.title() });
+            console.log(page);
```
mnmkng, line 49: `console.log(page)` won't do anything good because the `page` is not nicely loggable. `console.log(await page.content())` would be better.

I'm sorry if I made it look like you need to remove all complexity from the examples. I meant only the unnecessary one, like figuring out selectors, or setting up `gotoFunction` and blocking requests and so on.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -13,13 +13,11 @@ paths:
...
 This page contains code examples for connecting to [datacenter proxies]({{@link proxy/datacenter_proxy.md}}) using [Apify Proxy](https://apify.com/proxy).
 
-If you are building your own Apify [actor]({{@link actors.md}}), below are [examples](#using-the-apify-sdk) specific to the [Apify SDK](https://sdk.apify.com).
-
 See the [connection settings]({{@link proxy/connection_settings.md}}) page for connection parameters.
 
 ## [](#using-the-apify-sdk) Using the Apify SDK
 
-If you're developing an actor using the [Apify SDK](https://sdk.apify.com), you can use Apify Proxy in:
+If you are developing your own Apify [actor]({{@link actors.md}}) using the [Apify SDK](https://sdk.apify.com), you can use Apify Proxy in:
```
mnmkng, line 20: Missing `CheerioCrawler`. I'm not saying that we necessarily need an example, but it's weird for it to be left out, when everything else is there.

docs/proxy.md
```diff
@@ -52,13 +52,11 @@ Sessions are available for [datacenter]({{@link proxy/datacenter_proxy.md#sessio
...
+You can persist your sessions (use the same IP address) by setting the `session` parameter in the `username` [field]({{@link proxy/connection_settings.md#username-parameters}}). This assigns a single IP address to a **session ID** after you make the first request.
 
 **Session IDs represent IP addresses. Therefore, you can manage the IP addresses you use by managing sessions.** In cases where you need to keep the same session (e.g. when you need to log in to a website), it is best to keep the same proxy. By assigning an IP address to a **session ID**, you can use that IP for every request you make.
 
-For datacenter proxies, a session persists for **24 hours** ([more info]({{@link proxy/datacenter_proxy.md#session-persistence}})).
-
-For residential proxies, a session persists for **1 minute** ([more info]({{@link proxy/residential_proxy.md#session-persistence}})).
+For datacenter proxies, a session persists for **24 hours** ([more info]({{@link proxy/datacenter_proxy.md#session-persistence}})). For residential proxies, it persists for **1 minute** ([more info]({{@link proxy/residential_proxy.md#session-persistence}})).
```
mnmkng, line 59: Would be good to say that by using the session you refresh the session expiry timer.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -120,7 +118,10 @@ const Apify = require("apify");
 
 Apify.main(async () => {
     const requestList = await Apify.openRequestList(
-        "my-list", ["http://www.example.com"]
+        "my-list", [
+            "http://proxy.apify.com/?format=json",
+            "http://proxy.apify.com/?format=json",
```
mnmkng, line 123: This won't work because the URLs will be deduplicated so only one page will be opened. You can use this: https://sdk.apify.com/docs/typedefs/request-list-options#keepduplicateurls but I think it would just complicate the example. You can use it once with and once without the `format=json`. Those would not be deduplicated.

docs/proxy/residential_proxy/tips_and_tricks.md
```diff
@@ -13,7 +13,7 @@ paths:
 
 Residential proxy is priced by data traffic used. Thus, it's easy to quickly use up all your prepaid traffic. In particular, when accessing websites with large files loaded on every page.
 
-To reduce your traffic use, we recommend using the Apify SDK's `blockRequests()` [function](https://sdk.apify.com/docs/api/puppeteer#puppeteerblockrequestspage-options). 
+To reduce your traffic use, we recommend using the Apify SDK's `puppeteer.blockRequests()` [function](https://sdk.apify.com/docs/api/puppeteer#puppeteerblockrequestspage-options).
```
mnmkng, line 16: I think the full name would be better `Apify.utils.puppeteer.blockRequests()`

mnmkng (commented):
Some additional things I noticed.

mtrunkat (approved):

mnmkng (approved):

m-murasovs:
@gippy as the authority on Proxy stuff, could you take a look, please? Thanks!

docs/proxy/connection_settings.md
```diff
@@ -6,18 +6,85 @@ paths:
...
+
+To learn more about [sessions]({{@link proxy.md#sessions}}) and [IP address rotation]({{@link proxy.md#ip-address-rotation}}), see the proxy overview [page]({{@link proxy.md}}).
+
+## [](#connection-examples) Connection examples
+
+* [Datacenter proxy]({{@link proxy/datacenter_proxy/examples.md}})].
+
+* [Residential proxy]({{@link proxy/residential_proxy.md#connecting-to-residential-proxy}})].
+
+* [Google SERP proxy]({{@link proxy/google_serp_proxy/examples.md}})].
```
gippy, line 84: All 3 of these have `].` at the end that should not be there.

docs/proxy/datacenter_proxy/examples.md
```diff
@@ -0,0 +1,541 @@
...
+The Apify SDK's [ProxyConfiguration](https://sdk.apify.com/docs/api/proxy-configuration) enables you to choose which proxies you use for all connections. You can inspect the current proxy's URL and other attributes using the [ProxyInfo](https://sdk.apify.com/docs/typedefs/proxy-info) property in your crawler's [page function](https://sdk.apify.com/docs/typedefs/cheerio-crawler-options#handlepagefunction).
+
+### [](#rotate-ip-addresses) Rotate IP addresses
+
+IP addresses for each request are selected at random from all available proxy servers.
+
+To use IP addresses from specific proxy groups, add a `groups` [property](https://sdk.apify.com/docs/api/proxy-configuration#docsNav)
+to `createProxyConfiguration()` and specify the group names. For example, `groups: ["GROUP_NAME_1", "GROUP_NAME_2"]`.
+
+```marked-tabs
```
gippy, line 36: The examples here all use auto proxy, which in this case since proxy configuration is used makes it hard to imagine how to actualy provide the groups mentioned above the example. Maybe also provide example of providing a proxy group?

docs/proxy.md
```diff
@@ -7,21 +7,61 @@ paths:
...
-*   [**Connection settings**]({{@link proxy/connection_settings.md}})
-*   [**Datacenter proxy**]({{@link proxy/datacenter_proxy.md}})
-*   [**Residential proxy**]({{@link proxy/residential_proxy.md}})
-*   [**Google SERP proxy**]({{@link proxy/google_serp_proxy.md}})
-*   [**Troubleshooting**]({{@link proxy/troubleshooting.md}})
+## [](#our-proxies) Our proxies
 
+[Datacenter proxy]({{@link proxy/datacenter_proxy.md}}) – the fastest and cheapest option, it uses data centers to mask your IP address. Chance of blocking due to other users' activity. [[Code examples]({{@link proxy/datacenter_proxy/examples.md}})]
+
+[Residential proxy]({{@link proxy/residential_proxy.md}}) – IP addresses located in homes and offices around the world. These IPs have the lowest chance of blocking. [[Code examples]({{@link proxy/residential_proxy.md#connecting-to-residential-proxy}})]
```
gippy, line 22: The menu on the left and this link are both missing the actual examples of usage of residential proxy. I do not think this is important, but the inconsistency bothered me a bit :)

gippy (changes requested):
Looks great, small comments

docs/proxy.md
m-murasovs: @mnmkng brought up that there's a lot of duplication if we have a separate examples page for residential, since only the 'groups' property changes.

Let's try the minimalist approach, and if it doesn't work (we can find out once we get the feedback widget), then we can add the residential examples back in :)

docs/proxy/datacenter_proxy/examples.md
m-murasovs: I added an example in the SDK section

gippy (approved):

PR #134, opened 2020-11-16
Api reference: Added actor builds and runs into apiary docs.

drobnikj:
- Added actor builds and runs into apiary docs
- For now, I just copy and update descriptions of endpoints from Actors group.

### DIFF
--- a/docs/api_v2/api_v2_reference.apib
+++ b/docs/api_v2/api_v2_reference.apib
@@ -1938,6 +1938,144 @@ In order to save new items to the dataset, send HTTP POST request with JSON payl
     + Attributes
         - data (Run, required)
 
+# Group Actor runs
+
+The API endpoints described in this section enable you to manage Apify actor run.
+
+## Run object [/v2/actor-runs/{runId}{?token,waitForFinish}]
+
+### Get run [GET]
+
+Gets an object that contains all the details about a specific run of an actor.
+
+By passing the optional `waitForFinish=1` parameter the API endpoint will synchronously wait for the run to finish.
+This is useful to avoid periodic polling when waiting for actor run to complete.
+
+This endpoint does not require the authentication token, the calls are authenticated using a hard-to-guess ID of the run.
+
++ Parameters
+
+    + runId: `3KH8gEpp4d8uQSe8T` (string, required) - Run ID.
+    + waitForFinish: 60 (number, optional) - The maximum number of seconds the server waits for the run to finish. By default it is `0`, the maximum value is `300`. <!-- MAX_ACTOR_JOB_SYNC_WAIT_SECS -->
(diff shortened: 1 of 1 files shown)


mtrunkat (approved):

docs/api_v2/api_v2_reference.apib
```diff
@@ -1938,6 +1938,144 @@ In order to save new items to the dataset, send HTTP POST request with JSON payl
...
+
+# Group Actor builds
+
+The API endpoints described in this section enable you to manage Apify actor build.
+
+## Build object [/v2/actor-builds/{buildId}{?token,waitForFinish}]
+
+### Get build [GET]
+
+Gets an object that contains all the details about a specific build of an actor.
```
m-murasovs, line 2042: Same as above: would it not be good to also move Build Collection to here?
https://docs.apify.com/api/v2#/reference/actors/build-collection

docs/api_v2/api_v2_reference.apib
```diff
@@ -1938,6 +1938,144 @@ In order to save new items to the dataset, send HTTP POST request with JSON payl
...
+The API endpoints described in this section enable you to manage Apify actor run.
+
+## Run object [/v2/actor-runs/{runId}{?token,waitForFinish}]
+
+### Get run [GET]
+
+Gets an object that contains all the details about a specific run of an actor.
+
+By passing the optional `waitForFinish=1` parameter the API endpoint will synchronously wait for the run to finish.
+This is useful to avoid periodic polling when waiting for actor run to complete.
```
m-murasovs, line 1952: Would it not be good to have Run Collection here as well? Just to have all things related to runs in the same place?
https://docs.apify.com/api/v2#/reference/actors/run-collection

m-murasovs (changes requested):

m-murasovs:
Without changing the original `Actors` section it might be pretty confusing for users to find two bits of information that are the same except for the difference in the URL.

If it's a bad time to move the `Run Collection` and `Build Collection` out of the `Actors` section, it would be good to at least add `[Deprecated]` tags for the old `'Get Run'/'Abort Run'/etc` endpoints, no?

docs/api_v2/api_v2_reference.apib
drobnikj: We plan to move list od builds there as well, but for now, we moved just build object as it a prerequisite for javascript client v1.

docs/api_v2/api_v2_reference.apib
drobnikj: Same as comment above

drobnikj:
@m-murasovs Sorry, I forgot to add deprecation warning as we discussed.

m-murasovs (approved):

PR #150, opened 2020-12-03
Updating user API endpoint docs

mtrunkat:


docs/api_v2/api_v2_reference.apib
```diff
@@ -4018,53 +4034,55 @@ The user account is identified by the provided authentication token.
...
+    - teamAccountSeatCount:             1                   (number, required)
+    - supportLevel:                     `COMMUNITY`         (string, required)
+    - availableAddOns                                       (array, required)
+
+## UserUsage (object)
+- usageCycle                                        (object, required)
+    - startAt:      `2020-11-20T00:00:00.000Z`      (string, required)
+    - endAt:        `2020-12-19T23:59:59.999Z`      (string, required)
+- usage                                             (object, required)
+    - TOTAL_USAGE_USD:      0.00023407795744696     (number, nullable)
```
mtrunkat, line 4084: I am not sure about this being snake cased capitalized but it's how we have it everywhere internally.

jancurn: Hm, it doesn't look good. When it was internal it didn't matter too much, but now it will be like this forever...

Can we perhaps rename `usage` to `platformUsage` (for consistency with UI) and move the `TOTAL_USAGE_USD` to `totalPlatformUsageUsd` ?

docs/api_v2/api_v2_reference.apib
```diff
@@ -4018,53 +4034,55 @@ The user account is identified by the provided authentication token.
...
+    - dataRetentionDays:                14                  (number, required)
+    - availableProxyGroups                                  (object, required)
+      - SOMEGROUP:                      20                  (number, required)
+      - ANOTHERGROUP:                   200                 (number, required)
+    - teamAccountSeatCount:             1                   (number, required)
+    - supportLevel:                     `COMMUNITY`         (string, required)
+    - availableAddOns                                       (array, required)
+
+## UserUsage (object)
+- usageCycle                                        (object, required)
```
jancurn, line 4080: How does this work? This is a monthly usage cycle, or full billing cycle (e.g. entire year) ?

jancurn (changes requested):

docs/api_v2/api_v2_reference.apib
jancurn: I'm thinking, we had this philosophy of providing in API all info that we provide in UI. So user should also be able to get daily data for the platform usage. Well, we don't need to do this now, but let's say we add it in the future, so the data model should count with it... let's think this API through little more. Once we launch it, there is no going back

docs/api_v2/api_v2_reference.apib
```diff
@@ -3368,6 +3368,22 @@ The user account is identified by the provided authentication token.
...
+## Usage [/v2/users/me/usage{?token}]
+
++ Parameters
+
+    + token: `rWLaYmvZeK55uatRrZib4xbZs` (string, required) - API authentication token.
+
+
+### Get user usage [GET]
+
+Returns user's usage for a current billing period.
```
m-murasovs, line 3380: a => the

m-murasovs (approved):

jancurn:
I suggest we keep the usage part of API private for now, until we're sure we want to publish it.

docs/api_v2/api_v2_reference.apib
mtrunkat: I commented-out this endpoint. So if anybody will use it it's on own risk :)

jancurn: cool

jancurn: But it seems the `## Usage [/v2/users/me/usage{?token}]` section will still be in code

drobnikj (approved):

mhamas (changes requested):
Don't merge yet pls, api will slighly change

mtrunkat:
> Don't merge yet pls, api will slighly change

I removed the docs of user usage API. There is just updated user API.

PR #157, opened 2020-12-07
Use new URL for API client

m-murasovs:
Make API Client docs use externalSourceUrl like CLI docs do. The content is now in GitHub, not in S3.

This will involve removing the redirects on Web

Excuse the branch name, I always confuse the two.


m-murasovs:
Not necessary

PR #161, opened 2020-12-23
Update state persistence docs

m-murasovs:
Preview/staging: https://docs.apify.com/actors/development/state-persistence?version=update/state-persistence

### DIFF
--- a/docs/actors/development/state_persistence.md
+++ b/docs/actors/development/state_persistence.md
@@ -8,13 +8,13 @@ paths:
 
 # [](#state-persistence)State persistence
 
-In events like server shutdown, your [actor]({{@link actors.md}}) run may need to migrate to another server. For long-running actors, this means losing the state of their progress and re-starting on the new server. This can be costly. To avoid it, long-running actors should save (persist) their state periodically and listen for [migration events](https://sdk.apify.com/docs/api/apify#apifyevents). On start, these actors should [check for persisted state](#code-example), so they can continue where they left off.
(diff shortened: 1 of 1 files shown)


mnmkng (changes requested):
Looks good to me besides a few bullet points:

- "check for persisted state" link does not scroll to the example
- not sure if we want to tell users about spot instances @jancurn 
- The message conveyed by `We have dataset, key-value store, and request queue storage so we could store results and in-progress data.` is not very clear.

jancurn:
> * not sure if we want to tell users about spot instances @jancurn

Good point @mnmkng! Yeah, this is an internal detail that doesn't need to be communicated, it might only confuse people. We can just say that the jobs might be removed from the worker machines to optimize the workload or something like that.

docs/actors/development/state_persistence.md
```diff
@@ -20,22 +20,20 @@ A migration is when a process running on a server has to stop and move to anothe
 
 **When a migration event occurs, you only have a few seconds to save your work.**
 
-## [](#why-migrations-happen)Why do migrations happen?
+## [](#why-do-migrations-happen)Why do migrations happen?
 
-Migrations happen because of the availability of Amazon EC2's **spot instances**. [Visit Amazon for more information](https://aws.amazon.com/ec2/spot/?cards.sort-by=item.additionalFields.startDateTime&cards.sort-order=asc).
+- To optimize server workloads.
+- When a server crashes (unlikely).
+- During deploys of new [worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers) software to the server.
```
mnmkng, line 27: It's actually [this worker](https://github.com/apify/apify-worker), not a web worker, that we deploy :) 

Maybe we can just say "When we release new features and fix bugs."

mnmkng (changes requested):

docs/actors/development/state_persistence.md
m-murasovs: Yep, that's way better :D

mnmkng (approved):

PR #162, opened 2020-12-28
add pagination article

metalwarrior665:
Please check my English, structure and links  :)


m-murasovs:
@metalwarrior665 this is super thorough and clear! I did some pedantic changes and added a couple of images but really only cosmetic stuff. 

The only structural change is I put the 'using filter ranges' section before the other two that don't work, just so people can get straight to the good stuff :) 

I also updated [this](https://docs.apify.com/tutorials/scrape-websites-with-limited-pagination?version=add-article/scrape-websites-with-limited-pagination#special-cases-to-look-for) heading, pls check if I understood correctly.

It should be available on staging now: https://docs.apify.com/tutorials/scrape-websites-with-limited-pagination?version=add-article/scrape-websites-with-limited-pagination

Feel free to merge when you're happy with it - looks good on my end

m-murasovs:
Also, is this an article from knowledge base or a new one?

If it's from KB, would you mind replacing the original with a bullet-point version linking to here?

metalwarrior665:
Great! But to be honest, I would keep the original order of "How to overcome the limit" as it builds up the understanding. If we would put the correct one first, the other two are meaningless then.

Other than that it looks great, thanks for the images!

This an original article.

m-murasovs:
Yeah, that makes sense. I'll change it back

PR #210, opened 2021-03-12
Add the Memory and CPU article

m-murasovs:
I compiled all the information available to me, with some linking to Docker docs for the nerdy bits.

Preview here: https://docs.apify.com/actors/memory-and-cpu?version=revamp/memory-and-cpu-article


docs/actors/memory_and_cpu.md
```diff
@@ -0,0 +1,50 @@
...
+
+[Actors]({{@link actors.md}}) run in
+[Docker containers](https://www.docker.com/resources/what-container),
+which have a [limited amount of resources](https://phoenixnap.com/kb/docker-memory-and-cpu-limit). When starting, the actor needs to be allocated a certain share of those resources, such as how much CPU "horsepower" it will be able to use.
+
+![Setting an actor's memory]({{@asset actors/images/memory-settings.png}})
+
+By assigning an actor's **memory** capacity, you also assign it a proportional amount of **CPU power**. Every CPU core corresponds to 4 GB of memory. So, 512 MB = 1/8th of a core, 1 GB = 1/4 of core, 8 GB gives you 2 cores, and so on.
+
+In addition to CPU power, the memory allocation also influences [disk size and Input/Output (I/O) limits](https://docs.docker.com/config/containers/runmetrics/#metrics-from-cgroups-memory-cpu-block-io).
```
dragonraid, line 20: We are limiting disk size, but I/O throttling is still WIP. ATM we are not enforcing it. So I would leave that part out for now. We can update the docs when we implement it.

Also I would loose the link. You do not have it in CPU or Memory section either. Besides it points to runtime metrics not the I/O constraints.

dragonraid: Please keep disk size here (it's 2x memory right @mtrunkat ?) But loose I/O limits.

m-murasovs: 🤦  thanks!

docs/actors/memory_and_cpu.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Scaling
+
+The amount of CPU power available to you is always **scaled**. So, if you are using 1GB, you have 25% of a CPU core, however that 25% is your 100%. The same applies when you are using 32 GB of memory - it is still your 100%.
+
+## CPU usage spikes
+
+![A usage spike on an actor's start-up]({{@asset actors/images/memory-cpu-usage-spike.png}})
+
+Sometimes, you see the actor’s CPU use go over 100%. This is not unusual - to help an actor start up faster, it is allocated a free CPU boost. For example, if an actor is assigned 1GB (25% of a core), it will temporarily be allowed to use 2GB so it gets started quicker.
```
mtrunkat, line 44: There is no memory boost just a CPU boost. He he will only be allowed to use 1GB memory but he can exceed the CPU to let's say full core.

docs/actors/memory_and_cpu.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Storing large amounts of data.
+
+### Maximum memory
+
+Apify actors run in [Node.js](https://nodejs.org/en/), which uses a [single process thread](https://betterprogramming.pub/is-node-js-really-single-threaded-7ea59bcc8d64). Single-process applications cannot use more than 1 CPU core. Thus, unless you use Puppeteer or other libraries that use subprocesses, you will not gain more CPU power from assigning your actor more than 4 GB of memory.
+
+In other words, giving a simple, [Cheerio-based crawler](https://apify.com/apify/cheerio-scraper) 16GB of memory (4 CPU cores) will not make it faster because these crawlers simply cannot use more than 1 CPU core.
+
+## Scaling
```
mtrunkat, line 36: This scaling is only happening in the UI chart so perhaps it does not need to be here or if so then it should clearly say that it's comment to the chart in the UI

docs/actors/memory_and_cpu.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Actors using [Puppeteer](https://pptr.dev/): at least 1GB of memory.
+
+- Large and complex sites like [Google Maps](https://apify.com/drobnikj/crawler-google-places): at least 4GB for optimal speed and [concurrency](https://sdk.apify.com/docs/api/autoscaled-pool#autoscaledpoolminconcurrency).
+
+- Storing large amounts of data.
+
+### Maximum memory
+
+Apify actors run in [Node.js](https://nodejs.org/en/), which uses a [single process thread](https://betterprogramming.pub/is-node-js-really-single-threaded-7ea59bcc8d64). Single-process applications cannot use more than 1 CPU core. Thus, unless you use Puppeteer or other libraries that use subprocesses, you will not gain more CPU power from assigning your actor more than 4 GB of memory.
```
mtrunkat, line 32: Apify actors can be built using any language so better would be to reformulate it

Actors are most commonly written using Node.js. Unless you use some external binaries such as Chrome browser/Puppeteer or multi-threading libraries the Node.js itself is single threaded and therefore cannot use more than a 1 core....

mtrunkat (changes requested):

dragonraid (approved):

docs/actors/memory_and_cpu.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Memory requirements
+
+Each use case has its own minimum memory requirements. The larger and more complex your project, the more memory/CPU power it will require. Some examples which have minimum requirements are:
+
+- Actors using [Puppeteer](https://pptr.dev/): at least 1GB of memory.
+
+- Large and complex sites like [Google Maps](https://apify.com/drobnikj/crawler-google-places): at least 4GB for optimal speed and [concurrency](https://sdk.apify.com/docs/api/autoscaled-pool#autoscaledpoolminconcurrency).
+
+- Storing large amounts of data.
```
mtrunkat, line 30: I'd replace this with something like "Working with large amount of data in memory" because data are usually stored in our storage and not in the container. But to process a lot of data at once you need a memory.

docs/actors/memory_and_cpu.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Maximum memory
+
+Apify actors are most commonly written in [Node.js](https://nodejs.org/en/), which uses a [single process thread](https://betterprogramming.pub/is-node-js-really-single-threaded-7ea59bcc8d64). Unless you use external binaries such as the Chrome browser, Puppeteer, or other multi-threaded libraries you will not gain more CPU power from assigning your actor more than 4 GB of memory because Node.js cannot use more than 1 core.
+
+In other words, giving a simple, [Cheerio-based crawler](https://apify.com/apify/cheerio-scraper) 16GB of memory (4 CPU cores) will not make it faster because these crawlers simply cannot use more than 1 CPU core.
```
mtrunkat, line 36: Perhaps we could link some article on how to write multi thread application in NodeJS = i.e. how to offload part of the workload to separate thread https://dev.to/reevranj/multiple-threads-in-nodejs-how-and-what-s-new-b23

mtrunkat (commented):
Looks good, just 2 more ideas

mtrunkat (approved):

PR #214, opened 2021-03-16
Remove broken Page Analyzer link from Web Scraping Techniques

fnesveda:
We removed the Page Analyzer page from the web because it was broken and it requires a bigger rewrite, so I removed the now broken link from the Web Scraping Techniques docs page.


m-murasovs (approved):
Good catch!

PR #221, opened 2021-04-12
Change page title for "Publishing actors"

jancurn:
I'd like to add a more descriptive page title for Google, but keep the simpler one in the menu. For example, I’d like to have page title “Publishing actors” on google, but keep just “Publishing” in the menu.

@m-murasovs is this the right way?


m-murasovs (approved):
I pushed a change that should do it :) 

Screenshot 2021-04-13 at 09 36 19

jancurn:
Great cheers Misha, you're the docs hero :)

PR #244, opened 2021-06-03
API ref: Added gracefully abort actor run parameter

drobnikj:


drobnikj:
@m-murasovs I would add info about gracefull abort somewhere in docs as well, maybe somewhere in actor/development, what do you think?

m-murasovs:
@drobnikj maybe in Actors/Running? https://docs.apify.com/actors/running

We already have bits there about running the actor via API, so we could add a sub-section on aborting runs (with some general info) and include the graceful abort?

docs/api_v2/api_v2_reference.apib
```diff
@@ -1204,6 +1204,9 @@ Only runs that are starting or running are aborted. For runs with status `FINISH
     + actorId: `janedoe~my-actor` (string, required) - Actor ID or a tilde-separated owner's username and actor name.
     + runId: `3KH8gEpp4d8uQSe8T` (string, required) - Run ID.
     + token: `rWLaYmvZeK55uatRrZib4xbZs` (string, required) - API authentication token.
+    + gracefully: true (boolean, optional)    - If true passed, the actor run will abort gracefully.
+                                                It will send aborting and persistState event into run and force kill run after 30 seconds.
+                                                It is helpful in a case where you plan to resurrect run.
```
m-murasovs, line 1210: Here, could you use backticks on param/property names to make them stand out? e.g. \`true\`, \`aborting\`, \`persistState\`

Also, maybe we can use less violence? 😄  'force kill run' => 'force-stop the run' ?

Just some pedantic note: 'It is helpful in a case where you plan to resurrect run.' => 'It is helpful in cases where you plan to resurrect the run later.'

Thanks!

m-murasovs (changes requested):

docs/actors/running.md
```diff
@@ -71,6 +71,16 @@ Each run starts with the initial status **READY** and goes through one or more t
 |ABORTING|transitional|Being aborted by user|
 |ABORTED|terminal|Aborted by user|
 
+## [](#aborting-run)Aborting run
+
+Each run in statuses **READY**, **RUNNING**, **TIMING-OUT** can be aborted by user. There are two ways how to abort the run.
+
+* Force abort run - The user immediately aborts the actor. The force abort is the default option.
```
mtrunkat, line 78: @m-murasovs is it `force` abort a good name?

Perhaps we could have just "abort" and "graceful abort"? I don't really feel what is better.

mtrunkat (changes requested):

docs/actors/running.md
m-murasovs: Yeah, we can probably go without 'force', 'abort' on its own is strong enough :)

mtrunkat (approved):

m-murasovs (approved):

PR #245, opened 2021-06-04
Docs: run storages upadte

nguyeda1:


docs/api_v2/api_v2_reference.apib
```diff
@@ -1995,7 +1995,70 @@ parameter. You can also filter runs by status ([available statuses](https://docs
             - count: 2 (number, required)
             - items (array[RunShort1,RunShort2], required)
 
-## Run object [/v2/actor-runs/{runId}{?token,waitForFinish}]
+## Run object [/v2/actor-runs/{runId}{?token,waitForFinish,status}]
+
+### Run storage endpoints
+
+This is not a single endpoint, but an entire group of endpoints that lets you to
+retrieve the last run or any of its default storages.
```
mtrunkat, line 2003: not the "last run" but "run object"

docs/api_v2/api_v2_reference.apib
```diff
@@ -1995,7 +1995,70 @@ parameter. You can also filter runs by status ([available statuses](https://docs
...
+All the endpoints require an authentication token.
```
mtrunkat, line 2004: This is not true, it's authenticated by run ID.

docs/api_v2/api_v2_reference.apib
```diff
@@ -1995,7 +1995,70 @@ parameter. You can also filter runs by status ([available statuses](https://docs
...
+
+The endpoints accept the same HTTP methods and query parameters as
+the respective storage endpoints.
+The base path that represents the last actor run object is:
+
+`/v2/actor-runs/{runId}{?token,status}`
+
+
+Using the `status` query parameter you can ensure to only get a run with a certain status
```
mtrunkat, line 2013: Is status really there? I don't think that the filtration is needed in this case compared to the last run endpoints

mtrunkat (changes requested):

docs/api_v2/api_v2_reference.apib
m-murasovs: also 'that lets you to retrieve' => 'lets you retrieve'

docs/api_v2/api_v2_reference.apib
```diff
@@ -1995,7 +1995,70 @@ parameter. You can also filter runs by status ([available statuses](https://docs
...
+
+* `/v2/actor-runs/{runId}/request-queue/requests{?token,status}` [Request collection](#reference/request-queues/request-collection)
+* `/v2/actor-runs/{runId}/request-queue/requests/{requestId}{?token,status}` [Request collection](#reference/request-queues/request)
+* `/v2/actor-runs/{runId}/request-queue/head{?token,status}` [Queue head](#reference/request-queues/queue-head)
+
+For example, to download data from a dataset of the last succeeded actor run in XML format,
+send HTTP GET request to the following URL:
+
+```
+https://api.apify.com/v2/actor-runs/{runId}/dataset/items?token={yourApiToken}&format=xml&status=SUCCEEDED
```
m-murasovs, line 2052: Here, please remove the token param from the URL - we want to encourage users to pass their auth token in the `authorization` header

m-murasovs (changes requested):

docs/api_v2/api_v2_reference.apib
nguyeda1: Should I remove it altogehter from everywhere and from Parameters as well?

docs/api_v2/api_v2_reference.apib
nguyeda1: `This is not a single endpoint, but an entire group of endpoints lets you retrieve the last run or any of its default storages.` Does not make sense or does it? It probably should be something like `
This is not a single endpoint, but an entire group of endpoints letting you retrive the last run or any of its default storages.`

docs/api_v2/api_v2_reference.apib
```diff
@@ -1997,6 +1997,62 @@ parameter. You can also filter runs by status ([available statuses](https://docs
...
+https://api.apify.com/v2/actor-runs/{runId}/dataset/items?format=xml
+```
+
+In order to save new items to the dataset, send HTTP POST request with JSON payload to the same URL.
+
++ Parameters
+
+    + runId: `3KH8gEpp4d8uQSe8T` (string, required) - Run ID.
+    + token: `rWLaYmvZeK55uatRrZib4xbZs` (string, required) - API authentication token.
+
```
drobnikj, line 2055: Log endpoint is missing

drobnikj (changes requested):
Can you do it same way as last run endpoint to be able to see `Run storage endpoints` in the left menu

drobnikj:
cc how it looks for last run endpoint
Snímek obrazovky 2021-06-07 v 19 49 47
how it looks for run storage endpoints
Snímek obrazovky 2021-06-07 v 19 50 12

mtrunkat (changes requested):

docs/api_v2/api_v2_reference.apib
m-murasovs: I'd say leave it in the params, and just remove from the sample URL, pls :)

mtrunkat (changes requested):
I meant if the token is necessary in all the endpoints here - my guess is that all of them should work without it not only the `/v2/actor-runs/{runId}/log`

nguyeda1:
> I meant if the token is necessary in all the endpoints here - my guess is that all of them should work without it not only the `/v2/actor-runs/{runId}/log`

Actually I think we do check for the token... in `actor_jobs.js => getActorRun` which is called by each endpoint.. I think it is the same for the last run

m-murasovs (approved):

mtrunkat (approved):

drobnikj (changes requested):
Looks good, just last thing, I just change the order of text and put the Get Run endpoint at the beginning of the caption and after the storages. I feel it should be the first as it is more important than storages.

drobnikj (approved):

PR #256, opened 2021-06-24
Add webp converter script and convert existing images

m-murasovs:
I took the script from Web and adapted it a little - seems to be working OK

preview: https://docs.apify.com/tutorials/quick-start?version=feature/convert-images-to-webp#run-an-actor

sorry for the huge number of changes but it's mostly minor


jbartadev (approved):
Looks good

src/webp_converter.js
```diff
@@ -0,0 +1,30 @@
...
+    async function convertImagesToWebp(dirPath) {
+        // Converts all jpg and png files in current directory to webp
+        await imagemin([`${dirPath}/*.{jpg,png}`], {
+            destination: dirPath,
+            plugins: [
+                imageminWebp({ quality: 65 }),
+            ],
+        });
+
+        const files = fs.readdirSync(dirPath);
```
mtrunkat, line 20: nitpick - you have async function so it's always better to use async `fs.readdir` rather than blocking `readdirSync`:

`const { readdir } = require('fs/promises')`

`... await readdir`.

It's not a big deal here but to keep everything clean

src/webp_converter.js
```diff
@@ -0,0 +1,30 @@
...
+
+        // Find directories in current directory and convert their images to webp
+        await Promise.all(files.filter(file => fs.statSync(`${path.join(path.resolve(path.dirname('')), dirPath)}/${file}`).isDirectory())
```
mtrunkat, line 23: You do all the images in parallel so at some point the GitHub action could run out of memory. Might be better to do it in a serie:

```
for (const file of files) {
 ....
}
```

or chunks

```
for (const files of _.chunk(allFiles, 50)) {
   ...
}
```

but maybe not necessary now and you can fix it once it becomes a problem.

src/webp_converter.js
```diff
@@ -0,0 +1,30 @@
...
+            .map(async directory => convertImagesToWebp(`${dirPath}/${directory}`)));
+    }
+    console.log('Webp conversion START');
+    // Start converting from docs folder
+    await convertImagesToWebp('docs');
+    console.log('Webp conversion DONE');
+})();
```
mtrunkat, line 30: You are running this main async function but there is no try/catch so it may not even fail the GitHub action in a case of an error. It will just log unhandled promise rejection but wont fail the execution with current Node version. To properly handle It you can do one of:

```
(async function () {
  try {

  } catch (err) {
    log.exception(err, ....);
    process.exit(1);
  }
})();
```

```
(async function () {

})().catch((err) => {
    log.exception(err, ....);
    process.exit(1);
});
```

mtrunkat (changes requested):

m-murasovs:
@mtrunkat thanks for the feedback! I guess I'll have to update the script in the web repo as well, since I got the code from there 😄

mtrunkat:
Cook, last thing - all sync functiona are evil including `statSync`

mtrunkat (changes requested):
last note in a comment above

mtrunkat (approved):

PR #271, opened 2021-08-30
Improve wording in list of scraper tutorials

fnesveda:
The wording with all the parentheses and brackets was hurting my brain, I rewrote it a bit. It still sounds kinda forced, but I think it reads better.


m-murasovs (approved):
Yep, this is waaaaay better!

PR #284, opened 2021-11-10
Fix lint, update packages, add package-lock, update to Node 14

fnesveda:
Because we were not committing lockfiles, some linting package updated and broke lint. When I was fixing it, I got in too deep, and I:
- updated everything to Node 14
- updated packages
- committed the `package-lock.json`
- fixed the original lint issue
- fixed some more lint issues
- did a bit of cleanup

We need to merge this relatively quickly, because without it we can't deploy other changes. This should be easy to check, if if it builds and displays alright, then it's fine.


m-murasovs (approved):
Just a casual Wednesday :D

src/build.js
```diff
@@ -1,6 +1,6 @@
 const _ = require('underscore');
 const path = require('path');
-const log = require('apify-shared/log');
+const log = require('@apify/log').default;
```
valekjo, line 3: Why is the `default` needed?

valekjo (approved):
Looks good, just one question.

src/build.js
fnesveda: Because `require('some-module')` loads whatever is in `some-module`'s `module.exports`, and the default export is, according to the spec, just an export named `"default"`. `import` can deal with this automatically, if you want to import the "whole module", it reads the `default` export from `module.exports`, but `require()` can't.

It worked before because Babel had a hack that hack that if a module had only the default export and no other exports, it would assign `module.exports = theDefaultExport`, so that you could do the straightforward `require()`s without the `.default` part. But now that we switched `apify-shared` to TypeScript and lerna, we don't use Babel anymore, so the `require()`s of default exports have to be done like this.

PR #308, opened 2021-11-30
feat: Adding info about restricted residential service

mtrunkat:


mtrunkat:
@jancurn - we need to decide about naming - for me the restricted and full does make the most sense. 

CC @mhamas - here is the doc

docs/proxy/residential_proxy.md
```diff
@@ -19,7 +19,31 @@ Residential proxies support [IP address rotation]({{@link proxy.md#ip-address-ro
...
+Apify provides 2 levels of residential proxy:
+
+| Level | Availability | Supported domains | Other limitations |
+|-------|--------------|-------------------|-------------------|
+| Restricted residential proxy | Every user | >500 domains | Requires man in the middle access to connection |
+| Full residential proxy | Enterprise level accounts | Entire web| None |
+
+### Restricted residential proxy
+
+Restricted residential proxy is available for all the users with some limitations.
```
jancurn, line 33: Restricted ... with some limitation. That's like double restricted :)

docs/proxy/residential_proxy.md
```diff
@@ -19,7 +19,31 @@ Residential proxies support [IP address rotation]({{@link proxy.md#ip-address-ro
...
+
+First of restricted residential proxy supports only certain domains and paths. The pool of 500 domains is increased every month and covers the most frequent use cases.
```
jancurn, line 35: Typo? "First of all" or "First, ... " ?

jancurn:
> @jancurn - we need to decide about naming - for me the restricted and full does make the most sense.
> 
> CC @mhamas - here is the doc

Let's imagine you're talking to a custom and explaining to them what they need. If you say "full residential proxy" it might not click to them what tham means. How about "unrestricted" ? that's obvious what it means. 

 @m-murasovs what do you think?

mhamas:
Thx @mtrunkat , I added the route to my PR

docs/proxy/residential_proxy.md
```diff
@@ -19,7 +19,31 @@ Residential proxies support [IP address rotation]({{@link proxy.md#ip-address-ro
...
+| Level | Availability | Supported domains | Other limitations |
+|-------|--------------|-------------------|-------------------|
+| Restricted residential proxy | Every user | >500 domains | Requires man in the middle access to connection |
+| Full residential proxy | Enterprise level accounts | Entire web| None |
+
+### Restricted residential proxy
+
+Restricted residential proxy is available for all the users with some limitations.
+
+First of all, restricted residential proxy supports only certain domains and paths. The pool of 500 domains is increased every month and covers the most frequent use cases.
```
m-murasovs: > The pool of 500 domains is increased every month and covers the most frequent use cases.

Does this mean the user starts out with 500 domains and then gets more?

docs/proxy/residential_proxy.md
```diff
@@ -19,7 +19,31 @@ Residential proxies support [IP address rotation]({{@link proxy.md#ip-address-ro
...
+
+The second limitation is that restricted residential proxy uses the man-in-the-middle system to monitor traffic
+and activities and so requires acceptance of an SSL certificate. This is automatically handled by the [Apify SDK](https://sdk.apify.com/) for both Puppeteer and Playwright.
+
+### Full residential proxy
+
+Full residential proxy neither limits the domains you can access nor requires a man-in-the-middle access to traffic.
+However, it's provided only to enterprise-level accounts on a per-request basis and under an additional contract.
+
+[Contact us](https://apify.com/contact) if you would like to use the full residential proxy or for more information.
```
m-murasovs, line 46: Maybe a `mailto:support@apify.com` link here? We could even set a custom subject for the email

m-murasovs (changes requested):

m-murasovs:
@jancurn yep, 'unrestricted' would work. I think the customers will still need explanations about what the restricted one is, but there's nothing we can do on that :)

docs/proxy/residential_proxy.md
mtrunkat: Nope that the new domains are being enabled in the system globally

m-murasovs: So then you'll still be able to crawl 500 different domains but the overall list will be bigger?

docs/proxy/residential_proxy.md
```diff
@@ -19,7 +19,31 @@ Residential proxies support [IP address rotation]({{@link proxy.md#ip-address-ro
...
+
+### Unrestricted residential proxy
+
+Restricted residential proxy is available for all the users with some conditions.
+
+Firstly, restricted residential proxy supports only certain domains and paths. The pool of 500 domains is increased every month and covers the most frequent use cases.
+Any traffic outside this pool of domains will go through [datacenter proxy]({{@link proxy/datacenter_proxy.md}}).
+
+The second limitation is that restricted residential proxy uses the man-in-the-middle system to monitor traffic
+and activities and so requires acceptance of an SSL certificate. This is automatically handled by the [Apify SDK](https://sdk.apify.com/) for both Puppeteer and Playwright.
```
jancurn, line 39: Can we add a link to the certificate?

jancurn: Plus info how can you tell if the connection is MITM or not - on proxy.apify.com page

m-murasovs (approved):
Caught some HTML in Markdown - controversial! 😄

PR #309, opened 2021-11-30
Update storage.md

jancurn:
It is "Apify Console", not "the Apify console" ! I guess there are many other places where we have it like this


m-murasovs (approved):
Thanks for the catch - I genuinely thought I'd done this

PR #310, opened 2021-12-10
Update events.md

jancurn:


m-murasovs (changes requested):
Nice, just please fix the lint error. It should be `text` here

Screenshot 2021-12-10 at 11 56 24

jancurn:
> Nice, just please fix the lint error. It should be `text` here
> 
> Screenshot 2021-12-10 at 11 56 24

Aye Captain!

m-murasovs (approved):

PR #321, opened 2022-02-11
[DO NOT MERGE] Prepare repo structure for academy split

m-murasovs:
- restructuring the content so it would live in the below repos
- 	`/content/docs`
-  `/content/academy`

This way, we can add future folders there and not clutter the root of the dir.

- Updated the build script to automatically find all the folders in `/content` and create builds for them in `/build` (so we don't need to keep manually adding folders and duplicating code)
- Updated the webp converter script to be automatic as well
- Slight update to the README but I expect we'll need to add more stuff about the academy
- Tried to document the code better because it took me ages to understand what each function does. Please add suggestions how we can make it clearer

**Important**
I don't think we need to update the S3, apiary, and Branch scripts, but please have a look - I'm not too sure about how this works.

P.S. we really need more people who know about how the docs "backend" works because currently it's only Mara and kinda me

apify/apify-web#1574


src/webp_converter.js
```diff
@@ -19,21 +21,31 @@ const path = require('path');
...
-            await convertImagesToWebp('docs');
-            console.log('Webp conversion DONE');
+
+            // Get all the content dirs
+            const contentDirs = await getDirectoriesInPath(CONTENT_DIR_NAME);
+
+            // Call the conversion function for each content dir found
+            // It will recursively scan all the files in the given directory and convert them
+            contentDirs.map(async (dir) => {
+                await convertImagesToWebp(dir);
```
mtrunkat: Oh no, @m-murasovs, promises won't work this way! With map/forEach you would have to use the async package. I'd go with `for (... of`

mtrunkat (changes requested):

mtrunkat (approved):
Should be fine but we need to be careful with merge not to break our docs!

Strajk (approved):
The scripts folder seems little bit complicated for my taste, but it looks like it works correctly 👍

PR #326, opened 2022-03-02
API docs: add missing defaultRunOptions

m-murasovs:
As reported on Slack: https://apifier.slack.com/archives/C01GVMUSRUN/p1646135343743749


pocesar (commented):
LGTM. I just couldn't provide just one of the options, had to provide all of them

received: `Invalid value provided in updatedActor: defaultRunOptions.build is required` and ~`Invalid value provided in updatedActor: defaultRunOptions.memoryMBytes is not allowed by the schema`~ (nevermind, it's a typo, should be with a `b`)

PR #328, opened 2022-03-13
Write + Add API Scraping Article

mstephen19:


m-murasovs:
@mnmkng should we make the base for this PR your re-structure one, so we can avoid conflicts, etc?

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -0,0 +1,158 @@
...
+Sometimes, a response will look something like this:
+
+```JSON
+{
+    "title": "Scraping Academy Message",
+    "message": "SGVsbG8hIFlvdSBoYXZlIHN1Y2Nlc3NmdWxseSBkZWNvZGVkIHRoaXMgYmFzZTY0IGVuY29kZWQgbWVzc2FnZSEgV2UgaG9wZSB5b3UncmUgbGVhcm5pbmcgYSBsb3QgZnJvbSB0aGUgQXBpZnkgU2NyYXBpbmcgQWNhZGVteSE="
+}
+```
+
+Or some other encoding format. This example's `message` has some data encoded in <a href="https://en.wikipedia.org/wiki/Base64">Base64</a>, which is one of the most common encoding types. For testing out Base64 encoding and decoding, you can use <a href="https://www.base64encode.org/">base64encode.org</a> and <a href="https://www.base64decode.org/">base64decode.org</a>. Within a project where base64 decoding/encoding is necessary, a package such as <a href="https://www.npmjs.com/package/base-64">`base-64`</a> can be used to encode and decode Base64 data within Node.js code.
```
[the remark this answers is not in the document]
mnmkng: Also, you don't need a package in Node.js to decode base64:


```js
Buffer.from(string, 'base64').toString('utf-8')
```

mnmkng (changes requested):
Thanks a lot Matt. Highly appreciated and I hope many more articles like this will come 🎉 

I left a few comments, some about style of writing, others about content, but in general I think the text is written really well and I enjoyed reading it.

Aside from the comments, the only thing I'm missing are some runnable code examples. I think every dev tutorial should have some. That's what makes them interactive.

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -0,0 +1,158 @@
...
+
+Since the data is coming directly from the site's API, as opposed from the parsing of HTML content based on CSS selectors, it can be relied on more, as it is less likely to change. Typically, websites change their APIs much less frequently than they change the structure/selectors of their pages.
+
+### 2. Configurable
+
+Most APIs accept query parameters such as `maxPosts` or `fromCountry`. These parameters can be mapped to the configuration options of the scraper, which makes creating a scraper that supports various requirements and use-cases much easier. They can also be utilized to easily filter and/or limit data results.
+
+### 3. Fast and efficient
+
+Especially for <a href="https://blog.apify.com/what-is-a-dynamic-page/" target="_blank">dynamic sites</a>, in which a headless browser is required (which can sometimes be slow and cumbersome), scraping their API can prove to be much quicker and more efficient.
```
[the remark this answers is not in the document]
m-murasovs: Yep, they should do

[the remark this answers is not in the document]
mnmkng: Cool. So we can use markdown syntax here as well @mstephen19

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -0,0 +1,158 @@
...
+
+_Here's what we can see see in the Network tab after reloading the page:_
+
+![Network tab results after completing an action on the page which results in the API being called]({{@asset data_collection_basics/images/results-in-network-tab.png}})
+
+Let's say that our target data is a full list of Tiësto's uploaded songs on SoundCloud. We can use the "`Filter`" option to search for the keyword "`tracks`", and see if any endpoints have been hit that include that word. Multiple results may still be in the list when using this feature, so it is important to carefully examine the paylods and responses of each request in order to ensure that the correct one is found.
+
+> **Note:** The keyword/piece of data that is used in this filtered search should follow the same logic as the previous step - it should be a target keyword or a piece of target data that that can be assumed will most likely be apart of the endpoint.
```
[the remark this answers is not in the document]
mstephen19: I always make a mistake with this one even as a native! aaagh! apart = separated, a part = a piece of another thing

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -0,0 +1,158 @@
...
+const HEADERS = {
+    'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.110 YaBrowser/22.1.0.2500 Yowser/2.5 Safari/537.36',
+    Referer: 'https://soundcloud.com',
+    ...
+}
+```
+
+However, a much better option is to use either a custom implementation of generating random headers for each request, or to use a package such as <a href="https://www.npmjs.com/package/header-generator">`header-generator`</a> to automatically do this.
+
+For our SoundCloud example, testing the endpoint from the previous section in a tool like <a href="https://www.postman.com/">Postman</a> works perfectly, and returns the data we want; however, when the "`client_id`" parameter is removed, we receive a `401 Unauthorized` error. Luckily, the Client ID is the same for every user, which means that it is not tied to a session or an IP address (this is based on our own observations and tests). The big downfall is that the token being used by SoundCloud changes every few weeks, so it shouldn't be hardcoded. **A very strange case indeed**.
```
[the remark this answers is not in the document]
mstephen19: very true. doing this now

content/academy/data_collection_basics/api_scraping.md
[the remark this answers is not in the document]
mstephen19: Oh yeah duh, I did this the other day

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -0,0 +1,223 @@
...
+    await browser.close();
+    console.log(clientId); // log the retrieved client_id
+};
+
+scrapeClientId();
+```
+
+# [](#extra-challenges) Extra challenges
+
+> **Note:** these concepts will be covered more in-depth later on, but it is good to be aware of them now.
```
m-murasovs, line 188: It's generally good practice to add some text between two headings, so maybe this doesn't have to be a note? Instead, maybe you could introduce what's to come and add this at the end?

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -0,0 +1,223 @@
...
+
+## 1. Different data formats
+
+APIs come in all different shapes and sizes. That means every API will vary in not only the quality of the data that it returns, but also that format that it is in. The two most common formats are `JSON` and `HTML`.
```
m-murasovs, line 192:
```suggestion
APIs come in all different shapes and sizes. That means every API will vary in not only the quality of the data that it returns, but also that format that it is in. The two most common formats are JSON and HTML.
```

content/academy/data_collection_basics/js_in_html.md
```diff
@@ -0,0 +1,51 @@
...
+
+The advantages of using these objects instead of parsing the HTML are that parsing JSON is much simpler, and more reliable than parsing HTML elements. They are much less likely to change, while the CSS selectors are prone to updates and renamings every time the website is updated.
+
+> **Note:** In this tutorial, we'll be using [SoundCloud's website](https://soundcloud.com) as an example target, but the techniques described here can be applied to any site.
+
+## [](#locating) Locating JSON objects within `<script>` tags
+
+Using our DevTools, we can inspect our <a href="https://soundcloud.com/tiesto/tracks">target page</a>, or right click the page and click "View Page Source" to see the DOM. Next, we'll find a value on the page that we can predict would be in a potential API response. For our page, we'll use the "Tracks" count of `845`. On the "View Page Source" page, we'll do <kbd>⌘</kbd> + <kbd>F</kbd> and type in this value, which will show all matches for it within the DOM. This method can expose `<script>` tag objects which hold the target data.
```
m-murasovs, line 19: > <kbd>⌘</kbd> + <kbd>F</kbd>

This is cool! Didn't know it was a thing :)

m-murasovs (changes requested):
Thanks for adding these! Making Docs PRs gives you +50 Apify karma.

I just have some formatting tips to make it consistent with the rest of docs.

Also, it would be great to keep the images as small as possible, so I'd appreciate if you could trim the size (in pixels) of `content/academy/data_collection_basics/images/view-object-in-window.png` and `content/academy/data_collection_basics/images/results-in-network-tab.png`. Ideally,  they'd be around 150kb, so maybe you could trim all the non-essential info, zoom in so the small print is more visible, etc? The images get converted to webp but it's still good to keep them as lean as possible so Google doesn't rate the page badly

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -0,0 +1,223 @@
+---
+title: API Scraping
+description: In this API scraping tutorial, you will learn how to leverage website APIs in web scraping and collecting data.
+menuWeight: 21.1
+paths:
+    - data-collection-basics/api-scraping
+---
+
+# [](#what) What is API scraping?
```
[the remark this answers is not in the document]
mnmkng: So would it be better if I removed anchors from all the titles in the academy?

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -90,20 +76,39 @@ Since our request doesn't have any body/payload, we just need to analyze the URL
...
-Based on our observations, we can modify the URL and utilize the "`limit`" option to return 200 songs in one request, instead of just 20:
+Let's say we want to receive all of the user's tracks in one request. Based on our observations of the endpoint's different parameters, we can modify the URL and utilize the "`limit`" option to return more than just twenty songs:
 
 ```
-https://api-v2.soundcloud.com/users/141707/tracks?client_id=zdUqm51WRIAByd0lVLntcaWRKzuEIB4X&limit=200
+https://api-v2.soundcloud.com/users/141707/tracks?client_id=zdUqm51WRIAByd0lVLntcaWRKzuEIB4X&limit=99999
 ```
 
+By using the ridiculously large number of `99999`, we ensure that all of the user's tracks will be captured in this single request.
+
```
mnmkng, line 86: Sorry. I guess I wasn't entirely clear before. My point was: how did we find that we can use the number 200, or 99999, or any other number. I think it might not be entirely obvious to people who are new to API scraping that they can just change it to whatever they want and see what happens. And that they might get overflow errors, in which case they need to reduce the number, and so on.

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -90,20 +76,39 @@ Since our request doesn't have any body/payload, we just need to analyze the URL
...
+Luckily, there are ways to retrive and set cookies for requests prior to sending them, which will be covered more in-depth within future Scraping Academy modules. The most important things to know at the moment are:
 
 1. For sites that heavily rely on cookies for user-verification and request authorization, certain generic requests (such as to the website's main page, or to the target page) will return back a (or multiple) "`set-cookie`" header(s).
-2. The `set-cookie` response header(s) can be parsed and used as the `cookie` header in the headers of a request. A great package for parsing these values from a response's headers is <a href="https://www.npmjs.com/package/set-cookie-parser">`set-cookie-parser`</a>
+2. The `set-cookie` response header(s) can be parsed and used as the `cookie` header in the headers of a request. A great package for parsing these values from a response's headers is [`set-cookie-parser`](https://www.npmjs.com/package/set-cookie-parser). With this package, cookies can be parsed from headers like so:
+
+```JavaScript
+const axios = require('axios');
+const setCookieParser = require('set-cookie-parser'); // import the set-cookie-parser module
```
mnmkng, line 98: Let's not do this now, but we should think about promoting the SDK in the code examples and courses. Maybe we could add some helpers that make working with cookies easier. Ideas welcome.

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -122,9 +127,61 @@ const HEADERS = {
...
+        ],
+        devices: ['desktop'],
+        locales: ['de-DE', 'en-US'],
+        operatingSystems: ['windows', 'linux'],
+    },
+    headers: {
+        'some-header': 'Hello, Academy!'
+    }
+})
+```
```
mnmkng, line 153: This is an interesting thing to discuss. On one hand, this makes usage of got-scraping look difficult, which is not the case, because the defaults are reasonable and the headers will get automatically generated for you. On the other hand, without showing the code, you don't show all the options.

I think over time, we settled on using the simplest possible examples - not to scare the users - and explaining the options separately. But I'll happily hear your thoughts on this.

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -122,9 +127,61 @@ const HEADERS = {
...
+
+For our SoundCloud example, testing the endpoint from the previous section in a tool like [Postman](https://www.postman.com/) works perfectly, and returns the data we want; however, when the "`client_id`" parameter is removed, we receive a `401 Unauthorized` error. Luckily, the Client ID is the same for every user, which means that it is not tied to a session or an IP address (this is based on our own observations and tests). The big downfall is that the token being used by SoundCloud changes every few weeks, so it shouldn't be hardcoded. This case is actually quite common, and is not only seen with SoundCloud.
```
mnmkng, line 155: The link should go to the Postman article. For two reasons. Keeping the user longer on our domain is good for SEO and we can explain the web scraping intricacies of postman better in our own article than just linking to the Postman docs.

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -122,9 +127,61 @@ const HEADERS = {
...
+Here is a way you could dynamically scrape the `client_id` using Puppeteer:
+
+```JavaScript
+const puppeteer = require('puppeteer'); // import the puppeteer module
+
+const scrapeClientId = async () => {
+    const browser = await puppeteer.launch({ headless: false });
+    const page = await browser.newPage();
+
+    let clientId = null; // initialize a variable that will eventually hold the client_id
```
mnmkng, line 168: I would place comments on top of the code, instead of next to it. On smaller screens, they won't wrap that much.

content/academy/data_collection_basics/js_in_html.md
```diff
@@ -0,0 +1,51 @@
+---
+title: JavaScript in HTML
```
mnmkng, line 2: Maybe "JavaScript data in HTML" ? This title seems like the text is gonna be about adding scripts inside HTML.

content/academy/data_collection_basics/api_scraping.md
```diff
@@ -152,29 +158,37 @@ const response = await gotScraping({
...
+For our SoundCloud example, testing the endpoint from the previous section in a tool like [Postman](./postman.md) works perfectly, and returns the data we want; however, when the `client_id` parameter is removed, we receive a _**401 Unauthorized**_ error. Luckily, the Client ID is the same for every user, which means that it is not tied to a session or an IP address (this is based on our own observations and tests). The big downfall is that the token being used by SoundCloud changes every few weeks, so it shouldn't be hardcoded. This case is actually quite common, and is not only seen with SoundCloud.
```
mnmkng, line 161: Links to internal files work a little differently. But don't worry about it now, I'll change it anyway after the merge, to prepare for the upcoming Academy structure. Postman will go under Glossary.
image

mnmkng (approved):
looks good to me, I'll make final edits (menu weights, folder structure etc in my branch)

content/academy/data_collection_basics/js_in_html.md
```diff
@@ -0,0 +1,51 @@
...
+
+The advantages of using these objects instead of parsing the HTML are that parsing JSON is much simpler, and more reliable than parsing HTML elements. They are much less likely to change, while the CSS selectors are prone to updates and renamings every time the website is updated.
+
+> **Note:** In this tutorial, we'll be using [SoundCloud's website](https://soundcloud.com) as an example target, but the techniques described here can be applied to any site.
+
+## [](#locating-json-in-html) Locating JSON objects within `<script>` tags
+
+Using our DevTools, we can inspect our <a href="https://soundcloud.com/tiesto/tracks">target page</a>, or right click the page and click "View Page Source" to see the DOM. Next, we'll find a value on the page that we can predict would be in a potential API response. For our page, we'll use the "Tracks" count of `845`. On the "View Page Source" page, we'll do <kbd>⌘</kbd> + <kbd>F</kbd> and type in this value, which will show all matches for it within the DOM. This method can expose `<script>` tag objects which hold the target data.
```
m-murasovs, line 19: For regular links, we use Markdown syntax
```suggestion
Using our DevTools, we can inspect our [target page](https://soundcloud.com/tiesto/tracks), or right click the page and click **View Page Source** to see the DOM. Next, we'll find a value on the page that we can predict would be in a potential API response. For our page, we'll use the **Tracks** count of `845`. On the **View Page Source** page, we'll do <kbd>⌘</kbd> + <kbd>F</kbd> and type in this value, which will show all matches for it within the DOM. This method can expose `<script>` tag objects which hold the target data.
```

content/academy/data_collection_basics/js_in_html.md
```diff
@@ -0,0 +1,51 @@
...
+
+![Find the value within the DOM using CMD + F]({{@asset data_collection_basics/images/view-845.png}})
+
+These data objects will usually be attached to the window object (often prefixed with two uderscores - `__`). When scrolling to the beginning of the script tag on our "View Page Source" page, we see that the name of our target object is `__sc_hydration`. Heading back to DevTools and typing this into the console, the object is displayed.
```
m-murasovs, line 23:
```suggestion
These data objects will usually be attached to the window object (often prefixed with two uderscores - `__`). When scrolling to the beginning of the script tag on our **View Page Source** page, we see that the name of our target object is `__sc_hydration`. Heading back to DevTools and typing this into the console, the object is displayed.
```

content/academy/data_collection_basics/postman.md
```diff
@@ -0,0 +1,63 @@
...
+
+There are four main sections to get familiar when starting out with Postman:
+
+1. Tabs
+
+Multiple test endpoints/requests can be opened at one time, each of which will be held within its own tab.
+
+2. Address bar
+
+The section in which you select the type of request to send, the URL of the request, and of course, send the request with the "Send Request" button.
```
m-murasovs, line 27:
```suggestion
The section in which you select the type of request to send, the URL of the request, and of course, send the request with the **Send Request** button.
```

content/academy/data_collection_basics/postman.md
```diff
@@ -0,0 +1,63 @@
...
+
+3. Request options
+
+This is a very useful section where you can view and edit structured query parameters, as well as specify any authorization parameters, headers, or payloads.
+
+4. Response
+
+After sending a request, the response's body will be found here, along with its cookies and headers. The response body can be viewed in various formats - "Pretty-Print", "Raw", or "Preview".
```
m-murasovs, line 35:
```suggestion
After sending a request, the response's body will be found here, along with its cookies and headers. The response body can be viewed in various formats - **Pretty-Print**, **Raw**, or **Preview**.
```

content/academy/data_collection_basics/postman.md
```diff
@@ -0,0 +1,63 @@
...
+
+## [](#using-proxies) Using and testing proxies
+
+In order to use a proxy, the proxy's server and configuration must be provided in the "Proxy" tab in Postman settings.
```
m-murasovs, line 39:
```suggestion
In order to use a proxy, the proxy's server and configuration must be provided in the **Proxy** tab in Postman settings.
```

content/academy/data_collection_basics/postman.md
```diff
@@ -0,0 +1,63 @@
...
+
+![Proxy configuration in Postman settings]({{@asset data_collection_basics/images/postman-proxy.png}})
+
+After configuring a proxy, the next request sent will attempt to use it. To switch off the proxy, its details don't need to be deleted. The `Add a custom proxy configuration` option in settings just needs to be un-ticked to disable it.
```
m-murasovs, line 43:
```suggestion
After configuring a proxy, the next request sent will attempt to use it. To switch off the proxy, its details don't need to be deleted. The **Add a custom proxy configuration** option in settings just needs to be un-ticked to disable it.
```

content/academy/data_collection_basics/postman.md
```diff
@@ -0,0 +1,63 @@
...
+
+More often than not in these cases, the reason is because the endpoint being reached requires a valid `cookie` header to be present when sending the request, and because of Postman's cache, it is sending a valid cookie within each request's headers, while your scraper is not. Another reason this may happen is because you are sending Postman requests without a proxy (using your local IP address), while your scraper is using a proxy that could potentially be getting blocked.
+
+In order to check whether there are any cookies associated with a certain request are cached in Postman, click on the 'Cookies' button in any opened request tab:
+
+![Button to view the cached cookies]({{@asset data_collection_basics/images/postman-cookies-button.png}})
+
+Clicking on this button opens a "MANAGE COOKIES" window, where a list of all cached cookies per domain can be seen. If we had been previously sending multiple requests to `https://github.com/apify`, within this window we would be able to easily find cached cookies associated with github.com. Cookies can also be easily edited (to update some specific values), or deleted (to send a "clean" request without any cached data) here.
```
m-murasovs, line 55:
```suggestion
Clicking on this button opens a **MANAGE COOKIES** window, where a list of all cached cookies per domain can be seen. If we had been previously sending multiple requests to **https://github.com/apify**, within this window we would be able to easily find cached cookies associated with github.com. Cookies can also be easily edited (to update some specific values), or deleted (to send a "clean" request without any cached data) here.
```

content/academy/data_collection_basics/postman.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Postman
+description: Learn about Postman, a simple tool for testing requests and proxies
```
m-murasovs, line 3: It's good for SEO to have the meta descriptions between 120 and 160 characters long. If you can wizard up some longer descriptions for these files, that would be awesome ❤️

m-murasovs (changes requested):
A few more nitpicks, but it's coming together well!

m-murasovs (approved):
Awesome, I think we're good to go :)

PR #354, opened 2022-04-29
Puppeteer/Playwright course

mstephen19:


content/academy/puppeteer_playwright/browser.md
```diff
@@ -0,0 +1,90 @@
...
+
+const browser = await puppeteer.launch({ headless: false });
+
+// code will be here in the future
+
+await browser.close();
+</marked-tab>
+```
+
+In a few lessons from now, we'll be discussing the `browser.contexts()` (Playwright)/`browser.browserContexts()` (Puppeteer) functions and how to create **browser contexts** with another function on the **Browser** object.
```
mnmkng, line 80: This feels a bit out of place. At this moment, the readers don't know what the contexts are, so the functions are not very meaningful to them.

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,311 @@
...
+
+const browser = await puppeteer.launch({ headless: false });
+
+const page = await browser.newPage();
+
+await browser.close();
+</marked-tab>
+```
+
+Then, we can visit a website with the `page.goto()` method. Let's go to [Google](https://google.com) for now. We'll also create a custom function called `sleep`, which will force the program to wait for a number of seconds before moving quitting (otherwise, everything will just flash before our eyes and we won't really tell what's going on):
```
mnmkng, line 34: I think we could use `page.waitForTimeout` instead.

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,311 @@
...
+
+Though it seems complex, the wonderful **Page** API makes all of these actions extremely easy to perform.
+
+### Clicking & pressing keys
+
+Let's first focus on the first 3 steps listed above. By using `page.click()` and the CSS selector of the element to click, we can click an element:
+
+```JavaScript
+// Click the "I understand" button
+await page.click('button + button:nth-child(2)');
```
mnmkng, line 95: Playwright can click based on content. Maybe even Puppeteer can now. https://playwright.dev/docs/selectors This is really good for action buttons like "Log in" or "I understand" as the text is less likely to change than the CSS.

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,311 @@
...
+```
+
+Then, we can type some text into an input field with `page.type()`; passing a CSS selector as the first, and the string to input as the second parameter:
+
+```JavaScript
+// Type the query into the search box
+await page.type('input[title]', 'hello world');
+```
+
+Finally, we can press a single key by access the `keyboard` property of `page` and calling the `press()` function on it:
```
mnmkng, line 105: accessing

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,311 @@
...
+</marked-tab>
+```
+
+When we run it, we leave off on the results page:
+
+![Google results page reached by headless browser]({{@asset puppeteer_playwright/images/google-results.webp}})
+
+### Waiting for elements/events
+
+In a perfect world, every piece of content served on a website would be loaded instantly. We don't live in a perfect world though, and often times it can take anywhere 1/10th of a second to a few seconds to load some content onto a page. Puppeteer and Playwright don't sit around waiting for the pages to load though - if we tell it to do something with an element that hasn't rendered yet, it'll start trying to do it (which will result in nasty errors). We've got to tell it to wait.
```
mnmkng, line 169: anywhere between 1/10

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,311 @@
...
+
+Different events can be waited for using the various `waitFor...` methods offered, while elements can be waited for with `page.waitForSelector()` in both Playwright and Puppeteer. In Playwright you can also create a locator for an element and use the `locator.waitFor()` function.
+
+```marked-tabs
+<marked-tab header="Playwright" lang="javascript">
+// Create a pointer for the first result element
+const firstResult = page.locator(':nth-match(.g a, 1)');
+
+// Wait for the first result element to be visible
+await firstResult.waitFor();
```
mnmkng, line 179: I think Playwright auto-waits before clicking, so this shouldn't be necessary. Not sure about Puppeteer, but it didn't do that the last time I checked. Yet another difference that might be worth mentioning.

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,311 @@
...
+import * as fs from 'fs/promises';
+import path from 'path';
+
+// ...
+
+// Take the screenshot
+const screenshot = await page.screenshot();
+
+// Write it to the filesystem
+await fs.writeFile(path.join(path.resolve(), 'screenshot.jpg'), screenshot);
```
mnmkng, line 236: I think the path.join / resolve is redundant. It will write the screenshot to the PWD anyway.

Also, I'm not sure, but I think the default image type is PNG. At least it was that historically in Puppeteer.

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,311 @@
...
+```
+
+## [](#final-code) Final code
+
+Here's what our final code looks like:
```
mnmkng, line 241: I think by this time everyone forgot the 7 steps at the beginning so it might be worth repeating them.

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,311 @@
...
+</marked-tab>
+<marked-tab header="Puppeteeer" lang="javascript">
+// Click the first result
+await page.click('.g a');
+
+// Grab hold of the page's title
+const title = await page.title();
+
+// Log the title to the console
+console.log(title);
```
mnmkng, line 218: You're not waiting for the navigation to complete here. Not sure if playwright has some magic to do this, but in Puppeteer you always needed to do:


```js
await Promise.all([
  page.waitForNavigation(),
  page.click('button')
])

await page.title()
```

mnmkng (changes requested):
I like it. The pace is slow enough for beginners. I wouldn't be afraid to add more details and talk more about the various features and differences between the libraries.

content/academy/puppeteer_playwright.md
```diff
@@ -0,0 +1,57 @@
+---
+title: Puppeteer/Playwright course
+description: description
+menuWeight: 7
+category: courses
+paths:
+    - puppeteer-playwright
+---
+
+# [](#puppeteer-playwright-course) Puppeteer/Playwright course
```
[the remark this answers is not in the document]
mstephen19: @mnmkng In my opinion that's confusing and not as good for SEO as **Puppeteer/Playwright** course. **Puppeteer** and **Playwright** definitely have more search frequency than **Headless browser**, and also we aren't really teaching them in depth about headless browsers, we're teaching them in depth about two specific libraries which utilize them.

content/academy/puppeteer_playwright/page.md
mstephen19: You're right!

content/academy/puppeteer_playwright/page.md
mstephen19: now repeating them in comments

content/academy/puppeteer_playwright/page.md
mstephen19: @mnmkng In the previous section we're waiting for the button's selector (instead of waiting for navigation)

mstephen19: The Puppeteer code works as is

content/academy/puppeteer_playwright.md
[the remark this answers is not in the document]
mstephen19: @m-murasovs Opinion?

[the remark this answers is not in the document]
mnmkng: Sure. I guess I just don't like the slash in the title 😄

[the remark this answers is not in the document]
m-murasovs: How about `Headless browser course with Puppeteer and Playwright`?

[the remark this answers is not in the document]
mnmkng: Try fitting that into the left menu 😄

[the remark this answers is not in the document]
mstephen19: That's huge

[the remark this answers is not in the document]
m-murasovs: > That's huge
eyyyyy 🎉 

If this is an SEO issue, why not just put `headless browser course` into the menu and then mention `puppeteer and playwright` in the meta description? Should still count, no? Especially since we can also set the H1 to the "huge" version, which will have all the keywords

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -0,0 +1,311 @@
...
+                const scrollHeight = document.body.scrollHeight;
+
+                window.scrollBy(0, distance);
+                totalHeight += distance;
+
+                if (totalHeight >= scrollHeight - window.innerHeight) {
+                    clearInterval(interval);
+                    resolve();
+                }
+            }, 200);
```
mnmkng, line 41: Isn't this a bit too complex to showcase data collection?

content/academy/puppeteer_playwright/executing_scripts.md
```diff
@@ -0,0 +1,155 @@
...
+title: III - Executing scripts
+description: Understand the two different contexts which your code can be run, and how to run custom scripts in the context of the browser.
+menuWeight: 7.3
+paths:
+    - puppeteer-playwright/executing-scripts
+---
+
+# [](#executing-scripts) Executing scripts
+
+An important concept to understand when dealing with headless browsers is the **context** in which your code is being run. For example, if you try to use the native `fs` Node.js module (used in the previous lesson) while running code in the context of the browser, errors will be thrown saying that it is undefined. Similarly, if you are trying to use `document.querySelector()` or other browser-specific functions in the server-side Node.js context, errors will also be thrown.
```
mnmkng, line 11: This is super important and a source of a lot of confusion. Maybe we could reuse some of the content [from here](https://help.apify.com/en/articles/3195646-when-to-use-puppeteer-scraper) to explain it even more.

content/academy/puppeteer_playwright/executing_scripts.md
```diff
@@ -0,0 +1,155 @@
...
+
+Here is an example of a common mistake made by beginners to Puppeteer/Playwright:
+
+```JavaScript
+// THIS IS INCORRECT CODE
+import { chromium } from 'playwright';
+
+const sleep = (secs) => new Promise((r) => setTimeout(r, secs * 1000));
```
mnmkng, line 19: Get rid of `sleep` everywhere and just use `page.waitForTimeout()` It's a nice opportunity to remind the users of the waitFor functions

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -0,0 +1,311 @@
...
+
+// code will go here
+
+await sleep(10);
+
+await browser.close();
+</marked-tab>
+```
+
+Notice that we've written and are using a custom `autoScroll()` function, which will scroll to the very bottom of the page. We are doing this because the images of the products are lazy-loaded (they are only loaded once they've been scrolled into view).
```
mnmkng, line 105: Instead of slowly scrolling to the bottom, maybe we could insta-scroll there? It would greatly simplify the function. Or the images wouldn't load this way?

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -0,0 +1,311 @@
...
+
+```text
+Error: Evaluation failed: Error: oops
+    at __puppeteer_evaluation_script__:6:19
+    at Array.map (<anonymous>)
+    at __puppeteer_evaluation_script__:2:75
+    ...
+```
+
+So, just because our function is failing to collect one or two of the products, it's failing to collect all of them and completely crashing our program. The solution for this is to parse the data in the Node.js context instead.
```
mnmkng, line 174: I think this is a bit unfair reason to do the parsing in cheerio. You can handle the errors with a simple try catch. There are other reasons to do it the cheerio way:

- you can easily port the code between headless browsers and plain http scraping
- you don't have to worry in which context you're working
- I don't know, I'm sure something else 😄 

I would use this opportunity to
1. first do it the evaluate + document.querySelector way
2. then show them how to inject jQuery into the browser and use it in there
3. finally show them that the easiest way is probably to extract the whole HTML and do it the cheerio way in Node

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -0,0 +1,311 @@
...
+
+const $ = load(await page.content());
+
+// collection code will go here
+
+await browser.close();
+</marked-tab>
+```
+
+Now, to loop through all of the products, we'll make use of the `$` object and loop through them while safely in the server-side context rather than running the code in the browser.
```
mnmkng, line 281: I fail to understand how the server side context is safer than the browser context. An unhandled exception in any of them will crash the program :)

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -0,0 +1,311 @@
...
+
+```JavaScript
+const $ = load(await page.content());
+
+const products = [...$('a[class*="ProductCard_root"]')].map((elem) => {
+    const card = $(elem);
+
+    const name = card.find('h3[class*="ProductCard_name"]').text() || null;
+    const price = card.find('div[class*="ProductCard_price"]').text() || null;
```
mnmkng, line 290: do we need the `|| null` everywhere? It's seems a bit redundant.

mnmkng (changes requested):
Thanks, great content, as always. Just some nitpicks, additions and structure suggestions, as always 😄

Also, I think you missed all the collapsed comments from the previous review.

content/academy/puppeteer_playwright/collecting_data.md
mstephen19: @mnmkng The other option is to not collect the image URLs, or at least not collect all of them.

content/academy/puppeteer_playwright/executing_scripts.md
mstephen19: I really like the diagram there, so I'll make a new one that fits our case

content/academy/puppeteer_playwright/collecting_data.md
mstephen19: Otherwise it'll be `undefined`, which isn't really ideal but I'll delete it

mnmkng: I would be interested to hear why `null` is better than `undefined` 🙂

content/academy/puppeteer_playwright/executing_scripts.md
mstephen19: Working on the diagram

PR #357, opened 2022-05-05
Puppeteer/Playwright course (fixed branch name)

mstephen19:


mstephen19:
See preview [here](https://developers.apify.com/academy/web-scraping-for-beginners?version=puppeteer-playwright-course)

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -0,0 +1,284 @@
...
+```marked-tabs
+<marked-tab header="Playwright" lang="javascript">
+import { chromium } from 'playwright';
+
+const browser = await chromium.launch({ headless: false });
+const page = await browser.newPage();
+
+await page.goto('https://demo-webstore.apify.org/search/on-sale');
+
+await page.waitForLoadState('networkidle');
```
mnmkng, line 35: You can add this as an option to `page.goto`. Also, is this really needed? The best practice is to wait for the page to `load` or sometimes even ``domcontentloaded` to speed up loading and then wait for the specific elements you need separately. This should also be part of the "Waiting for content" lesson.

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -0,0 +1,284 @@
...
+```
+
+> Notice the slightly different syntax between Playwright and Puppeteer with `waitForLoadState('networkidle')` and `waitForNetworkIdle()`. Sometimes, the differences are fairly subtle like this, but later on we'll run into some more significant differences.
+
+## [](#collecting-in-page-evaluate) Collecting in the browser context
+
+Whatever is returned from the callback function in `page.evaluate()` will be returned by the evaluate function, which means that we can set it to a variable like so:
+
+```JavaScript
+const products = await page.evaluate(() => {});
```
mnmkng, line 68: I'm not sure an empty function is explanatory enough
```suggestion
const products = await page.evaluate(() => { foo: 'bar' });
console.log(products); // { foo: 'bar' }
```

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -0,0 +1,284 @@
...
+```
+
+We'll be returning a bunch of product objects from this function, which will be accessible back in our Node.js context after the promise has resolved. Let's now go ahead and write some data collection code to collect each product:
+
+```JavaScript
+const products = await page.evaluate(() => {
+    return [...document.querySelectorAll('a[class*="ProductCard_root"]')].map((elem) => {
```
mnmkng, line 75: It's a tutorial. Rather create more variables and use expressive code like `Array.from` than trying to leetcode it. The example becomes squashed and the map is hard to read this way because it's too far right. People might be coming to this from different language backgrounds.
CleanShot 2022-05-06 at 17 08 13@2x
.

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -0,0 +1,284 @@
...
+console.log(products);
+```
+
+When we run this code, we see this logged to our console:
+
+![Products logged to the console]({{@asset puppeteer_playwright/images/log-products.webp}})
+
+## [](#using-jquery) Using jQuery
+
+With the `page.addScriptTag()` function and the latest [jQuery CDN link](https://releases.jquery.com/), we can inject jQuery into the current page:
```
mnmkng, line 95: Thanks for adding it, but it's missing the rationale behind it. Something like "working with document.querySelector is cumbersome bla bla, but we can inject the good old jQuery that simplifies document manipulation bla bla"

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -0,0 +1,284 @@
...
+await page.addScriptTag({ url: 'https://code.jquery.com/jquery-3.6.0.min.js' });
+```
+
+Now, since we're able to use jQuery, let's translate our vanilla JavaScript code within the `page.evaluate()` function to jQuery:
+
+```JavaScript
+await page.addScriptTag({ url: 'https://code.jquery.com/jquery-3.6.0.min.js' });
+
+const products = await page.evaluate(() => {
+    return [...$('a[class*="ProductCard_root"]')].map((elem) => {
```
mnmkng, line 107: This is more readable thanks to being shorter, but I think jQuery has a native method that returns an array, which I think would be more readable than the `return [...spread].map(). A lot of things happening on one line.

content/academy/puppeteer_playwright/collecting_data.md
mnmkng: A short explanation of the addScriptTag function would also be nice. We're introducing it for the first time after all.

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -0,0 +1,284 @@
...
+const products = [...$('a[class*="ProductCard_root"]')].map((elem) => {
+    const card = $(elem);
+
+    const name = card.find('h3[class*="ProductCard_name"]').text();
+    const price = card.find('div[class*="ProductCard_price"]').text();
+
+    return {
+        name,
+        price,
+        image: `https://demo-webstore.apify.org${image}`,
```
mnmkng, line 206: leftover code i guess?

mnmkng (changes requested):
I know this is not the final, polished version, but it would be great to have it a bit more "clean" before you send it for review next time. I know that it's a pain in the ass. I also wrote many tutorials and I know that the moment you send it for review, everyone finds a million things that are not clear or should be explained more, but here I felt you were taking too many shortcuts and were not taking the opportunity to explain things on a deeper level.

With that said, it's really great content anyway. So keep rocking.

content/academy/puppeteer_playwright.md
```diff
@@ -0,0 +1,61 @@
...
+Additionally, since the requests aren't static, [dynamic content]({{@link dynamic_pages_and_spas.md}}) can be rendered and interacted with (or, data from the dynamic content can be scraped).
+
+## [](#setup) Setup
+
+For this course, we'll be jumping right into the features of these awesome libraries and expecting you to already have an environment set up. Here's how we set up our environment:
+
+1. Make sure you've installed [Node.js](https://nodejs.org/en/)
+2. Create a new folder called **puppeteer-playwright** (or whatever you want to call it)
+3. Run the command `npm init -y` within your new folder to automatically initialize the project
+4. Create a new file named **index.js**
```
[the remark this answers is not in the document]
mstephen19: Definitely will add this on the to-do list for concepts. Yeah, I noticed that you were using esm but didn't say anything about it so I figured not to do it. Adding it to the **Setup** part for now, but later will write the concept lesson about it

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,309 @@
...
+
+const browser = await puppeteer.launch({ headless: false });
+
+const page = await browser.newPage();
+
+await browser.close();
+</marked-tab>
+```
+
+Then, we can visit a website with the `page.goto()` method. Let's go to [Google](https://google.com) for now. We'll also use the `page.waitForTimeout()` function, which will force the program to wait for a number of seconds before moving quitting (otherwise, everything will just flash before our eyes and we won't really tell what's going on):
```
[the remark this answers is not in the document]
mstephen19: Small nitpick with the **be able to** but I'll change it :D

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,309 @@
...
+
+Though it seems complex, the wonderful **Page** API makes all of these actions extremely easy to perform.
+
+### Clicking & pressing keys
+
+Let's first focus on the first 3 steps listed above. By using `page.click()` and the CSS selector of the element to click, we can click an element:
+
+```JavaScript
+// Click the "I understand" button
+await page.click('button + button:nth-child(2)');
```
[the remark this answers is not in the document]
mstephen19: Because Puppeteer doesn't support this selector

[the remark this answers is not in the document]
mstephen19: I'll split the examples to highlight yet another difference between the two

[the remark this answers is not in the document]
mnmkng: That's true, but one of the main benefits of the combined course is to showcase the differences. It should not skip Playwright features when they're not available in Puppeteer. Quite the contrary. It should point all the differences out and explain them.

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,309 @@
+---
+title: II - Page
+description: Learn how to create a Page object on a Browser object, and how to use it to visit and programmatically interact with a website.
```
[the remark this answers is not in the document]
mstephen19: Now, **Page object** is only said 2 times in the lesson

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,309 @@
...
+</marked-tab>
+<marked-tab header="Puppeteer" lang="javascript">
+// Wait for the first result element to be visible
+await page.waitForSelector('.g a');
+</marked-tab>
+```
+
+> If we were to omit the `page.waitForSelector()` before `page.click()` in Puppeteer, an error would be thrown.
+
+### Getting general page data
```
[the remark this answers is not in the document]
mstephen19: Will be doing this in the next PR.

content/academy/puppeteer_playwright/page.md
```diff
@@ -0,0 +1,309 @@
...
+
+When we run it, we leave off on the results page:
+
+![Google results page reached by headless browser]({{@asset puppeteer_playwright/images/google-results.webp}})
+
+### Waiting for elements/events
+
+In a perfect world, every piece of content served on a website would be loaded instantly. We don't live in a perfect world though, and often times it can take anywhere between 1/10th of a second to a few seconds to load some content onto a page. Puppeteer and Playwright don't sit around waiting for the pages to load though - if we tell it to do something with an element that hasn't rendered yet, it'll start trying to do it (which will result in nasty errors). We've got to tell it to wait.
+
+Different events can be waited for using the various `waitFor...` methods offered, while elements can be waited for with `page.waitForSelector()` in both Playwright and Puppeteer. In Playwright you can also create a locator for an element and use the `locator.waitFor()` function. Luckily, we'll only have to use `page.waitForSelector()` before clicking the first result in our Puppeteer code, because Playwright automatically waits for the element to appear before clicking.
```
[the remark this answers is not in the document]
mstephen19: Since we've already written about this in **Dynamic pages & SPAs** and provided some examples, is it okay to just link to there?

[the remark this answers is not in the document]
mnmkng: I tried ctrl+F-ing for wait in there and haven't really found any references to all the puppeteer and playwright functions that we should explain, so no, I don't think so.

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -48,7 +48,7 @@ const page = await browser.newPage();
 
 await page.goto('https://demo-webstore.apify.org/search/on-sale');
 
-await page.waitForNetworkIdle();
+await page.waitForNavigation({ waitUntil: 'DOMContentLoaded' });
```
mnmkng, line 51: This isn't correct for multiple reasons.

- You don't need to `waitForNavigation` after `page.goto`. Especially not for `domcontentloaded`, because `goto` waits for `load` event by default, which fires after `domcontentloaded`.
   - You only need to wait for navigation BEFORE you click something which triggers a navigation and then await the promise after the click. It's all in the documentation examples I sent you.
- You can put the `waitUntil` directly in the `goto` function. No need for extra function.
- It's `domcontentloaded` not `DOMContentLoaded`. Maybe it works as well, but in the docs its lowercased.

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -32,7 +32,7 @@ const page = await browser.newPage();
 
 await page.goto('https://demo-webstore.apify.org/search/on-sale');
 
-await page.waitForLoadState('networkidle');
+await page.waitForLoadState('DOMContentLoaded');
```
mnmkng, line 35: Similar as the other comment a few lines below.

content/academy/puppeteer_playwright/collecting_data.md
mnmkng: Also, if we're changing the default load eventg from `load` to `domcontentloaded` we should explain why we're doing that.

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -58,23 +58,27 @@ await browser.close();
 </marked-tab>
 ```
 
-> Notice the slightly different syntax between Playwright and Puppeteer with `waitForLoadState('networkidle')` and `waitForNetworkIdle()`. Sometimes, the differences are fairly subtle like this, but later on we'll run into some more significant differences.
+> Notice the slightly different syntax between Playwright and Puppeteer with `waitForLoadState('DOMContentLoaded')` and `waitForNavigation({ waitUntil: 'DOMContentLoaded' })`. Sometimes, the differences are fairly subtle like this, but later on we'll run into some more significant differences.
```
mnmkng, line 61: This is probably redundant because `page.goto({ waitUntil: ...})` is the same in both.

content/academy/puppeteer_playwright/collecting_data.md
```diff
@@ -92,29 +96,33 @@ When we run this code, we see this logged to our console:
...
-            price,
-        };
-    });
+    const name = card.find('h3[class*="ProductCard_name"]').text();
+    const price = card.find('div[class*="ProductCard_price"]').text();
+
+    return {
+        name,
+        price,
+    };
```
mnmkng, line 124: this seems to have wrong indentation

mnmkng (changes requested):

content/academy/puppeteer_playwright/collecting_data.md
mstephen19: The fact that `load` is already waited for after `goto` is fired cleared this up

content/academy/puppeteer_playwright/page/interacting_with_a_page.md
```diff
@@ -0,0 +1,171 @@
...
+```marked-tabs
+<marked-tab header="Playwright" lang="JavaScript">
+// Click the "I agree" button
+await page.click('button:has-text("I agree")');
+<marked-tab>
+<marked-tab header="Puppeteer" lang="JavaScript">
+// Click the "I agree" button
+await page.click('button + button:nth-child(2)');
+<marked-tab>
+```
```
mnmkng, line 38:
```suggestion
```marked-tabs
<marked-tab header="Playwright" lang="JavaScript">
// Click the "I agree" button
await page.click('button:has-text("I agree")');
</marked-tab>
<marked-tab header="Puppeteer" lang="JavaScript">
// Click the "I agree" button
await page.click('button + button:nth-child(2)');
</marked-tab>
```
```

File: content/academy/puppeteer_playwright/page/interacting_with_a_page.md [hunk 18]
```diff
@@ -0,0 +1,170 @@
...
+</marked-tab>
+
+> Notice that in the Playwright example, we are using a different selector than in the Puppeteer example. This is because Playwright supports [many custom CSS selectors](https://playwright.dev/docs/selectors#text-selector), such as the **has-text** pseudo class.
+
+Then, we can type some text into an input field with `page.type()`; passing a CSS selector as the first, and the string to input as the second parameter:
+
+```JavaScript
+// Type the query into the search box
+await page.type('input[title="Search"]', 'hello world');
+```
```
mnmkng, line 46: this isn't rendered as well

content/academy/puppeteer_playwright/page/interacting_with_a_page.md
```diff
@@ -0,0 +1,170 @@
...
+await page.keyboard.press('Enter');
+
+await page.waitForTimeout(10000)
+await browser.close();
+</marked-tab>
+```
+
+When we run it, we leave off on the results page:
+
+![Google results page reached by headless browser]({{@asset puppeteer_playwright/page/images/google-results.webp}})
```
mnmkng, line 104: The screenshot is still Czech

content/academy/puppeteer_playwright/page/interacting_with_a_page.md
```diff
@@ -0,0 +1,170 @@
...
+
+Great! So now all we have to do is click the first result, which matches the selector `.g a`:
```
mnmkng, line 106:
```suggestion
Great! So now all we have to do is click the first result, which matches the CSS selector `.g a`:
```

mnmkng: I would explain that it will click the first element found with a `.g a` selector. Because there are many elements like that on the pag
CleanShot 2022-05-14 at 22 25 29@2x
e

content/academy/puppeteer_playwright/page/interacting_with_a_page.md
```diff
@@ -0,0 +1,170 @@
...
+// Click the "I agree" button
+await page.click('button:has-text("I agree")');
+</marked-tab>
+<marked-tab header="Puppeteer" lang="JavaScript">
+// Click the "I agree" button
+await page.click('button + button:nth-child(2)');
+</marked-tab>
+
+> Notice that in the Playwright example, we are using a different selector than in the Puppeteer example. This is because Playwright supports [many custom CSS selectors](https://playwright.dev/docs/selectors#text-selector), such as the **has-text** pseudo class.
+
```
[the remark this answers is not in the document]
mnmkng: It also makes sense to tell the people that it actually drags the mouse and clicks, instead of doing it programmatically like `element.click()` does, which helps the script be more human-like.

content/academy/puppeteer_playwright/page/interacting_with_a_page.md
```diff
@@ -0,0 +1,170 @@
...
+Let's first focus on the first 3 steps listed above. By using `page.click()` and the CSS selector of the element to click, we can click an element:
+
+```marked-tabs
+<marked-tab header="Playwright" lang="JavaScript">
+// Click the "I agree" button
+await page.click('button:has-text("I agree")');
+</marked-tab>
+<marked-tab header="Puppeteer" lang="JavaScript">
+// Click the "I agree" button
+await page.click('button + button:nth-child(2)');
```
mnmkng, line 36: The selector can be simplified to `button + button`

content/academy/puppeteer_playwright/page/interacting_with_a_page.md
```diff
@@ -0,0 +1,170 @@
...
+
+// Click the first result
+await page.click('.g a');
+
+await page.waitForTimeout(10000)
+await browser.close();
+</marked-tab>
+```
+
+But wait, when we try to run the Puppeteer code, we run into this nasty error:
```
mnmkng, line 154: I like that. Showing a problem and then continuing with a solution is a great teaching style.

mnmkng (changes requested):
I'll review the rest of the pages later. The `waiting` page isn't loading for me. Will try to find the problem.

content/academy/puppeteer_playwright/page/waiting.md
```diff
@@ -0,0 +1,136 @@
...
+If we remember properly, after clicking the first result, we want to console log the title of the result's page and save a screenshot into the filesystem. In order to grab a solid screenshot of the loaded page though, we should **wait for navigation**  before snapping the image. This can be done with `page.waitForNavigation()`.
+
+Naively, you might immediately think that this is the way we should wait for navigation after clicking the first result:
+
+```JavaScript
+await page.click('.g a');
+await page.waitForNavigation();
+```
+
+Though in theory this is correct, it can result in a race condition in which the page navigates quickly before the `page.waitForNavigation()` function is ever run, which means that once it is finally called, it will hang and wait forever for the `domcontentloaded` event to fire even though it already fired. To solve this, we can stick the waiting logic and the clicking logic into a `Promise.all()` call (placing `page.waitForNavigation()` first).
```
mnmkng, line 59: The default is load.

```suggestion
Though in theory this is correct, it can result in a race condition in which the page navigates quickly before the `page.waitForNavigation()` function is ever run, which means that once it is finally called, it will hang and wait forever for the [`load` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event) to fire even though it already fired. To solve this, we can stick the waiting logic and the clicking logic into a `Promise.all()` call (placing `page.waitForNavigation()` first).
```

content/academy/puppeteer_playwright/page/waiting.md
```diff
@@ -0,0 +1,136 @@
...
+```JavaScript
+await Promise.all([page.waitForNavigation(), page.click('.g a')]);
+```
+
+This implementation will do the following:
+
+1. Begin waiting for the page to navigate without blocking the `page.click()` function
+2. Click the element, firing off a navigating event
+3. Resolve once the page has navigated, allowing further code to run
+
```
mnmkng, line 70: I checked again and there's [page.waitForLoadState](https://playwright.dev/docs/api/class-frame#frame-wait-for-load-state) in Playwright. 

I would keep the paragraphs above that mention waitForNavigation, because it works the same in both Puppeteer and Playwright. Then I would add a paragraph and a short (not full runnable) example here that for Playwright we recommend using waitForLoadState because you don't have to use that ugly promise.all because the function will correctly resolve even if the event has already fired.

content/academy/puppeteer_playwright/page/waiting.md
```diff
@@ -0,0 +1,136 @@
...
+await page.click('button + button:nth-child(2)');
+
+// Type the query and visit the results page
+await page.type('input[title="Search"]', 'hello world');
+await page.keyboard.press('Enter');
+
+// Wait for the first result to appear on the page,
+// then click on it
+await page.waitForSelector('.g a');
+await Promise.all([page.waitForNavigation(), page.click('.g a')]);
```
mnmkng, line 96: Would replace with waitForLoadState. Yeah, I know, sorry that I didn't mention this earlier, but I didn't know about this function before. But it's better, so let's use it.

content/academy/puppeteer_playwright/page/page_methods.md
```diff
@@ -0,0 +1,130 @@
...
+title: Page methods
+description: Understand that the Page object has many different methods to offer, and learn how to use two of them to capture a page's title and take a screenshot. 
+menuWeight: 3
+paths:
+    - puppeteer-playwright/page/page-methods
+---
+
+# [](#page-methods) Page methods
+
+Other than having methods for interacting with a page and waiting for events and elements, the **Page** object also supports many various methods for doing other things, such as [reloading](https://pptr.dev/#?product=Puppeteer&version=v13.7.0&show=api-pagereloadoptions), [screenshotting](https://playwright.dev/docs/api/class-page#page-screenshot), [changing headers](https://playwright.dev/docs/api/class-page#page-set-extra-http-headers), and collecting the [page's content](https://pptr.dev/#?product=Puppeteer&version=v13.7.0&show=api-pagecontent).
```
mnmkng, line 11:
```suggestion
Other than having methods for interacting with a page and waiting for events and elements, the **Page** object also supports various methods for doing other things, such as [reloading](https://pptr.dev/#?product=Puppeteer&version=v13.7.0&show=api-pagereloadoptions), [screenshotting](https://playwright.dev/docs/api/class-page#page-screenshot), [changing headers](https://playwright.dev/docs/api/class-page#page-set-extra-http-headers), and collecting the [page's content](https://pptr.dev/#?product=Puppeteer&version=v13.7.0&show=api-pagecontent).
```

content/academy/puppeteer_playwright/page/page_methods.md
```diff
@@ -0,0 +1,130 @@
...
+
+// Write the image to the filesystem
+await fs.writeFile('screenshot.png', screenshot);
+```
+
+> The image will by default be **.png**. To change the image to **.jpeg** type, set the (optional) `type` option to **jpeg**.
+
+## [](#final-code) Final code
+
+Here's our final code:
```
mnmkng, line 51:
```suggestion
Here's our final code which extracts the page's title, takes a screenshot and saves it to our project's folder as `screenshot.png`:
```

content/academy/puppeteer_playwright/page/page_methods.md
```diff
@@ -0,0 +1,130 @@
...
+
+```marked-tabs
+<marked-tab header="Playwright" lang="javascript">
+import puppeteer from 'puppeteer';
```
mnmkng, line 55: The examples are mixed up. And in the previous lesson as well. Please fix it by swapping the code, not the titles. Playwright should always be the first example.

content/academy/puppeteer_playwright/page/page_methods.md
```diff
@@ -0,0 +1,130 @@
...
+</marked-tab>
+```
+
+When you run this code, you should see this logged to console:
+
+```text
+"Hello, World!" program - Wikipedia
+```
+
+Additionally, you should see a new image named **screenshot.png** that looks something like this:
```
mnmkng, line 124:
```suggestion
Additionally, you should see a new image named **screenshot.png** in your project's folder that looks something like this:
```

content/academy/puppeteer_playwright/executing_scripts.md
```diff
@@ -0,0 +1,143 @@
+---
+title: III - Executing scripts
+description: Understand the two different contexts which your code can be run, and how to run custom scripts in the context of the browser.
```
mnmkng, line 3:
```suggestion
description: Understand the two different contexts which your code can be run in, and how to run custom scripts in the context of the browser.
```

content/academy/puppeteer_playwright/executing_scripts.md
```diff
@@ -0,0 +1,143 @@
...
+menuWeight: 7.3
+paths:
+    - puppeteer-playwright/executing-scripts
+---
+
+# [](#executing-scripts) Executing scripts
+
+An important concept to understand when dealing with headless browsers is the **context** in which your code is being run. For example, if you try to use the native `fs` Node.js module (used in the previous lesson) while running code in the context of the browser, errors will be thrown saying that it is undefined. Similarly, if you are trying to use `document.querySelector()` or other browser-specific functions in the server-side Node.js context, errors will also be thrown.
```
mnmkng, line 11: Would be nice to have the diagram here.

content/academy/puppeteer_playwright/executing_scripts.md
```diff
@@ -0,0 +1,143 @@
...
+
+![Google with the background color changed to green]({{@asset puppeteer_playwright/images/green-google.webp}})
+
+## [](#using-variables-in-page-evaluate) Using variables in `page.evaluate()`
+
+Within our code, we generate a `randomString` in the Node.js context:
+
+```JavaScript
+const randomString = Math.random().toString(36).slice(2);
+```
```
mnmkng, line 90: How about just passing a different color instead of the random string? To simplify it and connect it with the previous example. Say something like "but what if the information about the color we want is only available in the Node.js context? We can pass a variable...."

content/academy/puppeteer_playwright/executing_scripts.md
```diff
@@ -0,0 +1,143 @@
...
+</marked-tab>
+```
+
+Now, when we run this code, we can see the title change on the page's tab:
+
+![Google with the background color changed to green]({{@asset puppeteer_playwright/images/title-changed.webp}})
+
+## [](#next) Next up
+
+Next, we'll be learning a bit about how to collect data using Playwright/Puppeteer. There are two main ways to do this, so [next exciting lesson]({{@link puppeteer_playwright/collecting_data.md}}) about both of them!
```
mnmkng, line 141:
```suggestion
Next, we'll be learning a bit about how to collect data using Playwright/Puppeteer. There are two main ways to do this, so [next exciting lesson]({{@link puppeteer_playwright/collecting_data.md}}) will be about both of them!
```

content/academy/puppeteer_playwright/executing_scripts.md
```diff
@@ -0,0 +1,143 @@
...
+
+await browser.close();
+</marked-tab>
+```
+
+Now, when we run this code, we can see the title change on the page's tab:
+
+![Google with the background color changed to green]({{@asset puppeteer_playwright/images/title-changed.webp}})
+
+## [](#next) Next up
```
mnmkng, line 139: This lesson is really short. How about adding a section about [`page.exposeFunction`](https://playwright.dev/docs/api/class-page#page-expose-function) to it?

I would first show a similar error as before. Trying to run a Node.js function from the browser context and failing. And then using exposeFunction to make it runnable.

Then I would talk about the limits of crossing the browser-node barrier. That only serializable objects can cross it and that you can't send full class instances back and forth.

mnmkng: Thinking about it, I think that `executing_scripts` should not be a lesson, but a section and it should have multiple lessons about the various eval functions, as well as `evaluateOnNewDocument` vs `addInitScript`. `addScriptTag` (and some Puppeteer alternative) should be discussed too. It can be somehow intertwined with the "Collecting data" lesson.

content/academy/puppeteer_playwright/reading_intercepting_requests.md
```diff
@@ -0,0 +1,308 @@
...
+menuWeight: 7.5
+paths:
+    - puppeteer-playwright/reading-intercepting-requests
+---
+
+# [](#reading-intercepting-requests) Reading & intercepting requests
+
+On any website that serves up images, makes [XMLHttpRequests](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest), or fetches content in some other way, you can see those requests (and their responses) in the [**Network** tab]({{@link api_scraping/general_api_scraping/locating_and_learning.md}}) of your browser's DevTools. Lots of data about the request can be found there, such as the headers, payload, and response body.
+
+In Playwright and Puppeteer, it is also possible to read (and even intercept) requests being made on the page after its initial load - programmatically. This is very useful for things like reading dynamic headers, saving API responses, blocking certain resources, and much more.
```
mnmkng, line 13: You can read/intercept even the initial request.

```suggestion
In Playwright and Puppeteer, it is also possible to read (and even intercept) requests being made on the page - programmatically. This is very useful for things like reading dynamic headers, saving API responses, blocking certain resources, and much more.
```

content/academy/puppeteer_playwright/reading_intercepting_requests.md
```diff
@@ -0,0 +1,308 @@
...
+    console.log('Request was made!')
+});
+</marked-tab>
+<marked-tab header="Puppeteer" lang="javascript">
+// Listen for all requests
+page.on('request', (req) => {
+    // If the URL doesn't include our keyword, ignore it
+    if (!req.url().includes('followings')) return;
+
+    console.log('Request was made!')
```
mnmkng, line 72:
```suggestion
    console.log('Request for followers was made!')
```

content/academy/puppeteer_playwright/reading_intercepting_requests.md
```diff
@@ -0,0 +1,308 @@
...
+});
+</marked-tab>
+```
+
+> Note that you should always define any request reading/interception code prior to calling the `page.goto()` function.
+
+Cool! So now when we run our code, we'll see this logged to the console:
+
+```text
+Request was made!
```
mnmkng, line 82:
```suggestion
Request for followers was made!
```

content/academy/puppeteer_playwright/reading_intercepting_requests.md
```diff
@@ -0,0 +1,308 @@
...
+Let's go ahead and listen for this request in our code:
+
+```marked-tabs
+<marked-tab header="Playwright" lang="javascript">
+// Listen for all requests
+page.on('request', (req) => {
+    // If the URL doesn't include our keyword, ignore it
+    if (!req.url().includes('followings')) return;
+
+    console.log('Request was made!')
```
mnmkng, line 63:
```suggestion
    console.log('Request for followers was made!')
```

content/academy/puppeteer_playwright/reading_intercepting_requests.md
```diff
@@ -0,0 +1,308 @@
...
+  app_version: '1652347025',
+  app_locale: 'en'
+}
+```
+
+## [](#reading-responses) Reading responses
+
+Listening for and reading responses is very similar to reading requests. The only difference is that we need to listen for the **response** event instead of **request**. Additionally, the object passed into the callback function represents the response instead of the request.
+
+<!-- To access the response object from **Route** in Playwright, we need to wait for it with `await route.request().response()`. Nothing else needs to change. -->
```
mnmkng, line 153: old comment?

content/academy/puppeteer_playwright/reading_intercepting_requests.md
```diff
@@ -0,0 +1,308 @@
...
+</marked-tab>
+```
+
+Upon running this code, we'll see the API response logged into the console:
+
+![API response in console]({{@asset puppeteer_playwright/images/api-response-tiesto.webp}})
+
+## [](#intercepting-requests) Intercepting requests
+
+One of the most popular ways of speeding up website loading in Puppeteer and Playwright is by blocking certain resources from loading. These resources are usually CSS files, images, and other miscellanous resources that aren't super necessary (mainly because the computer doesn't have eyes - it doesn't care how the website looks!).
```
mnmkng, line 186: Im totally missing the part about cache being turned off by turning on interception and how that's dangerous.

For blocking requests, there's [a completely different way of doing that that we have in the SDK](https://github.com/apify/apify-js/blob/55e0cce99dc7bfc5c303bb76885cdaf037ff6cd7/src/puppeteer_utils.js#L187) which does not turn off cache. It would be nice to mention the CDP method here as the correct way for blocking requests when performance is important. Also pls mention that it only works in Chromium based browsers.

mnmkng (changes requested):
Next batch of comments. Will finish proxies tomorrow.

content/academy/puppeteer_playwright/page/interacting_with_a_page.md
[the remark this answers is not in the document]
mstephen19: Found the issue on why it wasn't in the rendered page.

content/academy/puppeteer_playwright/proxies.md
```diff
@@ -0,0 +1,180 @@
...
+const browser = await puppeteer.launch({ headless: false });
+const page = await browser.newPage();
+await page.goto('https://google.com');
+
+await page.waitForTimeout(10000);
+await browser.close();
+</marked-tab>
+```
+
+For both Puppeteer and Playwright, the proxy's server URL should be passed into the options of the `launch()` function; however, it's done a bit differently depending on which library you're using.
```
mnmkng, line 50:
```suggestion
For both Puppeteer and Playwright, the proxy server's URL should be passed into the options of the `launch()` function; however, it's done a bit differently depending on which library you're using.
```

content/academy/puppeteer_playwright/proxies.md
```diff
@@ -0,0 +1,180 @@
...
+First, let's go ahead and create a variable called `proxy` which will point to our proxy server:
+
+> Note that this proxy may no longer be working at the time of reading. If you don't have a proxy to use during this lesson, we recommend using Proxy Scraper for a list of free ones, or checking out [Apify proxy](https://apify.com/proxy)
+
+```marked-tabs
+<marked-tab header="Playwright" lang="javascript">
+import { chromium } from 'playwright';
+
+// our proxy server
+const proxy = '103.214.9.13:3128';
```
mnmkng, line 26: I can't decide if this is confusing or not. We define the variable, but we don't use it anywhere in this example. Maybe we should show only the one line with the variable definition as the code example?

content/academy/puppeteer_playwright/proxies.md
```diff
@@ -0,0 +1,180 @@
...
+await browser.close();
+</marked-tab>
+<marked-tab header="Puppeteer" lang="javascript">
+import puppeteer from 'puppeteer';
+
+const proxy = '103.214.9.13:3128';
+
+// Using the "args" option, which is an array of Chromium command
+// line switches, we pass the server URL in with "--proxy-server"
+const browser = await puppeteer.launch({ headless: false, args: [`--proxy-server=${proxy}`] });
```
mnmkng, line 82: Multi-lining it, because the important param was partially cut off and one needed to scroll to se it.

```suggestion
const browser = await puppeteer.launch({
    headless: false,
    args: [`--proxy-server=${proxy}`]
});
```

mnmkng (changes requested):
Final batch of comments. Good job man.

content/academy/puppeteer_playwright/executing_scripts.md
mstephen19: I considered doing that, but since it's our first time showing that we can execute scripts, I really wanted to show them a few examples of what they can do. Change the background color, change the title - now they understand that they can do pretty much anything.

content/academy/puppeteer_playwright/proxies.md
mstephen19: I am showing them the boilerplate code for the rest of the lesson + the variable declaration at the same time

content/academy/puppeteer_playwright/executing_scripts.md
mnmkng: Yeah, my point was mostly around the random number generation. We can change the title to `Apify is Awesome` and it will work the same without adding the complexity of generating a random string.

content/academy/puppeteer_playwright/page/interacting_with_a_page.md
```diff
@@ -0,0 +1,177 @@
...
+await page.click('button + button');
+</marked-tab>
+```
+
+With `page.click()`, Puppeteer and Playwright actually drag the mouse and click, allowing the bot to act more human-like. This is different from programmatically clicking with `Element.click()` in vanilla client-side JavaScript.
+
+> If you're not already familiar with CSS selectors and how to find them, we recommend referring to [this lesson]({{@link web_scraping_for_beginners/data_collection/using_devtools.md}}) in the **Web scraping for beginners** course.
+
+Notice that in the Playwright example, we are using a different selector than in the Puppeteer example. This is because Playwright supports [many custom CSS selectors](https://playwright.dev/docs/selectors#text-selector), such as the **has-text** pseudo class. As a rule of thumb, using text selectors is much more preferable to using regular selectors, as they are much less likely to break. If Google makese the sibling above the **I agree** button a `<div>` element instead of a `<button>` element, our `button + button` selector will break. However, the button will always have the text **I agree**; therefore, `button:has-text("I agree")` is more reliable.
```
mnmkng, line 44:
```suggestion
Notice that in the Playwright example, we are using a different selector than in the Puppeteer example. This is because Playwright supports [many custom CSS selectors](https://playwright.dev/docs/selectors#text-selector), such as the **has-text** pseudo class. As a rule of thumb, using text selectors is much more preferable to using regular selectors, as they are much less likely to break. If Google makes the sibling above the **I agree** button a `<div>` element instead of a `<button>` element, our `button + button` selector will break. However, the button will always have the text **I agree**; therefore, `button:has-text("I agree")` is more reliable.
```

content/academy/puppeteer_playwright/reading_intercepting_requests.md
```diff
@@ -0,0 +1,352 @@
...
+```marked-tabs
+<marked-tab header="Playwright" lang="javascript">
+// Notice that the callback function is now async
+page.on('response', async (res) => {
+    if (!res.request().url().includes('followings')) return;
+
+    // Grab the response body in JSON format
+    const json = await res.json();
+    console.log(json);
+});
```
mnmkng, line 164: The res.json() can throw if the response is not JSON, causing an unhandled promise rejection. Having unhandled errors or rejections in the page handlers is quite a common error of devs, so I think it would make sense to add a try catch and use that opportunity to explain why that's important.

content/academy/puppeteer_playwright/reading_intercepting_requests.md
```diff
@@ -0,0 +1,352 @@
...
+
+Here's what we see when we run this logic:
+
+![SoundCloud with no CSS or image resources loaded]({{@asset puppeteer_playwright/images/ugly-soundcloud.webp}})
+
+This confirms that we've successfully blocked the CSS and image resources from loading.
+
+#### [](#quick-note) Quick note about resource blocking
+
+Something **very** important to note is that by using request interception, the browser's cache is turned **off**. This means that resources on websites that would normally be cached (and pulled from the cache instead on the next request for those resources) will not be cached, which can have varying negative effects on performance, especially when making multiple requests to the same page.
```
mnmkng, line 252:
```suggestion
Something **very** important to note is that by using request interception, the browser's cache is turned **off**. This means that resources on websites that would normally be cached (and pulled from the cache instead on the next request for those resources) will not be cached, which can have varying negative effects on performance, especially when making many requests to the same domain, which is very common in web scraping.
```

content/academy/puppeteer_playwright/executing_scripts/injecting_code.md
```diff
@@ -0,0 +1,109 @@
...
+In the previous lesson, we learned how to execute code on the page using `page.evaluate()`, and though this fits the majority of use cases, there are still some more unusual cases. For example, what if we want to execute our custom script prior to the page's load? Or, what if we want to define a function in the page's context to be run at a later time?
+
+We'll be covering both of these cases in this brief lesson.
+
+## [](#pre-injecting) Pre-injecting scripts
+
+Sometimes, you need your custom code to run before any other code is run on the page. Perhaps you need to modify an object's prototype, or even re-define certain global variables before they are used by the page's native scripts.
+
+Luckily, Puppeteer and Playwright both have functions for this. In Puppeteer, we use  the [`page.evaluateOnNewDocument()`](https://puppeteer.github.io/puppeteer/docs/puppeteer.page.evaluateonnewdocument/) function, while in Playwright we use [`page.addInitScript()`](https://playwright.dev/docs/api/class-page#page-add-init-script). We'll use these functions to override the native `addEventListener` function, setting it to `null`. This will prevent event listeners from being added to elements.
```
mnmkng, line 19:
```suggestion
Luckily, Puppeteer and Playwright both have functions for this. In Puppeteer, we use  the [`page.evaluateOnNewDocument()`](https://puppeteer.github.io/puppeteer/docs/puppeteer.page.evaluateonnewdocument/) function, while in Playwright we use [`page.addInitScript()`](https://playwright.dev/docs/api/class-page#page-add-init-script). We'll use these functions to override the native `addEventListener` function, setting it to a function that does nothing. This will prevent event listeners from being added to elements.
```

content/academy/puppeteer_playwright/executing_scripts/injecting_code.md
```diff
@@ -0,0 +1,109 @@
...
+await browser.close();
+</marked-tab>
+<marked-tab header="Puppeteer" lang="javascript">
+import puppeteer from 'puppeteer';
+
+const browser = await puppeteer.launch({ headless: false });
+const page = await browser.newPage();
+await page.goto('https://google.com');
+
+const returnMessage = () => 'Hello world!';
```
mnmkng, line 94:
```suggestion
const returnMessage = () => 'Apify academy!';
```

content/academy/puppeteer_playwright/executing_scripts/injecting_code.md
```diff
@@ -0,0 +1,109 @@
...
+
+```marked-tabs
+<marked-tab header="Playwright" lang="javascript">
+import { chromium } from 'playwright';
+
+const browser = await chromium.launch({ headless: false });
+const page = await browser.newPage();
+await page.goto('https://google.com');
+
+const returnMessage = () => 'Hello world!';
```
mnmkng, line 76:
```suggestion
const returnMessage = () => 'Apify academy!';
```

mnmkng (approved):

PR #359, opened 2022-05-05
Write 'Custom queries'

mstephen19:


metalwarrior665 (changes requested):
As discussed, looks good but I would use `got-scraping` so we stay with the one and true HTTP client :) (also works the same in CheerioCrawler) and I would mention `.graphlql` files for people that will not want to install the query language package.

metalwarrior665 (approved):

PR #368, opened 2022-05-18
Update deploying section

mstephen19:


mstephen19:
View preview [here](https://developers.apify.com/academy/apify-platform/deploying-your-code?version=update-deploying#our-example-project)

content/academy/apify_platform/deploying_your_code.md
```diff
@@ -12,7 +12,13 @@ This section will discuss how to use your newfound knowledge of the Apify platfo
 
 Because actors are basically just chunks of code running in Docker containers, you're able to **_Actorify_** just about anything!
 
-Actors are language agnostic, which means that the language your project is written in does not affect your ability to actorify it. Though the majority of actors currently on the platform were written in Node.js, there are a few examples in other languages:
+![The deployment workflow]({{@asset apify_platform/images/deployment-workflow.webp}})
+
+Actors are language agnostic, which means that the language your project is written in does not affect your ability to actorify it.
+
+![Supported languages]({{@asset apify_platform/images/supported-languages.webp}})
```
mnmkng, line 19: Can we make this image smaller? It's a bit overwhelming.

mnmkng (changes requested):

mstephen19:
Image size is now 70% of what it originally was.

mnmkng:
> Image size is now 70% of what it originally was.
It's still the same. Full screen.
CleanShot 2022-05-19 at 09 49 44@2x

mstephen19:
> I understand what you meant by size. The KB of the image. Well, that's also important. I notice the deployment image is pretty huge and the languages are not small either.
> 
> The way to do this is to reduce the PNG resolution first - this will not only reduce it's KBs, but also it's size on the page. Then you can try to make it even smaller in KBs with tinypng and if that does not work, then there's the WEBP conversion script in the docs build process.

What I did was take the original PNG, decrease the resolution, export it as JPG, then put it through JPG compression. I'll come up with a better solution for this

mnmkng (approved):

PR #373, opened 2022-05-19
Add overview file check for folders

m-murasovs:
A safety check that makes sure new contributors are less likely to break docs. 
It checks that each folder (with exceptions) has an overview file that's needed for the menu to work properly.

Screenshot 2022-05-19 at 12 04 44


mstephen19:
Really nice!

mtrunkat (approved):

PR #376, opened 2022-05-19
Add NPM links to Client docs

m-murasovs:
GitHub is already linked in the `Edit` button, so I didn't add this.


m-murasovs:
This doesn't fix it since we don't use these metadata on web

PR #380, opened 2022-05-31
Small edits

mstephen19:


metalwarrior665 (changes requested):
There is also

https://developers.apify.com/academy/web-scraping-for-beginners/data-collection/node-continued#querying-with-cheerio
console.log(element.textContent);
does not work. Should be .text() instead of .textContent

metalwarrior665 (approved):
Thanks, let;s merge. Not sure about the failing checks here.

PR #381, opened 2022-06-01
Update storage.md Node version requirement to 14.19.3

martinxluptak:


martinxluptak:
10.17 didn't work because of Node 14 syntax sugar `??`.

I only tried v14.19.3 and would not recommend any other version than v14.19.3 because [this is the oldest supported LTS](https://nodejs.org/en/about/releases/)

Issue #320, opened 2022-01-24
Improve webhooks docs


m-murasovs:
It appears the docs are out of date and could do with improvement. A couple of people on Discord were confused, as reported by Paulo. Conversation: https://apifier.slack.com/archives/CQ96RHG2U/p1642786975000300

Also check: https://apifier.slack.com/archives/DUC52PG0J/p1647352424354999

Check out Discord to see what they need and check with Platform team how we can improve.

Probably move all the content to one page?

Also see: https://apifier.slack.com/archives/CQ96RHG2U/p1669482412125069

We need to document retries and exponential backoff, this is not described anywhere and users are surprised by this ([Slack](https://apifier.slack.com/archives/C03NSCJ9X47/p1693324765494209)).

fnesveda:
I was reading through the docs now, and I think the issues that people complained about are fixed:
- actions are listed
- variable interpolation is explained
- retries and exponential backoff are there
- the structure is better after Mára's restructuring

I will just close this, if it's needed we can reopen it.

PR #387, opened 2022-06-06
Fix link, add info about resource available to webhook actions

mvolfik:
cc #320 (though this is far from a complete fix)


mstephen19 (changes requested):
There seems to be a build conflict tho

mvolfik:
I have no idea why the action failed - it says it's missing some token, which I doubt could be affected by changing some text in docs :shrug:

PR #388, opened 2022-06-07
Fix links

mvolfik:
Grepped for `obscurity` and `#/store`


mstephen19 (approved):
Looks good. Don't forget to request review from Misha or me after making a PR. Thanks for fixing this! 😄

PR #391, opened 2022-06-09
Puppeteer/Playwright: Common use cases section + two new lessons

mstephen19:


mstephen19:
Preview [here](https://developers.apify.com/academy/puppeteer-playwright/common-use-cases/logging-into-a-website?version=puppeteer-playwright)

mstephen19:
Ready for review

content/academy/puppeteer_playwright/common_use_cases/logging_into_a_website.md
```diff
@@ -0,0 +1,396 @@
...
+        await sendEmailContext.close();
+    })()
+);
+
+await Promise.all(promises);
+
+await browser.close();
+</marked-tab>
+```
+
```
metalwarrior665, line 393: 1. Not sure what you agreed with this lesson but I would also add Crawlee code. That is what 90% of our users will be using so they need to be able to almost copy/paste that.
2. Mention that we can also pass in cookies to the actor and only inject them
(if you think these do not belong here, we should add them elsewhere)

metalwarrior665 (changes requested):

content/academy/puppeteer_playwright/common_use_cases/logging_into_a_website.md
mstephen19: Screenshotting your comment + adding it to backlog.

PR #392, opened 2022-06-10
Create switching to Typescript section

mstephen19:


mstephen19:
Preview [here](https://developers.apify.com/academy/anti-scraping?version=switching-to-typescript)

content/academy/switching_to_typescript.md
```diff
@@ -0,0 +1,63 @@
...
+};
+
+const addAges = (num1, num2) => num1 + num2;
+
+console.log(addAges(bob.age, john.age));
+```
+
+This code doesn't actually throw an error, but it does output `27undefined`. That's not good. The first issue is that `john.age` is **undefined**, and the second issue is that `bob.age` is a string and must be converted to a number to work properly in the `addAges` function. Despite these two significant mistakes, JavaScript doesn't tell us at all about them and lets the code run with bugs.
+
+With TypeScript, these types of issues stick out like a sore thumb, and depending on your configurations, the compiler will refuse to compile it until they have been fixed.
```
metalwarrior665, line 41: Add a link explaining what is a compiler

content/academy/switching_to_typescript/unknown_and_type_casting.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+## [](#final-thoughts) Final thoughts
+
+Even though you've learned about them in the same lesson, type guards and type casting are inherently very different things with different use cases.
+
+**Type guards** make a runtime check of whether or not a value passes a check that determines that it can be safely used as a certain type. They are great when dealing with values that might hold inconsistent data types (such as user input) where you aren't 100% sure if a certain property will exist.
+
+**Type casting** tells TypeScript to take a value of one type and to treat it as if it were another type. No runtime checks are made. A common use case is casting the response body of an API call (usually has the `any` type depending on what you're using to fetch the data) to a custom type to receive TypeScript support on the data.
+
```
metalwarrior665, line 98: Would mention here also the `!` token which lets you force the type if you know it is there but the compiler does not.

metalwarrior665 (commented):
Nice one, will there be a convert a JS Apify project tutorial? That is a must I think :)

content/academy/switching_to_typescript/unknown_and_type_casting.md
B4nan: Worth noting that the term "type casting" should be in fact ["type assertions"](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions), there is no casting happening, and it does not work with "values" at all. We all use the term "casting" as it feels familiar (and I would indeed mention it here too, but not as a headline), but it does not cast any value, it changes the type. Let's use the right terms so its easier to find appropriate docs for it.

And agreed we should mention the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-) too.

mnmkng:
@B4nan did you get a chance to review? Or are you prioritizing Crawlee release? That's fine, just say so and we'll launch it and can improve it later.

content/academy/switching_to_typescript/unknown_and_type_casting.md
mstephen19: both resolved

mstephen19:
> @B4nan did you get a chance to review? Or are you prioritizing Crawlee release? That's fine, just say so and we'll launch it and can improve it later.

@mnmkng He added a small review comment, along with @metalwarrior665. I made the fixes. Do you think it's ok to merge?

B4nan (commented):
here are few more small comments, i dont have the time to do full review now unfortunately, will try to get back to it but hard to say when, i need to focus on crawlee release now

content/academy/switching_to_typescript/enums.md
```diff
@@ -0,0 +1,125 @@
+---
+title: Enums
+description: Learn how to easily define, use, and manage constant values using this epic feature called "enums" that TypeScript brings to the table.
```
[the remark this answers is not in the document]
mstephen19: Though tbh I fucking love enums haha

content/academy/switching_to_typescript.md
```diff
@@ -0,0 +1,63 @@
...
+
+![]({{@asset switching_to_typescript/images/typescript-error.webp}})
+
+This means that when using TS (a popular acronym for "TypeScript") on a large project, you'll run into much less runtime errors and catch the majority of them during the development process.
+
+## [](#advantages-of-typescript) What are the advantages of using TypeScript?
+
+1. The ability to **optionally** [statically type](https://developer.mozilla.org/en-US/docs/Glossary/Static_typing) your variables and functions.
+2. [Type Inference](https://www.typescriptlang.org/docs/handbook/type-inference.html), which provides you the benefits of using types, but without having to actually statically type anything. For example, if you create a variable like this: `const num = 5`, TypeScript will automatically infer that `num` is of a **number** type.
```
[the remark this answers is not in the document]
mstephen19: Good catch. I switched it to `let`

PR #394, opened 2022-06-14
Add note on using environment variables during build

valekjo:


content/docs/actors/development/source_code.md
```diff
@@ -135,6 +135,8 @@ console.log(process.env.SMTP_HOST);
 
 The actor runtime sets additional environment variables for the actor process during the run. See [Environment variables]({{@link actors/development/environment_variables.md}}) for details.
 
+The environment variables can also be used for the build process. In this case, the variables are treated as [Docker build arguments](https://docs.docker.com/engine/reference/builder/#arg). Meaning that they should not be used for secrets and in order to access them in Dockerfile, you have to use `ARG variable_name` instruction.
```
m-murasovs, line 138:
```suggestion
The environment variables can also be used for the build process. In this case, the variables are treated as [Docker build arguments](https://docs.docker.com/engine/reference/builder/#arg). This means that they should not be used for secrets and, in order to access them in Dockerfile, you have to use the `ARG variable_name` instruction.
```

m-murasovs (changes requested):

m-murasovs (approved):

PR #403, opened 2022-07-18
feat: Add max concurrent actor runs limit

jirimoravcik:
The problem is that the limit is only 25 for free users and also it's not in `@apify/consts` package like the others, maybe we should add it there?


mtrunkat (changes requested):
y, we should move this into a constant

mtrunkat (approved):

PR #421, opened 2022-08-08
Fix headings

PerVillalva:
Remove "strong" element from the "Price and trial length" section headings and replace it with proper heading priority.


PerVillalva:
Hey Matt. Hanz mentioned that a few headings in this section had the "strong" element, which is not ideal. I fixed the issue with this PR. It's a very minor change, but I assigned you as a reviewer just to ensure that everything is fine :) Thanks!

mstephen19 (approved):

PR #423, opened 2022-08-09
fix: Try catch block

mstephen19:
Bringing value to the try catch block


metalwarrior665 (changes requested):
There are more of these on this page and maybe also on other pages. Can you just remove those try/catches completely? I think it adds unnecessary clutter at this stage. These calls should almost never fail. 

You want to try/catch only blocks that have reasonable chance to fail

PR #425, opened 2022-08-11
docs: update imports add Crawlee, Actor import, etc

AndreyBykov:
I am pretty sure you will not go through all changes here, but for the record:
- ~~`got` (not got-scraping, pure `got` examples don't work with `GOOGLE_SERP` - no idea why.~~
- Storage pages - let's quicky go through them tomorrow, I am a bit lost, should we use Crawlee storages or Actor storages there?
- snippets should be up-to-date, all imports updated to ESM.
- ~~TODO: should double-check snippet comments~~
- TODO: should go throught text/links (and here I am also not 100% where should be imports from what package :/ )

### DIFF
--- a/content/docs/actors/development/state_persistence.md
+++ b/content/docs/actors/development/state_persistence.md
@@ -43,8 +43,10 @@ The [Apify SDK](https://sdk.apify.com) persists its state automatically, using t
 To persist state manually, you can use the [Apify.events](https://docs.apify.com/actors/development/state-persistence) method in the Apify SDK.
 
 ```javascript
-Apify.events.on('migrating', () => {
-    Apify.setValue('my-crawling-state', {
+import { Actor } from 'apify';
+
+Actor.on('migrating', () => {
+    Actor.setValue('my-crawling-state', {
         foo: 'bar',
     });
 });
@@ -53,7 +55,9 @@ Apify.events.on('migrating', () => {
 To check for state saved in a previous run, use:
 
 ```javascript
-const previousCrawlingState = await Apify.getValue('my-crawling-state') || {};
+import { Actor } from 'apify';
+
+const previousCrawlingState = await Actor.getValue('my-crawling-state') || {};
 ```
 
 To improve your actor's performance, you can also [cache repeated page data]({{@link tutorials/cache_data_to_improve_performance.md}}).
--- a/content/docs/actors/development/state_persistence.md
+++ b/content/docs/actors/development/state_persistence.md
@@ -45,11 +45,15 @@ To persist state manually, you can use the [Apify.events](https://docs.apify.com
 ```javascript
 import { Actor } from 'apify';
 
+await Actor.init();
+// ...
 Actor.on('migrating', () => {
     Actor.setValue('my-crawling-state', {
         foo: 'bar',
     });
 });
+// ...
+await Actor.exit();
 ```
 
 To check for state saved in a previous run, use:
@@ -57,7 +61,11 @@ To check for state saved in a previous run, use:
 ```javascript
 import { Actor } from 'apify';
 
+await Actor.init();
+// ...
 const previousCrawlingState = await Actor.getValue('my-crawling-state') || {};
+// ...
+await Actor.exit();
 ```
 
 To improve your actor's performance, you can also [cache repeated page data]({{@link tutorials/cache_data_to_improve_performance.md}}).
--- a/content/docs/proxy/google_serp_proxy/examples.md
+++ b/content/docs/proxy/google_serp_proxy/examples.md
@@ -165,27 +165,6 @@ console.log(data);
 </marked-tab>
 
 
-<marked-tab header="Node.js (got)" lang="javascript">
-import got from 'got';
-import { HttpsProxyAgent } from 'hpagent';
-
-const agent = {
-    https: new HttpsProxyAgent({
-        // Replace <YOUR_PROXY_PASSWORD> below with your password
-        // found at https://console.apify.com/proxy
-        proxy: 'http://groups-GOOGLE_SERP:<YOUR_PROXY_PASSWORD>@proxy.apify.com:8000',
-    }),
-};
-
-const url = 'http://www.google.com/search';
-const searchParams = { q: 'wikipedia' };
-
-const { body } = await got(url, { agent, searchParams });
-
-console.log(body);
-</marked-tab>
-
-
 <marked-tab header="Python 3" lang="python">
 import urllib.request as request
 import urllib.parse as parse
@@ -289,27 +268,6 @@ const params = { q: 'Apple iPhone XS 64GB', tbm: 'shop' }
 const { data } = await axios.get(url, { proxy, params });
 
 console.log(data);
-</marked-tab>
-
-
-<marked-tab header="Node.js (got)" lang="javascript">
-import got from 'got';
-import { HttpsProxyAgent } from 'hpagent';
-
-const agent = {
-    https: new HttpsProxyAgent({
-        // Replace <YOUR_PROXY_PASSWORD> below with your password
-        // found at https://console.apify.com/proxy
-        proxy: 'http://groups-GOOGLE_SERP:<YOUR_PROXY_PASSWORD>@proxy.apify.com:8000',
-    }),
-};
-
-const url = 'http://www.google.co.uk/search';
-const searchParams = { q: 'Apple iPhone XS 64GB', tbm: 'shop' };
-
-const { body } = await got(url, { agent, searchParams });
-
-console.log(body);
 </marked-tab>
 
 
--- a/content/docs/proxy/google_serp_proxy/examples.md
+++ b/content/docs/proxy/google_serp_proxy/examples.md
@@ -27,47 +27,48 @@ Get a list of search results for the keyword **wikipedia** from the USA (`google
 
 ```marked-tabs
 <marked-tab header="CheerioCrawler" lang="javascript">
-const Apify = require('apify');
+import { Actor } from 'apify';
+import { CheerioCrawler } from 'crawlee';
 
-Apify.main(async() => {
-    const proxyConfiguration = await Apify.createProxyConfiguration({
-        groups: ['GOOGLE_SERP'],
-    });
-    const url = 'http://www.google.co.uk/search?q=wikipedia';
+await Actor.init();
 
-    const requestList = await Apify.openRequestList('my-list', [url]);
-
-    const crawler = new Apify.CheerioCrawler({
-        requestList,
-        proxyConfiguration,
-        handlePageFunction: async ({ request, response, body }) => {
-            // ...
-            console.log(body)
-        },
-    });
-
-    await crawler.run();
+const proxyConfiguration = await Actor.createProxyConfiguration({
+    groups: ['GOOGLE_SERP'],
 });
+
+const crawler = new CheerioCrawler({
+    proxyConfiguration,
+    async requestHandler({ body }) {
+        // ...
+        console.log(body)
+    },
+});
+
+await crawler.run(['http://www.google.co.uk/search?q=wikipedia']);
+
+await Actor.exit();
 </marked-tab>
 
 
-<marked-tab header="requestAsBrowser()" lang="javascript">
-const Apify = require('apify');
+<marked-tab header="gotScraping()" lang="javascript">
+import { Actor } from 'apify';
+import { gotScraping } from 'got-scraping';
 
-Apify.main(async () => {
-    const proxyConfiguration = await Apify.createProxyConfiguration({
-        groups: ['GOOGLE_SERP'],
-    });
+await Actor.init();
 
-    const proxyUrl = proxyConfiguration.newUrl();
-
-    const { body } = await Apify.utils.requestAsBrowser({
-        url: 'http://www.google.com/search?q=wikipedia',
-        proxyUrl,
-     });
-
-    console.log(body);
+const proxyConfiguration = await Actor.createProxyConfiguration({
+    groups: ['GOOGLE_SERP'],
 });
+const proxyUrl = await proxyConfiguration.newUrl();
+
+const { body } = await gotScraping({
+    url: 'http://www.google.com/search?q=wikipedia',
+    proxyUrl,
+});
+
+console.log(body);
+
+await Actor.exit();
 </marked-tab>
 ```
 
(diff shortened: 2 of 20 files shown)


AndreyBykov:
But generally - there should be updated a loooooooooooot of stuff, I am not sure how deep I should dive in there now =D

B4nan:
cc @mnmkng @mstephen19 

> got (not got-scraping, pure got examples don't work with GOOGLE_SERP - no idea why.

cc @szmarczak

AndreyBykov:
Regarding got - e.g. axios works completely fine. With got - I tried the agent from old examples, from got docs, even the old tunnel - while axios return the correct US based results, got always just using local Czech IP. Either I am doing something wrong, or either there's a problem with got or agent 🤷‍♂️

content/docs/actors/development/state_persistence.md
```diff
@@ -43,8 +43,10 @@ The [Apify SDK](https://sdk.apify.com) persists its state automatically, using t
...
 ```javascript
-Apify.events.on('migrating', () => {
-    Apify.setValue('my-crawling-state', {
+import { Actor } from 'apify';
+
+Actor.on('migrating', () => {
+    Actor.setValue('my-crawling-state', {
         foo: 'bar',
     });
 });
```
mnmkng, line 52: Let's not worry about this now, but maybe we could add `useState` to the SDK as well? https://crawlee.dev/docs/upgrading/upgrading-to-v3#auto-saved-crawler-state

content/docs/proxy/google_serp_proxy/examples.md
```diff
@@ -142,56 +145,44 @@ Select this option by setting the `username` parameter to `groups-GOOGLE_SERP`.
...
+    }),
 };
 
-useProxy();
+const url = 'http://www.google.com/search';
+const searchParams = { q: 'wikipedia' };
+
+const { body } = await got(url, { agent, searchParams });
+
+console.log(body);
```
mnmkng, line 185: I would replace this with `got-scraping` so that we don't have to worry about got not working for now.

content/docs/proxy/google_serp_proxy/examples.md
```diff
@@ -281,61 +272,44 @@ Set the domain (your country of choice) in the URL (in the `response` variable).
...
-        searchParams: {
-          query: 'iApple iPhone XS 64GB',
-          tmb: 'shop',
-        },
-        agent: {
-            https: new HttpsProxyAgent(proxyUrl),
-        }
-    });
-    console.log(response.body);
+import got from 'got';
```
mnmkng, line 296: dtto

content/docs/storage/dataset.md
```diff
@@ -61,26 +61,26 @@ To add data to the default dataset, you can use the example below, however using
...
-    await Apify.pushData({ foo: 'bar' });
+// ...
+// Add one item to the default dataset
+await Actor.pushData({ foo: 'bar' });
 
-    // Add multiple items to the default dataset
-    await Apify.pushData([{ foo: 'hotel' }, { foo: 'cafe' }]);
-});
+// Add multiple items to the default dataset
+await Actor.pushData([{ foo: 'hotel' }, { foo: 'cafe' }]);
```
mnmkng, line 71: Missing `init` and `exit` ? I would probably include it everywhere to make sure people don't forget it as it might cause weird bugs.

content/docs/tutorials/analyze_pages_and_fix_errors.md
```diff
@@ -109,29 +109,36 @@ Snapshots can tell you if:
 
 In Apify scrapers (**Web Scraper** ([apify/web-scraper](https://apify.com/apify/web-scraper)), **Cheerio Scraper** ([apify/cheerio-scraper](https://apify.com/apify/cheerio-scraper)) and **Puppeteer Scraper** ([apify/puppeteer-scraper](https://apify.com/apify/puppeteer-scraper))), you can use their built-in `context.saveSnapshot()` function. Once called, it saves a screenshot and HTML into the run's **key-value store**.
 
-When **building your own actors** with [Puppeteer](https://pptr.dev) or the the [Apify SDK](https://sdk.apify.com) package, you can use the powerful `utils.puppeteer.saveSnapshot()` [function](https://sdk.apify.com/docs/api/puppeteer#puppeteersavesnapshot). It allows you name the screenshot, so you can identify it later.
+When **building your own actors** with [Puppeteer](https://pptr.dev) or the [Apify SDK](https://sdk.apify.com) package, you can use the powerful `utils.puppeteer.saveSnapshot()` [function](https://sdk.apify.com/docs/api/puppeteer#puppeteersavesnapshot). It allows you name the screenshot, so you can identify it later.
```
mnmkng, line 112: There aren't any utils.puppeteer in SDK v3 I believe so this is misleading. Needs to reference Crawlee

content/docs/tutorials/analyze_pages_and_fix_errors.md
```diff
@@ -109,29 +109,36 @@ Snapshots can tell you if:
...
 The most common approach is to save on error. We can enhance our previous try/catch block like this:
 
 ```javascript
+import { Actor } from 'apify';
+import { puppeteerUtils } from 'crawlee';
+
+// ...
 // storeId is ID of current key value store, where we save snapshots
-const storeId = Apify.getEnv().defaultKeyValueStoreId;
+const storeId = Actor.getEnv().defaultKeyValueStoreId;
```
mnmkng, line 133: Does get env still exist or is it preferred? I thought the preferred way to read this is now through Configuration.

content/docs/tutorials/cache_data_to_improve_performance.md
```diff
@@ -42,23 +42,24 @@ Because [all objects in JavaScript are just references](https://www.freecodecamp
...
+// If we don't, we just initialize the cache to an empty object
+const cache = (await Actor.getValue('CACHE')) || {};
+
+// Now, we set up the persistence. You can choose between 'migrating' and 'persistState' events
+// 'migrating' only saves on migration, so it is a little "cheaper"
+// 'persistState' is usually preferred, it will also help if you abort the actor
+Actor.on('persistState', async () => {
+    await Actor.setValue('CACHE', cache);
+});
+// We have secured the persistence and can now pass on the cache and use it like we want
```
mnmkng, line 60: `useState` could also be useful here

content/docs/tutorials/cache_data_to_improve_performance.md
```diff
@@ -107,85 +108,83 @@ Our cache will be an object where the **keys** will be the seller IDs (imagine a
...
-            } else if (label === 'SELLER') {
-                // And finally we handle the seller page
-                // We scrape the seller data
-                const sellerData = extractSellerData($);
-
-                // We populate the cache so we can access all of this seller's other products from there
-                cache[sellerData.sellerId] = sellerData;
-
-                // We merge seller and product data and push
+    async requestHandler({ request, $ }) {
```
mnmkng, line 132: Might be better with a router? But I guess we can update this later.

content/docs/tutorials/scrape_paginated_sites.md
```diff
@@ -257,13 +267,16 @@ const { min, max } = getFiltersFromUrl(request.url);
 const newFilters = splitFilter({ min: min * 100, max: max * 100 });
 
 // And we just enqueue those 2 new filters so the process will recursively repeat until all pages get to the PAGINATION phase
+const requestsToEnqueue = [];
 for (const filter of newFilters) {
-    await requestQueue.addRequest({
+    await requestsToEnqueue.push({
```
mnmkng, line 272: extra await?

mnmkng (changes requested):

content/docs/proxy/google_serp_proxy/examples.md
AndreyBykov: Well - there's already `got-scraping` snippet came instead of `requestAsBrowser`, but we could ditch the `got` completely - `axios` is working fine, so it's `axios` for node, python, php?

content/docs/storage/dataset.md
AndreyBykov: I did originally, but it kinda increases the size of some snippets quite a lot - was not sure whether we want to do it even for some limited examples or not. Also - there were these comments:

// The optional Apify.main() function performs the
// actor's job and terminates the process when it is finished

Do we want to use Actor methods or Crawlee storages directly here? If it's gonna be `Actor` nearly everywhere - then well - maybe it's a good idea to keep init/exit everywhere

content/docs/tutorials/analyze_pages_and_fix_errors.md
AndreyBykov: I did not touch the references at all yet, this PR is about snippets only, looks like I just looked at this `the` and did not like it and just removed =D

content/docs/tutorials/analyze_pages_and_fix_errors.md
AndreyBykov: It exists, yeah, but I was not updating much logic, I just replaced the existing wrong usage of `Apify` class - properly updating all examples will take more time, I thought we're focusing on correctness of the existing stuff for now.

content/docs/tutorials/scrape_paginated_sites.md
AndreyBykov: good catch, I thought I removed it, missed this one

content/docs/proxy/google_serp_proxy/examples.md
mnmkng: Yeah, we don't need to showcase multiple Node http clients.

content/docs/storage/dataset.md
mnmkng: I would do it like this. If the example has an `import` statement, then it should have `init` and `exit` to make it runnable. If it's just some one liner or two liner showcasing a function, it does not need to have those. What do you think?

mnmkng: And yeah, I would use `Actor.` methods everywhere unless there's an absolute need to import Crawlee, because there's some crawling in the example.

content/docs/tutorials/scraping_dynamic_content.md
```diff
@@ -99,7 +99,8 @@ Now we simply pass it to `waitForFunction`:
 
 ```javascript
 // In Puppeteer you need to inject JQuery with
-// await Apify.utils.Puppeteer.injectJQuery(page);
+// await puppeteerUtils.injectJQuery(page);
+// imported from 'crawlee' package
```
B4nan, line 103: fyi we also have `utils.puppeteer.*` in the crawlee package

AndreyBykov: But it's imported only from meta-package, right? I mean - it would work here, but in the end people will opt for specific packages I guess. But I am not sure, I could keep the `utils.puppeteer` - it's kinda a matter of preference I'd say

content/docs/storage/dataset.md
AndreyBykov: Went through the changed places - kinda decided to add it everywhere - I guess it will not hurt to form this pattern - if you're using actor - use these two before/after your main code.

content/docs/proxy/google_serp_proxy/examples.md
```diff
@@ -292,27 +271,6 @@ console.log(data);
 </marked-tab>
 
 
-<marked-tab header="Node.js (got)" lang="javascript">
-import got from 'got';
-import { HttpsProxyAgent } from 'hpagent';
-
-const agent = {
-    https: new HttpsProxyAgent({
```
szmarczak: Have you noticed that there's `https` but you're querying `http` instead? :D

szmarczak (changes requested):

content/docs/proxy/google_serp_proxy/examples.md
AndreyBykov: yup, I saw the docs of the agent, and I tried all options, it would either throw or return local (CZ) results

content/docs/actors/development/state_persistence.md
B4nan: All what useState does is already available, this is the implementation:

```ts
async useState<State extends Dictionary = Dictionary>(defaultValue = {} as State): Promise<State> {
    const kvs = await KeyValueStore.open(null, { config: this.config });
    return kvs.getAutoSavedValue<State>(BasicCrawler.CRAWLEE_STATE_KEY, defaultValue);
}
```

So we can just expose `Actor.getAutoSavedValue` as a static helper with some default state key, or even have an `Actor.useState` for that, if it would use some static state key that might be more in line.

content/docs/tutorials/analyze_pages_and_fix_errors.md
B4nan: It is actually there, but only in the metapackage:

https://github.com/apify/crawlee/blob/master/packages/crawlee/src/index.ts#L18-L25

AndreyBykov:
@mnmkng what do you think about merging the snippets (changes already reviewed within this PR), and I'll continue with actual descriptions etc updates in another PR? I still plan to rather focus on just updating the references for now, replace actor SDK with Crawlee and so on. There are also some links to some old github repos (like hello world example or something like that using SDK 0.xx). Otherwise I would just continue pushing the updates here, what's your preference?

Also maybe @mstephen19 if you wanna have a look?

content/docs/proxy/google_serp_proxy/examples.md
szmarczak: Look what I found :joy:

image

AndreyBykov: I think I even tried the direct URL 🤔 will check it once again once I am back to my laptop 👌

AndreyBykov: So - I had `q` in my latest tries, but it does not matter - both should work I think. The problem is that proxy is ignored. It supposed to return US results, but returning Czech ones. It's with `https`, with `http` I have this:
Screenshot 2022-08-15 at 12 01 08

szmarczak: Can you try the code I posted?

AndreyBykov: welp, looks like it finally worked, wtf 🤯

AndreyBykov: however - it still does not work with the `hpagent` from `got` docs. Fails with `code: 'ERR_GOT_REQUEST_ERROR', literally the only thing changed is the `agent`.

mnmkng (approved):
@AndreyBykov yeah sure, if you think there are no obvious issues anymore, we can merge and deal with other issues later.

AndreyBykov:
@mnmkng I would merge it as this one covers the snippets only. I would rather re-iterate with the next PR with just text/links/descriptions so that it would be easier to review it at least. Also, when is https://apify.github.io/apify-sdk-js/ moving to https://sdk.apify.com/? So that I could put the correct links in descriptions?

@B4nan what do you think? Does not matter much for me - I could go with both options.

mstephen19 (approved):

B4nan:
> Also, when is [apify.github.io/apify-sdk-js](https://apify.github.io/apify-sdk-js/) moving to [sdk.apify.com](https://sdk.apify.com/)?

I'd like to move it later this week or early next week, definitely before launch. IMO me its ready, I can move it there on thursday when I'm back. One thing to note - we have only one algolia app for both v2 and v3, I'd use it for v3 and keep v2 docs snapshot without search - in the end all the content is in the new v3 docs too.

I am fine with merging if it makes sense to you.

content/docs/proxy/google_serp_proxy/examples.md
szmarczak: It doesn't work with `hpagent` because the proxy doesn't support `CONNECT` method: https://github.com/delvedor/hpagent/blob/cc90c2b60a0b946574993d5598d93229ce452f72/index.js#L18

```
RequestError: Bad response: 400
Google query failed: The proxy group "GOOGLESERP" cannot be used over HTTPS connection.
```

The protocol is `http:` but the superproxy mistakenly thinks it's HTTPS.

---

Basically there are two ways to proxy a request.

1. We can make a GET request like normal but include the full absolute URI:

```
GET http://example.com/ HTTP/1.1
Host: example.com
...
```

and receive a response from the proxy server.

2. We can use the `CONNECT` method. It's meant for setting up TCP over HTTP.

```
CONNECT example.com:80 HTTP/1.1
Host: example.com:80
...

[raw TCP data]
```

It's used to connect to other servers, they don't have to be HTTP.

---

I'd solve this in a superproxy this way. If there's a CONNECT method, set up a dummy server (doesn't have to listen on any port), and redirect request/responses through that one.

mnmkng: Not sure if this is resolved, but I think it would make sense not to have to get into this level of detail in the examples 😅

AndreyBykov: got examples are removed, I was just trying to understand what am I doing wrong here :) I don't plan to push it further. But if I would need it in real life - I would just use another package for agent, or even just use axios instead of got all along 🤷‍♂️

AndreyBykov:
Will continue with other PR.

content/docs/proxy/google_serp_proxy/examples.md
szmarczak: @mnmkng Since those agents are problematic, I think we could move the `proxyUrl` option to Got? Agents indeed may be too low-level for newbies.

szmarczak: For now you could recommend just using `got-scraping`.

szmarczak: Or we could implement CONNECT for SERP as [mentioned here](https://github.com/apify/apify-docs/pull/425#discussion_r946538384) - this would be much more user friendly.

AndreyBykov: I think we would not put `got` there, old docs had `requestAsBrowser`, which is now replaced with `got-scraping`, so we are kinda emphasizing it already, the rest is just illustrative - to basically show that you could use our proxies with other libraries too.

As for `proxyUrl` for `got` - personally, I think it's a good idea, because imho it's much user-friendly :)

PR #438, opened 2022-09-12
Output schema

milan-lepik:
link to doc page preview https://docs.apify.com/actors/development/output-schema?version=output-schema#specification-version-1


content/docs/actors/development/output_schema.md
```diff
@@ -0,0 +1,322 @@
...
+            }
+        }
+    }
+}
+```
+
+**B)** .actor/actor.json links to other sub-config files in the same folder, e.g.:
+
+```JSON
+//file: .actor/actor.json
```
mstephen19, line 49: Do we need these when you already state the file name above?

mstephen19 (approved):

content/docs/actors/development/output_schema.md
milan-lepik: I would rather keep it there for users that are copy-pasting code, to help them put the code in the correct file.

PR #443, opened 2022-09-19
Update the schedules section

### LINKED ISSUE #420: Rework schedules section
See: https://docs.apify.com/schedules#setting-up-a-new-schedule

- It does not reflect the new schedules UI from @valekjo. We should describe briefly how to use this and point out advanced users to the CRON configuration, which can be described on a separate page to keep it lean.
- We need to describe behaviour of input and options:
     - If the input is missing, then what input is used? Example or saved?
     - If the input is there, then it's merged with the default or saved input?
     - If options are null (can be set this way via a scheduler) then default run options are used.

m-murasovs:
Refreshes the schedules section to reflect the new UI

closes #420


content/docs/schedules.md
```diff
@@ -103,9 +105,17 @@ You can add multiple actor and task runs to a schedule with a single POST reques
 
 For more information, see the [schedules section](https://docs.apify.com/api/v2#/reference/schedules/schedule-object/get-schedule) in the API documentation.
 
+## Schedule setup
+
+The schedule setup tool uses [cron expressions](https://en.wikipedia.org/wiki/Cron#CRON_expression), which we parse using the [cron-parser](https://www.npmjs.com/package/cron-parser) package, to specify run times. If you're familiar with how to use them and need a specific run schedule, you can dive right in. If not, don't worry - the setup tool has a visual custom schedule builder that provides the same level of control cron expressions.
```
valekjo, line 110: - I'm not sure if I'd link the library directly (but it's ok if we do it elsewhere)
 - `same level of control cron expressions` => `same level of control as cron expressions` - but it's not exactly true, it's strong enough, but doesn't provide the same level of control

content/docs/schedules.md
```diff
@@ -103,9 +105,17 @@ You can add multiple actor and task runs to a schedule with a single POST reques
...
+
+![Schedule setup tool]({{@asset images/schedule-setup.webp}})
+
+You can find more information and examples of cron expressions on [crontab.guru](http://crontab.guru/). For additional and non-standard characters, see [this](https://en.wikipedia.org/wiki/Cron#CRON_expression) Wikipedia article.
```
valekjo, line 114: I'd mention the "live feedback" that gives you 5 (or so) next runs, so you can experiment with it.

valekjo (changes requested):

valekjo (approved):

PR #449, opened 2022-09-25
fix: missing labels import

molaycule:
since labels are not imported yet, this throws an error for the user when the code is copied directly


mstephen19 (approved):
thanks!

PR #457, opened 2022-10-12
academy/content-migration

PerVillalva:
@mstephen19 Can you please check if there aren't any structural components missing? As for the new Actor README section, I will ask Dave to review it.


PerVillalva (commented):
@davidjohnbarton can you please review the text in the newly added actor_readme.md file?
I migrated the content from the guidelines you wrote for Zuzka in Notion.

davidjohnbarton:
Looks fine to me. No changes needed.

package.json
```diff
@@ -32,6 +32,7 @@
         "@apify/log": "^1.1.5",
         "fs-extra": "^10.0.0",
         "markdown-yaml-metadata-parser": "^3.0.0",
+        "nvm": "^0.0.4",
```
mstephen19, line 35: You accidentally installed `nvm` as a dependency of the project rather than globally. Also, it's not recommended to use the NVM installed from NPM, as it's not the correct one (see [here](https://www.npmjs.com/package/nvm)). The real one lives on https://nvm.sh/

To fix this issue, run this command in your terminal within the project's directory:

```shell
npm uninstall nvm && curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.2/install.sh | bash
```

content/docs/tutorials/building_public_actors.md
```diff
@@ -11,9 +11,9 @@ paths:
 There are four main stages of building a public actor:
 
 1. [Development]({{@link actors/development.md}}).
-2. [Publication]({{@link actors/publishing.md}}).
+2. [Publication](../../academy/apify_platform/publishing_actors_on_apify_store.md).
```
mstephen19, line 14: Please use the `{@link}`s to link to other files instead of these absolute paths. Otherwise, things can (and will) break. For example, try clicking on **Publication** in the [deployment preview](https://docs.apify.com/tutorials/building-public-actors?version=academy-content-migration-publishing-actors) for this branch. It doesn't work, because the file path is different from the actual URL path that leads to the page you want.

content/academy/apify_platform/publishing_actors_on_apify_store/seo_and_promotion.md
```diff
@@ -88,16 +87,14 @@ The README․md file represents a complete description of your actor, intended f
...
+An important part of SEO is internal and external linking. Don’t be shy - add links to your README leading to similar [actors](https://apify.com/store) or any interesting article in [Apify Help](https://help.apify.com/) that could assist your users. If your actor is scraping a certain website, make sure to add a link to that website. If the actor uses any external libraries, make sure to link their pages.
 
 If you’ve created other actors, it’s completely fine to pop in a promo for them at the end. It’s even better if you edit your old actors and put in a link to your new actor. We’ll take care of extra promotion on the [Apify homepage](https://apify.com/) or [Apify Store](https://apify.com/store) - especially if you’ve done a great job of describing your actors!
 
 > **GOOD**: Like my <a href="#">Amazon Crawler</a>? See my <a href="#">Amazon Review Crawler</a>.
-<br/>
-> **AVOID**: Check out my Amazon Crawler.
+> <br/> **AVOID**: Check out my Amazon Crawler.
```
mstephen19, line 97: What's the `<br />` HTML tag doing here? It is generally bad practice to use HTML in Markdown unless there is a good reason for doing so.

mstephen19 (changes requested):
Thanks a ton Percy! Really great work. Just a few important technical issues, but the content looks good to me 👍

package.json
PerVillalva: Thanks for noticing it, Matt. I accidentally installed it locally while debugging the libjpeg error with Misha. I couldn't download packages globally because of some permission issues with my mac, but it is solved now. I will fix this in the next commit.

content/docs/tutorials/building_public_actors.md
PerVillalva: I had a hard time using `{@link}` to link files from the docs to the academy. For example, the build was returning an error for this path `({{@link academy/apify_platform/publishing_actors_on_apify_store.md}})`.

Could you please give me an example of how it should be done for this particular link so I can then update the others?

content/academy/apify_platform/publishing_actors_on_apify_store/seo_and_promotion.md
PerVillalva: The `<br/>` is ensuring the "AVOID" text stays on a new line without breaking the "card" style. If I simply used space, then this "card-look" would split in two. See the image attached for reference.

Screenshot 2022-10-14 at 11 33 15 AM

content/docs/tutorials/building_public_actors.md
mstephen19: You're right. The `build` script doesn't allow for using `../` to access the parent directory within `@link` tags because the **academy** and **docs** are built separately. So, you cannot safely reference docs articles from the academy, or academy lessons from the docs section. Because of this, I'd recommend just using URLs to the content instead of file paths.

content/academy/apify_platform/publishing_actors_on_apify_store/seo_and_promotion.md
mstephen19: Just add a return between the two quotes to make them not merge together instead of using `<br>`

mstephen19: Ok makes sense

mstephen19 (approved):

PR #473, opened 2022-11-16
Academy - Output schema guide

PerVillalva:
Migrating and updating content about the Output tab from Apify help (https://help.apify.com/en/articles/6660702-how-to-setup-output-tab-in-just-5-minutes).


content/academy/deploying_your_code/output_schema.md
```diff
@@ -0,0 +1,174 @@
...
+                            "SKU": {
+                                "label": "SKU",
+                                "format": "text"
+                            },
+                            "inStock": {
+                                "label": "In Stock",
+                                "format": "boolean"
+                            },
+                            "onStock": {
+                                "label": "On Sale",
```
mnmkng, line 129: `onStock` `On Sale`. One of them probably isn't correct :)

content/academy/deploying_your_code/output_schema.md
```diff
@@ -0,0 +1,174 @@
...
+                                "label": "Product image",
+                                "format": "image"
+                            },
+                            "url": {
+                                "label": "Link",
+                                "format": "link"
+                            },
+                            "brand": {
+                                "label": "Brand",
+                                "format": "text"
```
mnmkng, line 114: I'm not sure, but I think that you can skip the definitions if the only thing you do is capitalization of the label. It should be done automatically.

content/academy/deploying_your_code/output_schema.md
PerVillalva: Nice catch. It was supposed to be onSale 😄 Fixed it.

content/academy/deploying_your_code/output_schema.md
PerVillalva: I think it uses the variable name by default, but it wouldn't be capitalized.

mnmkng: Sorry for delay. I keep forgetting to check GitHub now that I don't code often. It does capitalize. It even splits camel case into separate words and capitalizes all of them. See the difference between first number and second number. First has no entry in output schema, second is manually entered.

CleanShot 2022-11-28 at 11 04 30@2x

PerVillalva: Thanks, Ondra. I've updated my actor code and the output schema to reflect the correct usage of the labels. I also added a short note explaining the schema's default behavior.

mnmkng (approved):

PR #486, opened 2022-12-27
Update pro_scraping.md

miladdotnetdeveloper:
Added two missing userData


B4nan:
those are not missing, `Request.label` is a shortcut

PR #500, opened 2023-02-06
docs: remove duplicate section

moazammalek:
IP session consistency section was repeated twice on the same page


B4nan (approved):
thanks!

PR #533, opened 2023-03-02
feat: academy subnavbar

### LINKED ISSUE #508: STRUCTURE: Add top level menu to the academy
The top-level menu bar will contain:
- Courses
- Tutorials
- Snippets - original Glossary

For **Courses**, we will just display the following in the left menu:

obrazek

For **Tutorials** we will have the following in the left menu:

obrazek

For **Snippets** we may need to rename the menu links to match the snippet's name. Or @PerVillalva @mnmkng, do you want to keep it as it is?

obrazek

barjin:
This is imo the solution of least resistance (and link breaking), but might be a bit convoluted. 

fixes #508 

<video src="https://user-images.githubusercontent.com/61918049/222437553-4f38f935-5b53-4e32-a126-868b6be300c5.mp4"/>


mtrunkat:
obrazek

barjin:
obrazek

It's your call, but with "Academy" in both navbars it feels a bit redundant to me... Maybe show no title in the subnav at all?

docusaurus.config.js
```diff
@@ -22,47 +23,31 @@ module.exports = {
...
+                    pathRegex: '/academy',
+                    items: [
+                        {
+                            label: 'Courses',
+                            to: `/academy`,
+                            activeBaseRegex: [
+                                'academy$',
+                                ...collectSlugs(`${__dirname}/sources/academy/webscraping`),
+                                ...collectSlugs(`${__dirname}/sources/academy/platform`),
+                            ].join('|'),
```
B4nan, line 37: i dont think we need this, the whole academy is under `academy` url prefix, same for platform docs

i think you can't have a doc inside that would escape this url prefix (defined by the plugin instance)

B4nan:
> Maybe show no title in the subnav at all?

Its important for the other projects, where you dont know where you are - as the top navbar button can be under a dropdown

docusaurus.config.js
B4nan: maybe we should change the URL structure first, before we introduce the categories? @mtrunkat 

so it would have common url prefix for each subnavbar section, to get around this craziness. it will work but it feels super ugly

barjin: huh, so you beat me to it :) i'll still post the explanation I had written, maybe it helps someone

> No yeah, that's exactly why we need that :) This is there to have the "active" class set on the current academy subcategory you are in. The slugs for the documents in the categories are unfortunately semi-random, e.g.:
> - tutorials: "/academy/tutorials", "/academy/node-js/\*", "/academy/python/\*"  etc.
> - courses: "/academy/web-scraping-for-beginners", "/academy/puppeteer-playwright/"...

barjin:
Right, sorry - maybe show no title in the subnav at all _(in this category)_? :) Not sure about the design though, it might feel a bit inconsistent when we make the title optional.

B4nan:
Yes, that would be quite confusing, moving with navigation elements this way seems like an antipattern UX-wise.

(I also don't like the current solution, but this does not feel much better)

m-murasovs:
Sorry, I was checking this out and pressed `Update branch` out of habit 😄

barjin:
This got a bit stale... what's the current status on this? Did we decide on the file structure / visuals already (CC @mtrunkat )?

docusaurus.config.js
B4nan: How does this work under the hood? Does it generate some huge client-side list somewhere in the DOM? I'd be a bit afraid of that, but I guess we could live with it as long as it works.

I guess we don't want to have the URL prefixes there, as we've added quite a lot of redirects that are doing the opposite.

barjin: No, AFAIR this constructs a gigaregex which is used for matching the URLs for the "active" class on the subnavbar... this is resolved during build, the resulting pages just contain the correct class on the correct button for the current page.

mtrunkat:
And from discussion with @barjin - maybe we don't need to be highlighting the main link in second level menu, it looks better tho

B4nan:
Maybe it could use a different style than the highlighted button, e.g. with less contrast or without a background. Or maybe we could remove it and keep the indent so the first button in first and second level nav starts on the same position?

barjin:
How about this?
## Old
obrazek
## New (proposal)
obrazek

it doesn't make much sense to me to have two links highlighted on the same level of a menu. In this proposal, all the items have slightly lower `font-weight` except for the subtitle. This would be site-wide, i.e. in all the other projects as well.

itstrangvu:
Yeah, instead of "Apify Docs", in the secondary level of the menu, it should be "Academy". Otherwise, I have nothing to it :)

itstrangvu:
> How about this?
> 
> ## Old
> obrazek
> 
> ## New (proposal)
> obrazek
> 
> it doesn't make much sense to me to have two links highlighted on the same level of a menu. In this proposal, all the items have slightly lower `font-weight` except for the subtitle. This would be site-wide, i.e. in all the other projects as well.

I'm fine with this proposal and approve this.

barjin:
Good lord, is this what frontend devs do all day long? 😛

obrazek

obrazek

B4nan (approved):
Great, I'd say this is ready to go, shall we merge @mtrunkat or is there any blocker?

> Good lord, is this what frontend devs do all day long? 😛

Yes, they are basically moving pixels :] jk

mtrunkat:
Let's get it out

PR #562, opened 2023-04-03
chore: Rename docs.yml to docs.yaml in the publish workflow, fix readme

jirimoravcik:
`docs.yml` has been renamed in all repos or is in PR with rename


fnesveda (approved):
Let's make sure we merge all the PRs at the same time, this one last.

PR #577, opened 2023-04-17
fix: platform/tutorials redirected to academy

barjin:


jbartadev (commented):
Ad @B4nan mentioned it'd be better to do the redirect on the nginx level

barjin:
> Ad @B4nan mentioned it'd be better to do the redirect on the nginx level

I have seen that (and wholeheartedly agree), but unfortunately, I don't have access there :)

@fnesveda I see you got tagged there too, do you have the nginx access?

jbartadev:
@barjin Already been solved by https://github.com/apify/apify-docs-v2-private/pull/30. Feel free to close the PR

PR #582, opened 2023-04-27
feat: Replaced all mentions about multiple source file with Web IDE

jbartadev:
@mtrunkat I think the API docs should be fine as it only explains that the "Actor source code is comprised of multiple files" which is still true and it stays like this. The Web IDE is only connected to the platform's user interface and I replaced all the mentioning of the `Multiple source files` type in this PR.

Closes [apify-web #2597](https://github.com/apify/apify-web/issues/2597)

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ The `./sources/platform/api_v2` directory contains the source file for the API r
 
 1. Install Apiary gem `gem install apiaryio`
 2. After that, you can open the generated doc with the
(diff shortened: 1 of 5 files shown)


dodaniel (approved):
Looks good to me

mtrunkat (approved):

mtrunkat:
Perhaps we could mention in the API that multiple source files as shown and editable in the Apify Console UI as web IDE. But the rest is fine.

B4nan (approved):

PR #601, opened 2023-05-26
docs: fix secret image url

novotnyj:
The image is `png` not `jpg`. Also other images are loaded using `require`, there is no `@asset` elsewhere in docs...

### DIFF
--- a/sources/platform/actors/development/secret_input.md
+++ b/sources/platform/actors/development/secret_input.md
@@ -36,7 +36,7 @@ To make an input field secret, just add a `"isSecret": true` setting to the inpu
 ```
 
 The editor for this input field will then turn into a secret input, and when you edit the field value, it will be stored encrypted.
-<img src="{{@asset actors/development/images/secret-input-editor.png}}" alt="Secret input editor" style={{width: '100%', maxWidth: '822px'}}/>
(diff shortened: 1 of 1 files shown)


sources/platform/actors/development/secret_input.md
```diff
@@ -36,7 +36,7 @@ To make an input field secret, just add a `"isSecret": true` setting to the inpu
 ```
 
 The editor for this input field will then turn into a secret input, and when you edit the field value, it will be stored encrypted.
-<img src="{{@asset actors/development/images/secret-input-editor.jpg}}" alt="Secret input editor" style={{width: '100%', maxWidth: '822px'}}/>
+<img src="{require("./images/secret-input-editor.png").default}" alt="Secret input editor" style="width: '100%'; max-width: 822px"/>
```
B4nan, line 39: I think the part with `style` was correct, that needs to stay as it needs to be valid JSX, same for `src`, something like this should work

```suggestion
<img src={require("./images/secret-input-editor.jpg").default} alt="Secret input editor" style={{width: '100%', maxWidth: '822px'}}/>
```

novotnyj: Thanks, changed.

B4nan (approved):

PR #609, opened 2023-06-09
docs: replace PHP tutorial

valekjo:
In the end I've replaced the old tutorial with one that is more up to date and added redirect.

Maybe we might still want to keep the old one?


sources/academy/tutorials/php/using_apify_from_php.md
```diff
@@ -0,0 +1,271 @@
+---
+title: Use Apify from PHP
+description: Learn how to access Apify's REST API endpoints from your PHP projects using the guzzle package. Follow a tutorial to run an actor and download its data.
+menuWeight: 3.93
+paths:
```
B4nan, line 5: you will need to migrate this to the new format

this is the script we used back then: https://github.com/apify/apify-docs-v2/blob/master/tools/convert.mjs

valekjo: Yeah, I've added few more changes.

B4nan (approved):

PR #618, opened 2023-06-20
docs: add changelog.md path example to actor config

metalwarrior665:
Please check that the behavior is the same as for readme so this is actually a valid description


mhamas (approved):

B4nan:
looks like this one got forgotten, @metalwarrior665 can you please rebase and merge?

metalwarrior665:
At least this was already implemented :D

PR #619, opened 2023-06-21
feat: Add Maximum number of metamorphs

drobnikj:
I not sure I would consider it as platform limit same as runtime, but runtime fits for in this case.


mtrunkat (approved):

drobnikj:
I add mentions about the limit into places talking about metamorph. Can you please check @TheoVasilis?

TheoVasilis (approved):

PR #625, opened 2023-06-26
feat: pull command

HonzaTuron:
Closes apify/apify-web#2725.

Adding mention about pulling actor from console (and how to develop locally)


sources/platform/actors/development/quick_start/images/actor-pull.png
jbartadev, line 1: I think we should use the Apify demo account for all the screenshots for docs so there isn't your username displayed but it's a minor thing

jbartadev (approved):
Should be enough to understand it.

PR #636, opened 2023-07-10
feat: add eslint to check js and json code in markdown files

metalwarrior665:
1. I think this achieves a pretty reasonable balance - stylistic issues are errors while missing and unused variables are just warnings because it is fine in some examples
2. I wasn't able to set it up for `ts`/`typescript` blocks, it currently shows an error for each block

```
Parsing error: ESLint was configured to run on `<tsconfigRootDir>/sources/academy/webscraping/switching_to_typescript/enums.md/0_0.ts` using `parserOptions.project`: /users/lukas/apify/apify-docs/tsconfig.eslint.json
However, that TSConfig does not include this file. Either:
- Change ESLint's list of included files to not include this file
- Change that TSConfig to include this file
- Create a new TSConfig that includes this file and include it in your parserOptions.project
See the typescript-eslint docs for more info: https://typescript-eslint.io/linting/troubleshooting#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file
```

3. Added JSON linter as well. For now it allows comments but not trailing commas (I saw few obvious bugs).


B4nan:
> I wasn't able to set it up for ts/typescript blocks, it currently shows an error for each block

And did you try the suggestions from it? You should have a tsconfig for the examples and specify it in the eslint config, you can see how its done in the crawlee docs here:

https://github.com/apify/crawlee/blob/master/docs/examples/.eslintrc.json
https://github.com/apify/crawlee/blob/master/docs/examples/tsconfig.json

(The paths mapping is not relevant here.)

metalwarrior665:
@B4nan Tried few configs but still no success, my understanding on these things is quite limited.

E.g. updating `tsconfig.eslint.json` with the last 2 lines still doesn't work on `.ts` and if changed to `.typescript` it understands it as JS only 

```json
{
    "extends": "./tsconfig.json",
    "include": [
        "src/**/*.js",
        "src/**/*.ts",
        "src/**/*.jsx",
        "src/**/*.tsx",
        "sources/**/*.md/*.ts",
        "sources/**/*.md/*.typescript"
    ]
}
```

I guess we can solve this later if no one has a good idea now.

B4nan:
I would not merge before having (most of) the warnings/errors resolved (including the TS problem), as this will make the CI fail, and even if we would comment that out, it would make the linting step useless. IMO we should enable the problematic rules gradually (and disable most of the warnings ⬇️).

> stylistic issues are errors while missing and unused variables are just warnings because it is fine in some examples

I disagree that we should have the warnings about unused variables etc enabled - we are linting code examples that are almost never complete (on purpose), they are destined to have such issues. The point of those code blocks is usually to show something, not to provide a complete/runnable code. What is the value of having such a spam in the linter output? I'd argue this makes things worse, not better, no one will be able to navigate the logs because of it.

I can imagine we will have different expectations from examples in separate JS/TS files, but imposing those rules on code blocks seems to be more disturbing than helpful.

--- 

Tried to get around the TS warnings, and looking at https://github.com/eslint/eslint-plugin-markdown/issues/114 I'd say this is simply not supported. Moreover, we use the eslint already for the actual source files in that repo, so we will need a separate config for the code blocks, as we want stricter rules for actual code.

B4nan:
I kinda like the idea of treating inline code blocks differently from the examples in separate TS files (which I would like to be complete, unlike the inline code snippets). And we should still have different eslint config for the examples.

> So on one side, you have inconvenience for us with floating warnings, and on the other one, you have broken code for users which is of course much worse.

My point is that with these warnings enabled, nobody will really take care as there will be hundreds of false positives and you will miss the few ones worth fixing. I like the lint command to print literally nothing when it's green, if we have lots of warnings we ignore on purpose, why should we have them in the first place? In a normal project, you can at least use ignore comments, but that's not a good fit for code examples.

metalwarrior665:
Let's go with the approach that I will add the fixes now in a batch and keep the warnings disabled (that's how it behaves normally). Then we keep only the rules that are not noisy but document the full-blown process somewhere so we can do a manual exercise twice a year or so.

mtrunkat (approved):
Seems legit to me but @B4nan is the one who knows here

sources/academy/platform/deploying_your_code/inputs_outputs.md
```diff
@@ -75,7 +75,7 @@ await Actor.init();
 
 const { numbers } = await Actor.getInput();
 
-const addAllNumbers = (...nums) => nums.reduce((total, curr) => (total += curr));
+const addAllNumbers = (...nums) => nums.reduce((total, curr) => (total + curr), 0);
```
B4nan, line 78: not that it would matter much, but imho the braces around function body make things less readable in the end

funny that we had missing params like this in the examples, great that we can now polish this

sources/academy/platform/getting_started/creating_actors.md
```diff
@@ -69,23 +69,23 @@ await Actor.exit();
...
+await Actor.init();
 
 console.log('Loading input');
 
 const input = await Actor.getInput();
 console.log('First number: ', input?.firstNumber);
 console.log('Second number: ', input?.secondNumber);
 
-const result = input?.firstNumber + input?.secondNumber;
+const result = (input?.firstNumber || 0) + (input?.secondNumber || 0);
```
B4nan, line 88: its a bit ugly when condensed into a single line, what about this? note that we want to use `??` here if something, not `||`

```suggestion
const firstNumber = input?.firstNumber ?? 0;
const secondNumber = input?.secondNumber ?? 0;
const result = firstNumber + secondNumber;
```

sources/academy/tutorials/api/run_actor_and_retrieve_data_via_api.md
```diff
@@ -102,15 +102,15 @@ import got from 'got';
 
 // Specify your API token
 // (find it at https://console.apify.com/account#/integrations)
-const myToken = <YOUR_APIFY_TOKEN>;
+const myToken = YOUR_APIFY_TOKEN;
```
B4nan, line 105: maybe better to keep that as string instead of some undefined var? something like this?

```suggestion
const myToken = '<YOUR_APIFY_TOKEN>';
```

sources/academy/tutorials/apify_scrapers/cheerio_scraper.md
```diff
@@ -70,9 +70,13 @@ And as we already know, there's only one.
...
 ```js
 // Using Cheerio.
-return {
-    title: $('header h1').text(),
-};
+async function pageFunction(context) {
+    const { $ } = context;
+    // ... rest of your code can come here
+    return {
+        title: $('header h1').text(),
```
B4nan, line 77: not that important, but we could use line highlighting for those examples that are not containing additional boilerplate

https://docusaurus.io/docs/markdown-features/code-blocks#line-highlighting

sources/academy/tutorials/apify_scrapers/puppeteer_scraper.md
```diff
@@ -442,28 +457,28 @@ We've shown two function calls, but how do we make this work together in the `pa
...
+    // ...
 
-let timeout; // undefined
-const buttonSelector = 'div.show-more > button';
-while (true) {
-    log.info('Waiting for the "Show more" button.');
-    try {
+    let timeout; // undefined
+    const buttonSelector = 'div.show-more > button';
+    for (;;) {
```
B4nan, line 464: is this really better than `while (true)`? :]

sources/academy/tutorials/node_js/block_requests_puppeteer.md
```diff
@@ -28,8 +28,8 @@ we could can use this piece of code
 ```js
 await page.setRequestInterception(true);
 page.on('request', (request) => {
-            if (someCondition) request.abort();
-            else request.continue();
+    if (someCondition) request.abort();
+    else request.continue();
```
B4nan, line 32: i'd rather see curly braces in control statements, its much more readable, especially if you have the else branch. this could be probably also autofixed via eslint.

sources/academy/tutorials/node_js/debugging_web_scraper.md
```diff
@@ -49,40 +49,42 @@ You can easily get all information you need by running a small snippet of your p
...
+        // other fields
+    });
+});
 ```
 
 Now the `results` variable stays on the page and you can do whatever you wish with it. Usually, simply log it to analyze if your scraping code is correct. Writing a single expression will also log it in a browser console.
 
 ```js
-results
+results;
```
B4nan, line 62: shouldn't this be `console.log(results)`?

B4nan (approved):
> The new .eslintrc is now set only for Academy but should be later moved to all sources (there are much less snippets in platform docs ofc so no hurry)

Why don't we able it for platform docs too right ahead? Especially if it's not that much work to be done? 

I went through a portion of the changes and left some suggestions, but its nothing serious, we can merge like this too. It's overall a huge improvement 👍

sources/academy/tutorials/node_js/handle_blocked_requests_puppeteer.md
```diff
@@ -14,20 +14,21 @@ One of the main defense mechanisms websites use to ensure they are not scraped b
...
-    requestList: someInitializedRequestList,
-    handleRequestFunction: async ({request}) => {
-        const response = await requestPromise ({
-            url: request.url,
-            proxy: proxyUrl
-        })
-   }
-})
+    requestList: someInitializedRequestList,
+    handleRequestFunction: async ({ request }) => {
```
B4nan, line 25: there will be probably more examples like this based on v2, this is nowadays called `requestHandler`, and the import should be from `crawlee`, not `apify`. those are probably worth fixing in a separate PR

sources/academy/tutorials/node_js/block_requests_puppeteer.md
metalwarrior665: I agree, just didn't go through these manaully. We would need to add a new rule to our lint.

sources/academy/tutorials/node_js/debugging_web_scraper.md
metalwarrior665: This section is about printing them in browser dev console. It is a bit weird with the semicolon but silences the linter.

sources/academy/tutorials/node_js/handle_blocked_requests_puppeteer.md
metalwarrior665: Yeah, those are some old not-important articles someone copied without much thought, I will have another pass on them in the future

sources/academy/tutorials/apify_scrapers/cheerio_scraper.md
metalwarrior665: How would you apply it here?

sources/academy/platform/deploying_your_code/inputs_outputs.md
metalwarrior665: Went too fast to even spot this :) removed the parens now

sources/academy/tutorials/apify_scrapers/cheerio_scraper.md
B4nan: the example is about the cheerio usage, so i would highlight those lines maybe?

metalwarrior665:
Ad platform docs - There are a few snippets but they have weird formats so I didn't want to get stuck on discussions around it. Let explore it next week.

PR #643, opened 2023-07-17
feat: use tweaked Algolia search index

### LINKED ISSUE #558: Improve docs search
When searching, it's not clear to what part of the docs the result is pointing. Is it SDK? Is it Python or JS SDK?

Ideally, this should be visible in the results, and also we should enable users to search in their current category - JS client or CLI, or academy.

obrazek

---

One more point to this from [Slack](https://apifier.slack.com/archives/C0L33UM7Z/p1681390378193129) by @mnmkng 

> The Docs search is dismal. For example, when I search for Key-value store on the docs homepage, I’m presented with a bunch of useless interfaces, then classes and only then I finally learn what a key-value store is.

obrazek

barjin:
I didn't manage to change the index's name yet, so this is mostly for testing. All the changes have been made to the Algolia crawler setup ([here](https://crawler.algolia.com/admin/crawlers/2d5cc8df-8a58-4488-9986-bfb033d2a29b/configuration/edit)), unfortunately, it has no GitHub integration, so we cannot version it here (but is versioned on the Algolia platform). 

With this PR, the search is a bit clearer - `Interface` and `Enumeration` results are hidden and every result is assigned to its respective Docusaurus instance (Apify Platform, Apify SDK JS etc.) in the results. 

Missing context-aware search (ranking current Docusaurus instance first), as this requires patching the Docsearch package (I have this prepared, but not in a production ready setup yet). 

| before | after | 
|---|---|
|  obrazek |   obrazek |

closes #558


B4nan:
We will need to have those changes in all the docusaurus instances, so it would be indeed better to not have it as a git patch.

apify-docs-theme/src/config.js
```diff
@@ -241,8 +241,8 @@ const themeConfig = ({
     },
     algolia: {
         appId: 'N8EOCSBQGH',
-        apiKey: 'b43e67a96ed18c7f63f5fd965906a96d', // search only (public) API key
-        indexName: 'apify_sdk',
+        apiKey: 'e97714a64e2b4b8b8fe0b01cd8592870', // search only (public) API key
+        indexName: 'crawler_dev_apify_sdk',
```
B4nan, line 245: so we will keep the new index?

barjin: That's definitely something to think about - while Algolia versions the Crawler configs, it's IMO a bit opaque / hidden there. I don't really have any preference regarding this - I did it like this since it's a tad easier to revert from here in case of any problems.

B4nan: yeah sure, i was just curious if this is to be changed or final. i guess there is not much to review here, so let's merge and see it in the wild? worst case we can revert

barjin: indeed, I just need the approving review to be able to merge :)

B4nan (approved):

B4nan:
lol 
image

PR #644, opened 2023-07-17
feat: Restructuring and updating proxy section of platform docs 3

jirimoravcik:
I removed some examples as they were just repeating the stuff and did not provide any value. I also moved some code examples to the corresponding text sections. Did a few cleanups, improved links


mtrunkat:
Positive score :)

CleanShot 2023-07-19 at 11 48 01@2x

PR #648, opened 2023-07-20
docs: Update monetizing_your_actor.md

OlikEten:
Added T&C for Store + note about price changes (allowed once per month)


OlikEten:
good point @B4nan , I have changed it - hopefully correctly :)

PR #662, opened 2023-08-01
feat: Redesign of the main docs landing page and typography unification across the docs

jbartadev:
Snímek obrazovky 2023-08-17 v 11 21 18

Snímek obrazovky 2023-08-17 v 11 22 04


barjin:
Oof, just copied the `custom.css` stylesheet to one of the side projects (Apify SDK JS). It seems that the biggest problem is heading size that's specified in `rems`, though - and in Docusaurus, heading sizes are stored in the `--ifm-h1-font-size` CSS variables. It's the same story with Docusaurus-defined paddings / margins.

Btw you get the same problem in the main Docusaurus instance (`apify-docs`) when you go to one of the docs pages (Academy / Platform) - that way, you get HMR while editing the css at least 😬

https://github.com/apify/apify-docs/assets/61918049/25410f39-0723-42b6-acae-9282d37846fc

jbartadev:
> Oof, just copied the `custom.css` stylesheet to one of the side projects (Apify SDK JS). It seems that the biggest problem is heading size that's specified in `rems`, though - and in Docusaurus, heading sizes are stored in the `--ifm-h1-font-size` CSS variables. It's the same story with Docusaurus-defined paddings / margins.
> 
> Btw you get the same problem in the main Docusaurus instance (`apify-docs`) when you go to one of the docs pages (Academy / Platform) - that way, you get HMR while editing the css at least 😬
> 
>  Peek.2023-08-16.14-34.mp4

@barjin Do you know if is there a way to change the values of the Docusaurus CSS variables?

barjin:
I think it should be as simple as redefining them in the theme's `custom.css` file (we're already doing it, [see here](https://github.com/apify/apify-docs/blob/e4e26f57f4ff30526619224f443f99e001fa035e/apify-docs-theme/src/theme/custom.css#L38)) - haven't tried it though.

.github/workflows/publish-theme.yaml
```diff
@@ -42,14 +42,9 @@ jobs:
                 uses: actions/setup-node@v3
                 with:
                     node-version: 16
-
-            -   name: Setup git user and npm
-                run: |
-                    git config --global user.name "Apify Release Bot"
```
B4nan, line 48: I would guess this needs to stay, we need to be able to push to git from here. Same with the NPM thingy, that is so we can publish new themes to the `@apify/` namespace.

apify-docs-theme/src/theme/custom.css
```diff
@@ -393,36 +592,36 @@ html.plugin-docs .theme-doc-markdown {
...
-html.plugin-docs .theme-doc-markdown h1 {
+html .plugin-docs .theme-doc-markdown h1 {
     font-weight: 600;
     font-size: 48px;
     line-height: 64px;
     color: #000;
 }
 
-html[data-theme='dark'].plugin-docs .theme-doc-markdown h1 {
+html[data-theme='dark'] .plugin-docs .theme-doc-markdown h1 {
```
B4nan, line 607: This looks wrong, the plugin-docs class is in the html element unless you changed something.

.github/workflows/publish-theme.yaml
jbartadev: Ha, you're right. My bad, I thought it was connected only to the installation.

drobnikj (approved):
I like the new design!

apify-docs-theme/src/theme/custom.css
jbartadev: I thought so too but then I noticed it without the spaced it overrides all H1s everywhere and with the space it affects only the `.plugin-docs class`

B4nan:
Looks like this got somehow forgotten. I guess it was waiting for us to merge (while we are generally used to let author merge their own PRs once they are approved).

Would be good if someone rebases it.

barjin:
Unfortunately, this PR still breaks the styling pretty much everywhere in the docs. The only culprit is the `font-size: 10px` thing on the root `html` tag... I guess we cannot get rid of it soon @jbartadev , right? :/ 

> Flipping through the current docs and this PR:

Peek 2023-10-04 13-58

> Closer look at a random page with this PR changes:

obrazek

To me, it's mostly the smaller paddings + font sizes all over the page.

After merging the latest master, this PR also messes up the search modal :'(

apify-docs-theme/src/theme/custom.css
```diff
@@ -164,6 +165,8 @@ html[data-theme='dark'] {
     --ifm-h5-font-size: 1.4rem;
     --ifm-h6-font-size: 1.2rem;
 
+    --ifm-heading-line-height:
```
B4nan, line 168: FYI this one is missing a value

jbartadev: My bad, this is still in progress so you can just ignore the changes until we request the review again.

CC @webrdaniel

apify-docs-theme/src/theme/custom.css
```diff
@@ -74,6 +159,20 @@ html[data-theme='dark'] {
...
 
+    --ifm-leading: calc(var(--ifm-leading-desktop) * 1.6rem);
+    --ifm-list-margin: 1.6rem;
+    --ifm-pre-padding: 1.6rem;
+
+    --ifm-hr-margin-vertical: 2.4rem;
+
+    /* Comes from the Apify design theme content typography tokens */
+    --ifm-h1-font-size: 2.4rem;
+    --ifm-h2-font-size: 2rem;
```
OPesicka, line 170: These don't get applied in the docs. Even when I tried adding `!imporant` the font-size is still overridden by `.markdow`

image

apify-docs-theme/src/theme/custom.css
```diff
@@ -42,16 +125,18 @@ html[data-theme='dark'] {
     --ifm-font-weight-semibold: 600;
     --ifm-font-color-base: #242736;
```
OPesicka, line 127: We should get rid of the `Lota Grotesque` font family in favor of system fonts.

apify-docs-theme/src/theme/custom.css
```diff
@@ -74,6 +159,20 @@ html[data-theme='dark'] {
 
     --ifm-heading-color: #242736;
 
+    --ifm-leading: calc(var(--ifm-leading-desktop) * 1.6rem);
+    --ifm-list-margin: 1.6rem;
+    --ifm-pre-padding: 1.6rem;
+
+    --ifm-hr-margin-vertical: 2.4rem;
+
+    /* Comes from the Apify design theme content typography tokens */
```
OPesicka, line 168: The font sizes are correct but when we were defining them we forgot that H1 basically doesn't exist in markdown and you use H2 as the "main" heading. So we should push down the font-sizes so that H1 size is not H2 and come up with a bigger H1 size.

src/pages/index.module.css
```diff
@@ -1,18 +1,145 @@
...
-    align-items: stretch;
-    flex-wrap: wrap;
-    box-sizing: border-box;
+    flex-direction: column;
+    align-items: flex-start;
+    align-self: stretch;
+    gap: 2.4rem;
+}
+
+.bannerContentActions {
```
OPesicka, line 29: Just a nit the titles in these cards are slightly misaligned 
image

OPesicka (commented):
Did a first pass, but found some issues that need fixing before I dig deeper (mainly the heading sizes)

apify-docs-theme/src/theme/custom.css
```diff
@@ -635,10 +889,27 @@ html[data-theme='dark'] .actionLink:hover::after {
     margin-left: 8px;
 }
 
+.cardContentWrapper {
+    display: flex;
+    padding-top: 25.6px;
+    flex-direction: column;
+    align-items: flex-start;
+    gap: 25.6px;
```
OPesicka, line 897: This doesn't seem right. If ti only affects these cards (viz screenshot) I'd go for 16px
image

apify-docs-theme/src/theme/custom.css
```diff
@@ -346,14 +594,13 @@ aside button[class*="collapseSidebarButton"] svg {
 }
 
 nav.navbar .dropdown__menu {
-    min-width: 6rem;
+    min-width: 96px;
```
OPesicka, line 597: The padding is too small now. I'd do 12/16px 
image

OPesicka (approved):
2 nits then it's ready to go from my side 🚀

PR #666, opened 2023-08-06
Update interacting_with_a_page.md 

mr-paperbag:
Google uses textarea as opposed to input. Also added a note about title="search" differing in other languages.


B4nan:
Thanks! 

Using selectors based on translatable text feels pretty wrong to me. I see there is `type="search"` on the textarea, maybe better to use that in the selector?

PR #669, opened 2023-08-08
feat: add @apify/docs-search-modal

### LINKED ISSUE #655: Investigate possible improvements in the search modal
We recently found out the algolia developer portal has a very nice search experience, they describe how to do it in this article:

https://www.algolia.com/blog/ux/replicating-the-algolia-documentation-search-with-autocomplete/

They use this in the developer docs: https://www.algolia.com/doc/

We'd like to assess how much work would it be to adjust our search to have:
- section tags, the same way as they have it for different languages 
	- e.g. select `js sdk` tag and search only in there
- two-column layout, with the description + ToC, exactly as they do it
	- this will be probably also about how/what we index the content, not just about the UI
- highlighting the matching part in the content, which works on the right side too
	- right now, we have a lot of the matches only for the headings, and those do not contain the `content` property at all, so might be only about improving the index options, the functionality should be already there
	- I'd expect that the match for a heading should also contain some text (the one that the heading is for)

barjin:
Adds the new search modal with the result preview. 

obrazek

~The modal is definitely still wip - mostly small CSS things (dark mode is not dark), responsiveness / styling of the search "button" etc.~

Closes #655


barjin:
The PR is still missing the search in versioned files and the dynamic import of the modal.

btw did the dynamic import improve anything (considering it all gets `webpack`-ed into one bundle either way?)

barjin:
Video of the current state:
<video src="https://github.com/apify/apify-docs/assets/61918049/2c1ffef9-37a0-4025-aac1-c72c41e953e1" />

itstrangvu:
We don't have stagings for docs, right? :) I would like to take a peek if everything is okay design-wise. Do I need to run this locally?

barjin:
Yes, there is no staging for docs I'm afraid. If you're in the office, I can run it from my server and share the address on Slack :)

itstrangvu:
I'm in the office and I like the idea.

itstrangvu:
1. The fact that the first element in the list not visually pointing to the preview window on the right is weird. 
First item not linked
Proposal

itstrangvu:
I suggest applying [our color tokens](https://apify-web-brand.supernova-docs.io/latest/shared-tokens/color/semantic-palette-AcFqGGHy): For the text highlight, you can apply the Bamboo color from [Decorative palette](https://apify-web-brand.supernova-docs.io/latest/shared-tokens/color/decorative-palette-e8bM4Uoa)

Screenshot 2023-08-15 at 13 29 30

itstrangvu:
The breadcrumb in the preview has low line-height, resulting in certain letters being cut at the bottom. I suggest increasing the line-height. 
Screenshot 2023-08-15 at 13 34 52

itstrangvu:
Please replace the `>` with the respective icon. Both in the list and the preview.
Screenshot 2023-08-15 at 13 44 46

This is how it looks in the Docs. 
Screenshot 2023-08-15 at 13 46 22
Let's keep this consistent.

itstrangvu:
The whole preview is not clickable, just the links in the preview. 
Screenshot 2023-08-15 at 13 54 46

barjin:
All the mentioned issues have been taken care of, see the recording below. I'm also now injecting the "React Router" (client JS) navigation to the modal for faster load times and better UX. 

Feel free to leave any further comments if you find anything in the video.

https://github.com/apify/apify-docs/assets/61918049/8da3f45c-56ca-466d-b56e-59a28a01f3c6

barjin:
The new `@apify/docs-search-modal` version ([git repo here](https://github.com/apify/docs-search-modal)) features fixed fonts, same document icons as the current docsearch modal and a keyboard tooltip in the footer. 

@B4nan feel free to try out locally.

B4nan (changes requested):
Finally gave it a closer look locally, few things to address:
- shortcuts are not working 
	- `/` and `cmd+k` should open the modal
	- `esc` should close it
	- using arrows to iterate through the results is working, but it does not scroll when you go too low
- the input box in navbar should look the same, including the shortcut hint)
- we dont need the `cancel` button
- I'd keep the previous colors, this looks a bit too gray'ish 
	- blue instead of green as the highlight color
	- section headings (now with gray background) could have blue background (like when you hover over the results in the current one)
	- you could play a bit with different shades if you want to distinguish links from highlighting, but they can look almost the same imo
- the margin/padding between the description and "On this page" is too huge
- in the search input field, the font size (as well as the icon) should be bigger, as in the current version (16px vs 19.2px, it doesn't have to be exact same as before, but it could be a bit bigger)
- when the page contains code tabs (e.g. `academy -> proxies`), it gets rendered the titles first, and then the blocks, but I guess we cannot do much about that?

Generally speaking, I'd like the look and feel to be much closer to the current one.

Plus there is this problem:

image

I'd expect we won't take the parent page names into account, it feels weird to list everything under some section just because the search query is matching it. Looks like the current version does not suffer from this.

barjin:
I have addressed all of the (rendering | logic) issues in `@apify/docs-search-modal@1.0.18`. You can pull the latest commit here, reinstall deps and try it out. 

Regarding the last thing - I've updated the index so that it contains the name of the class (interface, enum...), so the results hopefully make more sense. I think it's kind of nifty to have the other results there too, especially if they are rendered in the hierarchy like this.
obrazek

barjin:
Following our conversation, hierarchy-only hits (where the result is considered relevant for the query only based on its parents) are now rendered only if the parent itself is one of the results. This is best shown on results from the reference pages - when searching for a class name, you will get the main document as well as the class's methods etc. 

Otherwise, those hierarchy-only matches are omitted from the results. 

obrazek

obrazek

B4nan (approved):
image

PR #676, opened 2023-08-11
feat: proofread and edit for a2a docs

m-murasovs:
It felt more appropriate to move the implementation details to the Index, just because it feels like it makes it easier to understand quicker.

Basically I just added some links, replaced irregular punctuation, put some stuff in lists, that kinda thing.

I think structure-wise it's good and there's a good amount of information. We could maybe get someone who hasn't been involved in developing this to read it and see if anything's unclear? Maybe Mato if he can spare some time, as he always has good questions


sources/platform/integrations/actors/index.md
```diff
@@ -5,16 +5,28 @@ sidebar_position: 11.10
...
 ---
 
-*The feature is experimental and currently available on-request, ask our support to enable it for your account.*
+> This is an experimental feature and currently available on request. Ask our support to enable it for your account.
 
-### What are Actors integrations?
+# What are Actor integrations?
 
-It is a way how to use Actors with other Actors easily. It provides new level of flexibility as adding a new integration simply means creating the Integration-ready Actor. Thus, new integrations can be created by the community itself.
+ Actor integrations provide a way to connect your Actors with other Actors easily. They provide a new level of flexibility, as adding a new integration simply means creating [integration-ready Actors]('/platform/integrations/integration-ready-actors'). Thus, new integrations can be created by the community itself, without relying on external parties.
```
valekjo, line 12: I'd omit the external parties, it's more "without relying on Apify team"

valekjo (changes requested):
One thing to be discussed, otherwise thanks a ton!

valekjo (approved):

PR #677, opened 2023-08-14
feat: explain usage usd on run and build objects

mhamas:
Explaining that when you pass auth token to run/build GET requests, you get dollar values using the pricing effective at the time of query.

### DIFF
--- a/sources/platform/actors/running/usage_and_resources.md
+++ b/sources/platform/actors/running/usage_and_resources.md
@@ -89,7 +89,7 @@ When you run an Actor it generates platform usage that's charged towards the use
 - **Storage operations**: Read, write, and other operations towards key-value store, dataset, and request queue.
 
 The platform usage can be represented either in raw units (e.g. gigabytes for data transfer, or number of writes for dataset operations), or in the dollar equivalents.
-Note that for technical reasons, every time you see the usage in dollars on a specific historical Actor run or build in the API or the Apify Console, your current service pricing is used to compute the dollar amount,
+Note that for technical reasons, every time you see the usage in dollars on a specific historical Actor run or build in the API or Apify Console, your current service pricing is used to compute the dollar amount,
 and hence it should be used only for informative purposes.
 
 > For detailed information, FAQ, and, pricing see the [platform pricing page](https://apify.com/pricing/actors).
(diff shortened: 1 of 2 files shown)


sources/platform/api_v2/api_v2_reference.apib
```diff
@@ -863,6 +863,10 @@ Gets an object that contains all the details about a specific build of an actor.
 By passing the optional `waitForFinish` parameter the API endpoint will synchronously wait for the build to finish.
 This is useful to avoid periodic polling when waiting for an actor build to finish.
 
+This endpoint does not require the authentication token, the calls are authenticated using a hard-to-guess ID of the build. If
```
TheoVasilis, line 866:
```suggestion
This endpoint does not require the authentication token. The calls are authenticated using a hard-to-guess ID of the build. If
```

sources/platform/api_v2/api_v2_reference.apib
```diff
@@ -1293,7 +1297,9 @@ Gets an object that contains all the details about a specific run of an actor.
 By passing the optional `waitForFinish` parameter the API endpoint will synchronously wait for the run to finish.
 This is useful to avoid periodic polling when waiting for actor run to complete.
 
-This endpoint does not require the authentication token, the calls are authenticated using a hard-to-guess ID of the run.
+This endpoint does not require the authentication token, the calls are authenticated using a hard-to-guess ID of the run. If
```
TheoVasilis, line 1300:
```suggestion
This endpoint does not require the authentication token. The calls are authenticated using a hard-to-guess ID of the run. If
```

TheoVasilis (dismissed):

drobnikj (changes requested):
The usageUsd and usageUsdTotal is presented in list of runs or builds. Plus in some endpoints which we can run just with token, we should add info there as well. 

You can write short info (or reuse current one) into https://docs.apify.com/platform/actors/running/usage-and-resources
and link it into API reference and in the console.

drobnikj (approved):

sources/platform/actors/running/usage_and_resources.md
```diff
@@ -88,6 +88,10 @@ When you run an Actor it generates platform usage that's charged towards the use
 - **Proxy costs**: Residential or SERP proxy usage.
 - **Storage operations**: Read, write, and other operations towards key-value store, dataset, and request queue.
 
+The platform usage can be represented either in raw units (e.g. gigabytes for data transfer, or number of writes for dataset operations), or in the dollar equivalents.
+Note that for technical reasons, every time you see the usage in dollars on a specific historical Actor run or build in the API or the Apify Console, your current service pricing is used to compute the dollar amount,
```
TheoVasilis, line 92:
```suggestion
Note that for technical reasons, every time you see the usage in dollars on a specific historical Actor run or build in the API or Apify Console, your current service pricing is used to compute the dollar amount,
```

TheoVasilis (dismissed):

PR #681, opened 2023-08-22
fix: typo

mvolfik:
This is quite an unfortunate error :/


drobnikj (approved):

drobnikj:
@mvolfik Thanks for the fix!

PR #700, opened 2023-09-07
chore: Change branch for building docs in child repos

fnesveda:
I've finally merged the `docs-v2` branches in `apify-sdk-python` and `apify-client-python`, and the last thing to do is to start triggering the `docs` workflow in the right branches when a new docs theme is published.

### DIFF
--- a/.github/workflows/publish-theme.yaml
+++ b/.github/workflows/publish-theme.yaml
@@ -97,21 +97,16 @@ jobs:
         needs: publish
         strategy:
             matrix:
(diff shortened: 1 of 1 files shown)


B4nan (approved):
[CONTRIBUTING](https://github.com/apify/apify-docs/blob/master/CONTRIBUTING.md#local-setup) needs to be updated too, otherwise lgtm

vdusek (approved):
Good job :+1:

PR #709, opened 2023-09-21
feat: add Run on Apify button support

B4nan:
only a few examples are converted, most of them are rather incomplete and some are dated and no longer work, unfortunately.

also uses node 18 in the CI


apify-docs-theme/src/theme/MDXComponents/Code.js
```diff
@@ -0,0 +1,42 @@
+import React, { isValidElement } from 'react';
```
B4nan, line 1: had to eject this completely, docusaurus now forbids wrapping of this component for some reason

PR #715, opened 2023-09-24
docs: fix broken link

AbdassalamAhmad:
… section

### DIFF
--- a/sources/academy/tutorials/apify_scrapers/getting_started.md
+++ b/sources/academy/tutorials/apify_scrapers/getting_started.md
@@ -108,7 +108,7 @@ Some of this information may be scraped directly from the listing pages, but for
 
 ### [](#the-start-url) The start URL
 
-Let's start with something simple. In the **Input** tab of the task we have, we'll change the **Start URL** from **<https://apify.com>**. This will tell the scraper to start by opening a different URL. You can add more **Start URL**s or even [use a file with a list of thousands of them](https://docs.apify.com/academy/node-js/scraping-urls-list-from-google-sheets), but in this case, we'll be good with just one.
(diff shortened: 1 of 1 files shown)


sources/academy/tutorials/apify_scrapers/getting_started.md
```diff
@@ -108,7 +108,7 @@ Some of this information may be scraped directly from the listing pages, but for
 
 ### [](#the-start-url) The start URL
 
-Let's start with something simple. In the **Input** tab of the task we have, we'll change the **Start URL** from **<https://apify.com>**. This will tell the scraper to start by opening a different URL. You can add more **Start URL**s or even [use a file with a list of thousands of them](//crawl-urls-from-a-google-sheet), but in this case, we'll be good with just one.
+Let's start with something simple. In the **Input** tab of the task we have, we'll change the **Start URL** from **<https://apify.com>**. This will tell the scraper to start by opening a different URL. You can add more **Start URL**s or even [use a file with a list of thousands of them](https://docs.apify.com/academy/apify-scrapers/getting-started#-crawling-the-website-with-pseudo-urls), but in this case, we'll be good with just one.
```
B4nan: i think it should be just this

```suggestion
Let's start with something simple. In the **Input** tab of the task we have, we'll change the **Start URL** from **<https://apify.com>**. This will tell the scraper to start by opening a different URL. You can add more **Start URL**s or even [use a file with a list of thousands of them](#-crawling-the-website-with-pseudo-urls), but in this case, we'll be good with just one.
```

AbdassalamAhmad:
yes, @b4nan  you are correct in case you want to refer to this section **"#-crawling-the-website-with-pseudo-urls"** of the same page, but I think it can be also referring to this docs page [scraping-urls-list-from-google-sheets](https://docs.apify.com/academy/node-js/scraping-urls-list-from-google-sheets)

Can you update me with the right URL and I will commit the new update.

B4nan:
It says `crawl-urls-from-a-google-sheet` which is an exact match to that anchor on the same page, while the other is a bit different.

AbdassalamAhmad:
oh yeah I see.
I will change it back.

B4nan (approved):

B4nan:
thanks!

PR #725, opened 2023-10-03
fix(docs): actor status message api params

katacek:
Fixing bug reported on [slack](https://apifier.slack.com/archives/C0L33UM7Z/p1696326166649169): The docs at https://docs.apify.com/api/v2#/reference/actor-runs/update-status-message/update-status-message are wrong, the params (except token) can't be in URL, but must be in PUT body as json

Checked w matej that it is ok as well ✅ 

Screenshot 2023-10-03 at 14 30 54

### DIFF
--- a/sources/platform/api_v2/api_v2_reference.apib
+++ b/sources/platform/api_v2/api_v2_reference.apib
@@ -2312,16 +2312,6 @@ you will typically do this in order to inform users of your actor about the acto
 
 The request body must contain `runId` and `statusMessage` properties. The `isStatusMessageTerminal` property is optional and it indicates if the status message is the very last one. In the absence of a status message, the platform will try to substitute sensible defaults.
 
-* `runId` is required string of run ID
-* `statusMessage` is required string of the status message
-* `isStatusMessageTerminal` is optional boolean indicating if this is the final status message of the actor
-
-```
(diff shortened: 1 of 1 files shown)


sources/platform/api_v2/api_v2_reference.apib
```diff
@@ -2308,15 +2308,27 @@ For more information, see the [Actor docs](https://docs.apify.com/platform/actor
 ### Update status message [PUT]
 
 You can set a single status message on your run that will be displayed in the Apify Console UI. During an actor run,
-you will typically do this in order to inform users of your actor about the actor's progress. If the status message is the very last one,
-you will also need to specify `isStatusMessageTerminal=true`. In the absence of a status message, the platform will try to substitute sensible defaults.
+you will typically do this in order to inform users of your actor about the actor's progress.
+
+The request body must contain `runId` and `statusMessage` properties. The `isStatusMessageTerminal` property is optional and it indicates if the status message is the very last one. In the absence of a status message, the platform will try to substitute sensible defaults.
+
+* `runId` is required string of run ID
```
jancurn, line 2315: The part here is just repeating `StatusMessageUpdate`, IMO no need for that

jancurn (approved):

PR #726, opened 2023-10-04
fix: correct properties for TaskUpdate

mvolfik:


B4nan:
this one is getting stale, can someone more informed than me review this please? maybe @valekjo?

valekjo:
The editable properties are:

```js
export const ACTOR_TASK_EDITABLE_FIELDS = ['description', 'input', 'name', 'title', 'options'];
```

So this is ok, unless we're aiming to make it complete (I'm not sure if we're planning to keep the api docs like this).

B4nan:
not sure if the openapi specs are more up to date or they need the same fix (https://docs.apify.com/api/v2-new#tag/Actor-tasks/operation/Updatetask)

PR #734, opened 2023-10-12
chore: bump @apify/docs-search-modal version

barjin:
Adds different icon for fragment-level records and expands the breadcrumbs.

See the left pane (results list) in the before / after comparison:

| before | after |
|---|---|
| obrazek | obrazek |


B4nan (approved):
shouldn't this happen automatically?

barjin:
that would indeed be nice, I might look into it soon

PR #739, opened 2023-10-18
feat(docs): Update info about input schema default, prefill, required, ...

jancurn:
See https://apifier.slack.com/archives/C0L33UM7Z/p1697624841289719

### DIFF
--- a/sources/platform/actors/development/actor_definition/input_schema/specification.md
+++ b/sources/platform/actors/development/actor_definition/input_schema/specification.md
@@ -110,11 +110,18 @@ Each field of your input is described under its key in `inputSchema.properties` 
 | `sectionCaption`     | String                                                                            | No       | If this property is set, <br/>then all fields following this field <br/>(this field included) will be separated <br/>into a collapsible section <br/>with the value set as its caption. <br/>The section ends at the last field <br/>or the next field which has the <br/> `sectionCaption` property set. |
 | `sectionDescription` | String                                                                            | No       | If the `sectionCaption` property is set, <br/>then you can use this property to <br/>provide additional description to the section. <br/>The description will be visible right under <br/>the caption when the section is open.                                                                           |
 
-Here is the rule of thumb for whether an input field should have a `prefill`, `default`, or empty value:
+### Prefill vs. default vs. required
 
-- **Prefill** - For fields that don't have reasonable defaults and user should change them (e.g. search keyword, example start URL)
-- **Default** - For everything that should be set for the Actor run to some value (e.g. max pages to crawl, proxy settings)
-- **Empty** - For purely opt-in features
+Here is a rule of thumb for whether an input field should have a `prefill`, `default`, or be required:
+
+- **Prefill** - Use for fields that don't have a reasonable default. The provided value is prefilled for the user, to show them an example how to use the field and make it easy to test the Actor (e.g., search keyword, start URLs)
+- **Required** - Use for fields that don't have a reasonable default and must be entered by the user (e.g., API token, password).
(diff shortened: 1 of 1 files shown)


sources/platform/actors/development/actor_definition/input_schema/specification.md
```diff
@@ -110,11 +110,18 @@ Each field of your input is described under its key in `inputSchema.properties`
...
+
+Here is a rule of thumb for whether an input field should have a `prefill`, `default`, or be required:
+
+- **Prefill** - Use for fields that don't have a reasonable default. The provided value is prefilled for the user, to show them an example how to use the field and make it easy to test the Actor (e.g., search keyword, start URLs)
+- **Required** - Use for fields that don't have a reasonable default and must be entered by the user (e.g., API token, password).
+- **Default** - Use for fields that must be set for the Actor run to some value (e.g., max pages to crawl, proxy settings), but where you don't need the user to change the default behavior.
+  If the user omits the value when starting the Actor, the platform passes the Actor this default value automatically.
+- **No particular setting** - Use for purely optional fields where it makes no sense to prefill any value (e.g., flags like debug mode or download files).
+
+In summary, you can use each option separately or use a combination of **Prefill + Required**, but the combinations **Prefill + Default** or **Default + Required** don't make sense to use.
```
B4nan, line 123: we actually use prefill+default+required for the proxy input in WCC now, with just default the UI allowed to set "no proxy", with default + required it failed with the reset button. so reading this, we should use just `prefill + required`? we initially had just the default, then added required (as we did not want to allow no proxy), then added prefill (as otherwise the UI reset would break it as mentioned on slack).

also many other options in WCC use prefill and default at the same time (but are not required)

jancurn: cc @valekjo @metalwarrior665 - that might have been a bug, which will get fixed

jancurn: ping again Pepa/Lukas - what do you think guys?

valekjo: Hey, today a fix is being deployed, that makes sure that default values are applied before validation. This happened on backend, but not in UI, and hopefully that was the cause.

Proxy validation is different from other fields, because it can be filled with `{ useApifyProxy: false }` - when it has this value, it's considered to be empty too.

So maybe I'd just mention this as only exception?

valekjo: And maybe I'd mention that `prefill` is just a console thing, whereas `default` and `required` are valid for both console and api.

jancurn: Cheers. I'd skip the example with `{ useApifyProxy: false }`, it would just add more confusion.

jancurn: Anyway, updated, hopefully it's correct now

metalwarrior665: Proxy should be default, not prefill, because it is something that must be set. The bug on platform prevented that but once it is fixed, it should be used.

jancurn: so the current version of text is okay? let's get this out

metalwarrior665 (approved):
I think it is fine.

sources/platform/actors/development/actor_definition/input_schema/specification.md
metalwarrior665: Yeah, I was thinking if the inclusion of required makes sense but after some thought, it is quite good :)

valekjo (approved):

PR #741, opened 2023-10-23
feat: add scrape single url action into Zapier docs

drobnikj:


sources/platform/integrations/zapier.md
```diff
@@ -113,6 +113,11 @@ Once you are happy with the test, you can publish the Zap. When it is turned on,
 
 > Runs a selected Actor task.
 
+### Scrape Single URL
+
+> Runs a scraper for the website and returns its content as text, markdown and HTML.
+> This action is made for getting content of a single page, for example, to use in large language models (LLM) flows.
```
TheoVasilis, line 119:
```suggestion
> This action is for getting content of a single page to use, for example, in LLM flows.
```

fnesveda (approved):
Theo's edit looks good, we should also use it directly in the Zapier action description.

TheoVasilis (approved):

PR #756, opened 2023-11-09
docs: fix typo in continuous integration page

janbuchar:


sources/platform/actors/development/deployment/continuous_integration.md
```diff
@@ -102,7 +102,7 @@ https://api.apify.com/v2/acts/YOUR-ACTOR-NAME/builds?token=YOUR-TOKEN-HERE&versi
 
 ## Set up automatic builds
 
-TODO: This duolicates somehow the above part
+TODO: This duplicates somehow the above part
```
B4nan, line 105: lets turn this into a comment right ahead? i dont like to have publicly visible TODOs, especially in the docs
```suggestion
[//]: # (TODO: This duplicates somehow the above part)
```

janbuchar: I'm concerned that nobody will notice (and resolve) the TODO now that it's hidden...

B4nan: The sad truth is more like "nobody resolves todos", I don't think its about printing them to users.

Would you `console.log('this is a TODO')` in your app code? :]

mtrunkat (approved):

PR #777, opened 2023-12-04
docs: rewrite schedules section

TC-MO:
update screenshot to reflect current UI
rewrites for clarity and conciseness
rewrites for consistency across docs
rewrites for style guide compliance


sources/platform/schedules.md
```diff
@@ -6,79 +6,84 @@ category: platform
...
-# Schedules
-
 **Learn how to automatically start your Actor and task runs and the basics of cron expressions. Set up and manage your schedules from Apify Console or via API.**
 
 ---
 
-[Schedules](https://console.apify.com/schedules) allow you to run your Actors and tasks at specific times. You schedule the run frequency using [cron expressions](#cron-expressions).
+Schedules allow you to run your Actors and tasks at specific times. You schedule the run frequency using [cron expressions](#cron-expressions).
+
+:::note Timezone & Daylight Savings Time
```
valekjo, line 15: Question: How does the `note` work? Is that some custom component we have?

sources/platform/schedules.md
```diff
@@ -6,79 +6,84 @@ category: platform
...
 
-You can set up and manage schedules from
-[Apify Console](https://console.apify.com/schedules)
-and via [API](/api/v2#/reference/schedules)
-(also with the
-[JavaScript](/api/client/js/reference/class/ScheduleClient) and
-[Python](/api/client/python/reference/class/ScheduleClient)
-API clients).
-When scheduling a new Actor or task run, you can override its input settings using a JSON object similarly to when invoking a schedule using the [Apify API](/api/v2#/reference/schedules/).
+When scheduling a new Actor or task run, you can override its input settings using a JSON object similarly to when invoking a schedule using the [Apify REST API](/api/v2#/reference/schedules/).
```
valekjo, line 26:
```suggestion
When scheduling a new Actor or task run, you can override its input settings using a JSON object similarly to when invoking an Actor or task using the [Apify REST API](/api/v2#/reference/schedules/).
```

sources/platform/schedules.md
```diff
@@ -6,79 +6,84 @@ category: platform
...
 
-Each schedule can be associated with a maximum of 10 Actors and 10 Actor tasks.
+Each schedule can be associated with a maximum of _10_ Actors and _10_ Actor tasks.
 
 ## Setting up a new schedule
 
 Before setting up a new schedule, you should have the [Actor](./actors/index.mdx) or [task](./actors/running/tasks.md) whose run you want to schedule prepared and tested.
 
-If you are planning to schedule an Actor run, you need to use the Actor before you can schedule any runs. Navigate to the Actor's page in [Apify Store](https://apify.com/store), click the **Try for free** button, then **Start** it with your preferred settings.
+To schedule an Actor, you need to have run it at least once before. To run the Actor, navigate to the Actor's page in [Apify Store](https://apify.com/store), and click the **Try for free** button. This will direct you to Apify Console, where you can configure and initiate the Actor's run with your preferred settings. After this initial run, you can then use Schedules to automate future runs of the Actor
```
valekjo, line 39: Question: Not sure if this is true, i think it's more like "having interacted" with the Actor, not necessarily running it.

Maybe this could be just some: Can't find the Actor? We're only showing Actor's you are using....

sources/platform/schedules.md
```diff
@@ -6,79 +6,84 @@ category: platform
...
 To add more Actors or tasks, just repeat the process.
 
-> You can add a maximum of 10 Actors and 10 tasks to each schedule.
-
 Now, all you need to do is click **Save & activate** and let the scheduler take care of running your jobs on time.
 
 For integrations, you can also add a [webhook](/platform/integrations/webhooks) to your tasks, which will notify you (or perform an action of your choice) every time the task runs.
 
-### Via API
+### Apify API
```
valekjo, line 74: Question: I guess we might have this section, briefly describing how to work with API, in all relevant places in docs. Or maybe just link to some "Getting started with API" section + link to docs of relevant endpoints?

valekjo (commented):
Nice! Leaving few questions.

sources/platform/schedules.md
TC-MO: nope, this is built in into docusaurus as default component, but we can style it if we want to, here is the relevant lint in docusaurus documentation

https://docusaurus.io/docs/2.x/markdown-features/admonitions

sources/platform/schedules.md
TC-MO: That is true, you just have to interact it, but are there any situation when you would interact with it but not run it? Maybe something along the lines

`To schedule an Actor, you need to interact with it at least once before. An example of such interaction is running ii` and here proceed with text as is to explain how to run an actor?

valekjo: There technically is - eg. saving the input, and not clicking "Start", but it's not very probable. So looking at it like this, let's stay with run :).

Maybe I'd just change the process from "go to store, try for free, run it" to one that would stay within the console?

sources/platform/schedules.md
valekjo: Nice, didn't know about this.

valekjo (approved):

PR #792, opened 2023-12-14
docs: integrate new og:image links

barjin:
Long time no `patch-package` - unfortunately, I didn't come up with a better solution (that wouldn't adding a frontmatter tag to every markdown file in the repo).

Related to apify/apify-web#3159

### DIFF
--- a/apify-docs-theme/src/config.js
+++ b/apify-docs-theme/src/config.js
@@ -145,7 +145,7 @@ const themeConfig = ({
         additionalLanguages: ['docker', 'log', 'php'],
     },
     // this needs to be absolute link otherwise it gets resolved wrongly in project docs
-    image: 'https://docs.apify.com/img/docs-og.png',
+    image: 'https://apify.com/api/og-image/docs-article',
     footer: {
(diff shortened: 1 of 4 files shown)


B4nan (approved):
would be great to PR this to docusaurus, i dont mind the patch but given we soon need to do the v3 update... :]

barjin:
I doubt they will want to merge `const ogImageURL = new URL('https://apify.com/api/og-image/docs-article');` anywhere in their codebase :) 

Unfortunately, this is a rather specific use case, I'm not too sure how it would fit the current Docusaurus API. I wish there was some way of passing a function for modifying the frontmatter, but that sounds like too much free labor with a very unsure outcome.

B4nan:
Yeah, I am saying we could PR a support for what we want - dynamically, not hardcode our URLs :] 

Maybe we could at least open an issue on their end, there could be a cleaner way than patching (e.g. some swizzling). But it's fine for now so we are not blocked. Good that its just a small change that is easy to reproduce after v3 update.

PR #794, opened 2023-12-14
fix: use `--legacy-peer-deps` in the CI

barjin:
Installing the `@apify-packages/ui-components` package only works with the `--legacy-peer-deps` flag (and it broke our CI).


B4nan (approved):
can we fix that package instead? 👀

barjin:
cc @jbartadev ⬆️ 🙏🏽 ?

Issue #805, opened 2023-12-28
Visual nitpics


valekjo:
- `line-height: tight` in search looks bad
 - double clicking title selects also prefix pill – makes it harder to copypaste
 
<img width="682" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMkNLQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7cb55b47e51398d75383570a36ac30c9adc30639/image.png" alt="image.png" />
<img width="960" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMkdLQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--123aff4398e2eb9c4cff6e548040a642f4103537/image.png" alt="image.png" />

Reported on slack: https://apifier.slack.com/archives/C0L33UM7Z/p1703684285889429

jbartadev:
@webrdaniel Seems like the issue is that it worked locally but not on prod and tooling team seems to be already dealing with it here https://github.com/apify/apify-docs/pull/807

m-murasovs:
cc @webrdaniel 
Reported by Pavel: This also causes some issues for google indexing, so it would be good to keep an eye on this

<img width="373" alt="Screenshot 2024-01-10 at 13 45 20" src="https://github.com/apify/apify-docs/assets/38246758/e27fe0ce-90d9-4b31-b474-2605d022fe60">

B4nan:
Hmm, maybe we should try to replace those badges with empty `<span>` that has the text added via CSS?

webrdaniel:
line-height fixed by https://github.com/apify/apify-docs/pull/807 
For the Crawlee issue created an issue here: https://github.com/apify/crawlee/issues/2282

B4nan:
> For the Crawlee issue

The same problem is present here as well, or better say, in the project docs, e.g. in SDK or client, where we also have API reference.

So I guess let's keep this one open so we don't forget to fix it here too.

webrdaniel:
> > For the Crawlee issue
> 
> The same problem is present here as well, or better say, in the project docs, e.g. in SDK or client, where we also have API reference. Hopefully this can be fixed from inside the shared docusaurus theme which is defined in this repository.
> 
[...]

PR #807, opened 2024-01-03
fix: use `postcss-preset-env` for processing new CSS syntax

### LINKED ISSUE #804: Search does not return correct results for "python"
I find it strange that when I search for “Python” the only thing I am offered is Usage concepts:

obrazek

looks like a frontend bug, the api results are returing a lot of results for python actually (more than 1000 hits)

barjin:
Docusaurus build step cannot process **CSS nesting** yet - not without an extra PostCSS configuration (part of this PR). As a result, the search modal (and possibly other parts of the docs) seemed fine in the dev environment (`docusaurus start`), but broke in prod - see [issue here](https://github.com/facebook/docusaurus/issues/9303).

This PR also bumps the @apify/docs-search-modal version, which fixes the missing results bug.

cc @jbartadev 

Related to #805 
Closes #804


docusaurus.config.js
```diff
@@ -1,3 +1,4 @@
+const postcssPreset = require('postcss-preset-env');
```
B4nan, line 1: don't we need to install this? 

(my favorite review card: undeclared dependency)

PR #821, opened 2024-01-08
fix: make docs work on windows

### LINKED ISSUE #791: It is impossible to setup docs locally on Windows
As of right now it is impossible to setup docs locally on windows.
This might be the culprit 
image

@barjin tagging as discussed

vladfrangu:
Closes #791


tools/utils/collectSlugs.js
```diff
@@ -1,8 +1,32 @@
...
-    return a.split('\n').filter((x) => x.startsWith('slug: ')).map((x) => x.replace('slug: ', ''));
+    const dir = opendirSync(pathname);
+    const res = [];
+
+    let direntry;
+
+    // eslint-disable-next-line no-cond-assign
+    while ((direntry = dir.readSync()) !== null) {
+        if (direntry.isFile() && direntry.name.endsWith('.md')) {
+            const mdContent = readFileSync(`${pathname}/${direntry.name}`, { encoding: 'utf-8' });
```
barjin, line 12: `path.join` would be probably safer here

barjin (approved):
Looks good, thanks for the rewrite :)

B4nan (approved):

docusaurus.config.js
```diff
@@ -34,17 +34,17 @@ module.exports = {
...
                             label: 'Tutorials',
                             to: `/academy/tutorials`,
-                            activeBaseRegex: collectSlugs(`${__dirname}/sources/academy/tutorials`).join('|'),
+                            activeBaseRegex: collectSlugs(`${__dirname}/sources/academy/tutorials`).join('$|'),
                         },
                         {
                             label: 'Glossary',
                             to: `/academy/glossary`,
-                            activeBaseRegex: collectSlugs(`${__dirname}/sources/academy/glossary`).join('|'),
+                            activeBaseRegex: collectSlugs(`${__dirname}/sources/academy/glossary`).join('$|'),
```
barjin, line 47: Also, the last regex in the array won't have a `$` at the end, so it might get treated a bit differently... but this is a complete nit IMO :)

vladfrangu: That is a good point. The main reason I did this is because we had one specific page in the glossary that made both the tutorials and glossary button show as active... I'll fix it so it's consistent

barjin: Yeah, I was wondering whether there was such a case that warranted this change... good catch :)

PR #825, opened 2024-01-11
docs: create new telegram integration guide

TC-MO:
add new screenshots for zapier & telegram integration within zapier
update outdated zapier screenshot
add new telegram integration guide


sources/platform/integrations/telegram.md
```diff
@@ -0,0 +1,115 @@
...
+
+Copy the token and paste it into the configuration form and continue with the "Yes, Continue to Apify" button.
+
+The connection is now created and the configuration form closed.
+
+## Connect Telegram bot with Zapier
+
+### Step 1: Create & connect new bot on Telegram
+
+After setting up Apify as your trigger within Zapier, it's time to set up telegram as your action that will happen based on triger.
```
TheoVasilis, line 74:
```suggestion
After setting up Apify as your trigger within Zapier, it's time to set up Telegram as the action that will occur based on the trigger.
```

TheoVasilis (changes requested):

TheoVasilis (approved):

sources/platform/integrations/telegram.md
```diff
@@ -0,0 +1,115 @@
...
+
+- Finished Actor Run - triggers upon the completion of a selected Actor run.
+
+- Finished Task Run - triggers upon the completion of a selected Actor task run.
+
+![Zapier choosing trigger event](./images/zapier-zap-choose-event.png)
+
+:::info Available Actors & Tasks
+
+Please note that only Actor runs or Actor task runs that you previously executed will be available to choose from.
```
mtrunkat, line 52: Itsn't this

```suggestion
Please note that only Actors or Actor tasks that you previously run will be available to choose from.
```

? You are choosing an Actor or Task not exact run, right? But in fact, is this true? I think that you can now select any Actor or Task without running it previously.

mtrunkat (changes requested):
One thing is not clear to me

sources/platform/integrations/langchain.md
```diff
@@ -1,12 +1,11 @@
 ---
-title: LangChain
+title: 🦜🔗 LangChain
```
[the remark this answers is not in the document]
TC-MO: That is a good point, I should strive to standarize it. In this case I went with `integration` since it was through third party, but maybe we should go with one approach for all of them, either with integration or without, let me think about it.

sources/platform/integrations/telegram.md
TC-MO: Yup, good catch it should have been Actor and Actor tasks not Actor run and Actor task runs.

Regarding the second point, has this changed within the last month since the publication of the video guide? I took the info from the video, and within it we have disclaimer that only Actors that were previously run will show up within Zapier when using Apify as trigger.

TC-MO: I removed integration from title in each case except Telegram, since this one is through 3rd party. I think this approach works best as each doc is already under `Integrations` folder in the sidebar.

sources/platform/integrations/langchain.md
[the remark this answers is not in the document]
TC-MO: I removed integration from title in each case except Telegram, since this one is through 3rd party. I think this approach works best as each doc is already under Integrations folder in the sidebar.

sources/platform/integrations/images/zapier-zap-message-text.png
drobnikj, line 1: I looks there is typo statut instead status

drobnikj (approved):
Just one typo on the screen to fix. Otherwise good 👍

sources/platform/integrations/images/zapier-zap-message-text.png
TC-MO: good catch this is a remnant of previous version and is not been used within the doc , I'll remove it.

sources/platform/integrations/telegram.md
mtrunkat: @drobnikj would know when this was changed - you updated the Zapier integration in autumn to be able to select any Actor, right? So it might be incorrect in the video which is not a big deal IMHO.

mtrunkat (approved):

sources/platform/integrations/telegram.md
TC-MO: I spoke with @drobnikj yesterday, when using Apify as trigger you can only use previously run Actors or Actor tasks. So the information is correct both here and within the video.

PR #827, opened 2024-01-12
feat: Improve section on string interpolation in webhook payloads

valekjo:
This PR improves the section on string interpolation, as discussed on slack.

### DIFF
--- a/sources/platform/integrations/webhooks/actions.md
+++ b/sources/platform/integrations/webhooks/actions.md
@@ -92,7 +92,7 @@ If the string being interpolated only contains the variable, the actual variable
 ```
 
 
-To turn on this new syntax, use "Interpolate variables in string fields" switch. In API it's called `shouldInterpolateStrings`. The field is always `true` when integrating Actors or tasks.
+To turn on this new syntax, use **Interpolate variables in string fields** switch within Apify Console. In API it's called `shouldInterpolateStrings`. The field is always `true` when integrating Actors or tasks.
 
 #### Payload template example
 
--- a/sources/platform/integrations/webhooks/actions.md
+++ b/sources/platform/integrations/webhooks/actions.md
@@ -27,8 +27,6 @@ For safety reasons, the webhook URL should contain a secret token to ensure only
 
 The payload template is a JSON-like string, whose syntax is extended with the use of variables. This is useful when a custom payload structure is needed, but at the same time dynamic data, that is only known at the time of the webhook's invocation, need to be injected into the payload. Aside from the variables, the string must be a valid JSON.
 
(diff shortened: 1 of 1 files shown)


sources/platform/integrations/webhooks/actions.md
```diff
@@ -68,6 +66,34 @@ The syntax of a variable therefore is: `{{oneOfAvailableVariables}}`. The variab
 }
 ```
 
+#### String interpolation
+
+The payload template **is not** a valid JSON by default. The resulting payload is, but not the template. In some cases this is limiting, so there is also updated syntax available, that allows to use templates that provide the same functionality, and are valid JSON at the same time.
```
TC-MO, line 71:
```suggestion
The payload template _is not_ a valid JSON by default. The resulting payload is, but not the template. In some cases this is limiting, so there is also updated syntax available, that allows to use templates that provide the same functionality, and are valid JSON at the same time.
```

sources/platform/integrations/webhooks/actions.md
```diff
@@ -68,6 +66,34 @@ The syntax of a variable therefore is: `{{oneOfAvailableVariables}}`. The variab
...
+
+If the string being interpolated only contains the variable, the actual variable value is used in the payload. For example `"{{eventData}}"` results in an object. If the string contains more than just the variable, the string value of the variable will occur in the payload:
+
+```json
+{ "text": "My user id is {{userId}}" }
+{ "text": "My user id is abf6vtB2nvQZ4nJzo" }
+```
+
+
+To turn on this new syntax, use "Interpolate variables in string fields" switch. In API it's called `shouldInterpolateStrings`. The field is always `true` when integrating Actors or tasks.
```
TC-MO, line 95:
```suggestion
To turn on this new syntax, use **Interpolate variables in string fields** switch within Apify Console. In API it's called `shouldInterpolateStrings`. The field is always `true` when integrating Actors or tasks.
```

sources/platform/integrations/webhooks/actions.md
TC-MO: Emphasis should be done with italics bold should be usually used for noting UI elements when describing something.

sources/platform/integrations/webhooks/actions.md
TC-MO: I am correct to assume that this refers to this switch within Console? 
image

mvolfik (approved):
Love it, thanks a lot! Maybe worth noting that in A2A only the second syntax is available? Or is that already explained in the docs page?

TC-MO:
I see that Vale is failing, fix for that issue is on another PR, waiting to be approved & merged

valekjo:
> Love it, thanks a lot! Maybe worth noting that in A2A only the second syntax is available? Or is that already explained in the docs page specific for A2A?

@mvolfik 

It's noted there as "The field is always `true` when integrating Actors or tasks.". Do you have another idea how to phrase it?

TC-MO (approved):

mvolfik:
Ah, missed that part. Maybe we could link to this section from https://docs.apify.com/platform/integrations/actors, but either way, LGTM

PR #828, opened 2024-01-12
chore: bump apify ui components to an existing version

vladfrangu:
In the time where I fixed the repo to be usable on windows and actually testing, it seems like the version expired (?) or was deleted. Either way, bumping its patch only (as it uses 0ver and any further bumps technically are major)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -85,6 +85,16 @@
                 "ajv": "^8.8.2"
             }
         },
+        "apify-docs-theme/node_modules/axios": {
+            "version": "1.6.5",
+            "resolved": "https://registry.npmjs.org/axios/-/axios-1.6.5.tgz",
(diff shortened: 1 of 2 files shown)


B4nan (approved):
hmm, this is so fragile, we should really make this package public so we can keep the versions. if the changes wouldnt be breaking for us, i wouldnt mind, but sometimes they are...

cc @jbartadev is this something your team can handle or can we somehow help/do it ourselves?

vladfrangu:
Fun times ahead. Bumping this somehow broke dev completely, because ui-components's dependencies for some reason import apify/log down the chain, a module that was never intended for browsers from the looks of things

jbartadev:
> hmm, this is so fragile, we should really make this package public so we can keep the versions. if the changes wouldnt be breaking for us, i wouldnt mind, but sometimes they are...
> 
> cc @jbartadev is this something your team can handle or can we somehow help/do it ourselves?

Yeah, let's make it public, it looks like it'll save us a lot of pain in the future. It'll require some dependencies removal from the package. We'll try to take a look at it ASAP.

vladfrangu:
I'll merge this as is for now, looking forward to moving to the public components 🙏

PR #830, opened 2024-01-13
docs: fix_commands_in_docs

souravjain540:
Screenshot 2024-01-14 at 1 18 16 AM
It should be only one `apify` in each of them


B4nan (approved):
thanks

mtrunkat (approved):
Good catch

Issue #831, opened 2024-01-15
Documentation Links for API Clients Redirect to Incorrect Languages (JavaScript and Python)


UsmanLiaqat404:
![image](https://github.com/apify/apify-docs/assets/117591938/25f66b67-f787-4499-89d6-6593572035cf)

when I click on [API client for JavaScript](https://docs.apify.com/api/client/python) it redirects me to Python and when I click on [API client for Python](https://docs.apify.com/api/client/js) it redirects me to js

UsmanLiaqat404:
PR created: https://github.com/apify/apify-docs/pull/832

PR #832, opened 2024-01-15
docs: Fix Documentation Links for API Clients Redirect to Incorrect Languages (JavaScript and Python)

UsmanLiaqat404:
Fixed Issue: https://github.com/apify/apify-docs/issues/831


B4nan (approved):
thanks, unfortunately the docs build is currently broken because of some internal dependencies, so the changes will be deployed only once we fix that

PR #839, opened 2024-01-17
docs: add lychee github action for test purposes

TC-MO:


barjin:
### Context for other reviewers:
- The setup in this PR currently checks for broken link URLs in the `.md` / `.mdx` files of the project.
- This is slightly dangerous - during build, [`@docusaurus/plugin-content-docs`](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs) compiles those markdown files, replacing (among others) [Markdown links](https://docusaurus.io/docs/markdown-features/links) - something that Lychee doesn't do and fails because of:
> obrazek
   See that there are also issues with mixing HTML and JS in the `.mdx` syntax.

### Proposed solution:
- We can run the Docusaurus build first and analyze the generated `html` files.
   - Note that this might introduce new issues (e.g. this with different names of newly generated JS bundles):
     > obrazek
     > this is probably easily removed with some `--exclude` directives
- It still might consider an article added to the tested PR as missing (as Lychee is taking the links from the "local version", but checking for the resources online, in the currently deployed version, right?)

TC-MO:
* If we would be able to check links only after build process is done this would save us a lot of headache with fixing file-path based links

* To the point of Lychee considering new page as false positive
	* While that is true, this also is an edge case that happens rarely that we would be linking to a new page from older docs, and even if we would to do this, we can just make some workaround for this.

TC-MO:
I thought that with this new approach we will have way less errors, since the links will be built by docusaurus, but if I see it correctly we have 10x the amount of checked, but similar count of errors? 

obraz

barjin:
When you look at the single documentation page, it contains many more links than the original markdown document - navigation bars, `previous` / `next` article or `Edit this page` links leading to the GitHub repository of the docs. The hike up in the links count is a result of this (and is IMO correct, as now Lychee tests every single link a user can click, not only the ones directly in the content). 

The `Edit this page` link also causes (most|many) of the issues - GitHub is perhaps not too happy with the Lychee HTTP client accessing it (seems like a bot). Maybe it's safe to ignore those links (`https://github.com/apify/apify-docs/edit/master/...`), as we can assume they are automatically 1:1 generated - and that they are also for internal use only? Without those, it seems to only report some actual, more interesting issues.

TC-MO:
Ah yes, fair enough the point about lychee now checking also navbars and other clickable content makes perfect sense.

* As for `Edit this page` , I don't think this is for internal use only, as our docs are open source, technically anyone can click on them and create fork/pr to our docs, at the same since those are generated by Docusaurus, if I'm not mistaken, we can safely assume that they are always ok and ignore those in link checking? Or we could just accept `429` status code as valid ? But that seems like not the greatest approach.

TC-MO:
One other thing to consider is that it seems that we have a lot of false positives in the lychee report since a lot of our links are linking to `.md` files but relatively for example in file `sources/academy/webscraping/web_scraping_for_beginners/data_extraction/index.md` we are linking to other file using `./browser_devtools.md`. Since we are using `-- base` argument in lychee it tries to check it against `docs.apify.com` and fails. But, if the link would be `rces/academy/webscraping/web_scraping_for_beginners/data_extraction/browser_devtools.md` then it seems that lychee checks it as internal md link instead trying to check it against docs.apify.com

B4nan:
> One other thing to consider is that it seems that we have a lot of false positives in the lychee report since a lot of our links are linking to .md files but relatively for example in file sources/academy/webscraping/web_scraping_for_beginners/data_extraction/index.md we are linking to other file using ./browser_devtools.md. Since we are using -- base argument in lychee it tries to check it against docs.apify.com and fails. But, if the link would be sources/academy/webscraping/web_scraping_for_beginners/data_extraction/browser_devtools.md then it seems that lychee checks it as internal md link instead trying to check it against docs.apify.com. 

How is that relevant? The relative paths are transformed to actual links on build time, lychee won't test those but the actual rendered links, right? Sounds like a misconfiguration to me.

TC-MO:
I think this boils down to which approach we choose:

1. Check after build
	This way we check a lot of extra stuff like navbar, github edit links etc. But we can filter them out with proper configuration and we can think less about how we link within our content (but still should default to one way of doing it and make all the links adhere to this way)
	
2. Check `.md` and `.mdx` files 
	This way we won't be checking all the extra stuff from the built page, just links within content itself. This approach tho necessitates that we fix all relative links md links because lychee tries to check them against `docs.apify.com`. Unless I'm mistaken about lychee and the way it tries to check the relative links when using `--base` argument .

B4nan:
> Check .md and .mdx files

Keep in mind docusaurus is already checking those links. A relative path link is always pointing to something local which is exactly the thing that docusaurus verifies (and breaks the CI checks if it fails)

> This way we won't be checking all the extra stuff from the built page, just links within content itself. This approach tho necessitates that we fix all relative links md links because lychee tries to check them against docs.apify.com. Unless I'm mistaken about lychee and the way it tries to check the relative links when using --base argument .

I disagree this is "something to fix", the links are not broken.

B4nan:
Found this example repo:

https://github.com/StarRocks/docusaurus-build

Which explicitly skips the links pointing to `.md` from lychee, sounds like a good approach to me:


      - name: link check
        if: always()
        uses: lycheeverse/lychee-action@v1.8.0
        with:
          fail: true
          args: >
            --config docs/lychee.toml
            --offline "docs/**/*.md"

TC-MO:
Fix was in a sense that it does throw false positives.

I don't think `--offline` will work in our case. From what I understand it blocks network requests, but we have links to other parts of documentation that need checking (docs > sdk docs for example) and external links that also would be nice to check.

TC-MO:
I think there might have been some misunderstanding regarding the approach to link checking with Lychee.

We're dealing with three types of links in our documentation:
1. `./foo/bar.md` - These are relative links within our project. They correctly resolve post-build but currently trigger false positives in Lychee because they're checked against our `--base` URL, leading to errors.
2. `/foo/bar` - These are relative URLs pointing to other parts of our documentation outside this repository (e.g., from docs to SDK docs or CLI docs). To properly check these, we need to use the `--base` argument.
3. Standard external links - These link to pages outside our documentation and are not problematic in our current setup.

The challenge arises with how Lychee handles the first type of links when the `--base` argument is applied. It misinterprets them and tries to resolve them against `--base` URL, hence the false positives.

We've identified two potential solutions:
1. **Post-build Lychee check**: This method seems to address the issue of false positives for internal links. However, it introduces new challenges, such as the need to exclude GitHub links to avoid rate limiting and additional false positives from external sources.
2. **Pre-build Lychee check with modifications to internal links**: By changing our internal linking strategy from `./foo/bar.md` to a more absolute path like `/foo/bar/foobar/bar.md`, we can prevent Lychee from misinterpreting them. This adjustment should eliminate the false positives without requiring changes to our use of the `--base` argument.

B4nan:
I don't think this is about misunderstanding, more like I don't like what you are saying and I have a hard time believing there isn't a better way that does not involve rewriting the content.

> ./foo/bar.md - These are relative links within our project. They correctly resolve post-build but currently trigger false positives in Lychee because they're checked against our --base URL, leading to errors.

Yes, and the solution to this is to skip those from lychee, it feels like you don't want to do that, but it is the correct fix.

> /foo/bar - These are relative URLs pointing to other parts of our documentation outside this repository (e.g., from docs to SDK docs or CLI docs). To properly check these, we need to use the --base argument.

Not necessarily, they can point to the same docusaurus instance as well, but yes. 

> The challenge arises with how Lychee handles the first type of links when the --base argument is applied. It misinterprets them and tries to resolve them against --base URL, hence the false positives.

Ignore rules that need to be applied to the source, not to post-processed URLs (so this needs to ignore anything, including the `base` arg), I am sure lychee allows to do that.

Maybe the `offline` option is not exactly what we need, but given the linked repo is dealing with exactly this problem, it feels like it might do what we need. We should first try that instead of reasoning about its name and description and whatnot. In the end, checking the links without network access sounds exactly like "skipping the actual problematic check", not sure what it checks in this mode, I'd guess it's gonna be more like an ignore rule rather than offline rule... I will take a closer look myself during the afternoon.

Changing the URLs would be a step back to me - the paths are pretty handy, as they decouple the content from the URL, and they provide intellisense/autocomplete too.

TC-MO:
Doesn't [docs.apify.com/api/v2](https://docs.apify.com/api/v2/) always return `404` and only after `404` loads the API ref? At least that is how I always saw it when entering API Reference for the first time.

barjin:
That doesn't seem to be the case - check the `Network` tab while accessing the URLs with and without the trailing slash:

obrazek

Let's not digress too much from the Lychee thing - we'll have enough time to fix broken links after we know where they are 😄

B4nan:
> Doesn't [docs.apify.com/api/v2](https://docs.apify.com/api/v2/) always return 404 and only after 404 loads the API ref? At least that is how I always saw it when entering API Reference for the first time.

`https://docs.apify.com/api/v2` works fine, as that is the actual URL of the page. If you add the trailing slash `https://docs.apify.com/api/v2/`, you get 404 (which is imo correct).

TC-MO:
Since we are moving forward with https://github.com/apify/apify-docs/pull/855 I'll go ahead and close this PR

PR #857, opened 2024-02-13
docs: (even more) explicitly mention Apiary when updating API reference

tobice:
It's already mentioned in the pre-release section just below but I missed it anyway 🙈


drobnikj (approved):
🆒

PR #861, opened 2024-02-14
chore: update eslint to check codeblocks in all markdown files

B4nan:
We had an eslint setup in place for the academy only, now its merged with the root eslint and polished a bit, plus the apify eslint config is updated and the code is modified to comply with them.

The main change that applies to the markdown files is that many of them were incomplete or contained "pseudo JSON", e.g. with trailing commas or with some placeholders. Those are mostly moved to `json5` which is more loose and works fine in this regard.


B4nan:
@TC-MO since this modifies a lot of files, it triggers a lot of unrelated vale checks, I'd ignore them and just merge it, I hope you are fine with that :]

TC-MO:
LGTM

TC-MO (approved):

PR #866, opened 2024-02-22
feat(academy): add 'How to retry failed request' article

metalwarrior665:


sources/academy/tutorials/api/retry_failed_requests.md
```diff
@@ -0,0 +1,61 @@
+---
+title: How to retry failed requests
+description: Learn how to resurrect your run but retrying only failed requests
+sidebar_position: 6
+slug: /api/retry-failed-requests
+---
+
+**Learn how to re-scrape only failed requests in your run.**
```
AndreyBykov, line 8: `of your run` maybe sounds better?

sources/academy/tutorials/api/retry_failed_requests.md
```diff
@@ -0,0 +1,61 @@
...
+
+---
+
+There are many reasons why requests for a scraper could fail. The most popular causes are different page layouts or proxy blocking issues ([check here on how to effectively analyze errors](https://docs.apify.com/academy/node-js/analyzing-pages-and-fixing-errors)). Both Apify and [Crawlee](https://crawlee.dev/) allow you to restart your scraper run from the point where it ended but there is no native functionality to re-scrape only failed requests. Usually, you also want to first analyze the problem, update the code and build it before trying again.
```
AndreyBykov, line 12: > different page layouts

I would maybe write it a bit differently: `changes to a page layout (compared to when the scraper was developed)`/`updated page layout (compared to when the scraper was developed)` or something similar.

differnt page layouts sounds a bit vague - like different to what/how?

sources/academy/tutorials/api/retry_failed_requests.md
```diff
@@ -0,0 +1,61 @@
...
+
+If you attempt to re-run an already finished run, it will likely immediately finish because all the requests in the [request queue](https://crawlee.dev/docs/3.7/guides/request-storage) are marked as handled. So you need to update the failed requests in the queue to be marked as pending again.
```
AndreyBykov, line 14: Let's not use the specific version :)
```suggestion
If you attempt to re-run an already finished run, it will likely immediately finish because all the requests in the [request queue](https://crawlee.dev/docs/guides/request-storage) are marked as handled. So you need to update the failed requests in the queue to be marked as pending again.
```

sources/academy/tutorials/api/retry_failed_requests.md
```diff
@@ -0,0 +1,61 @@
...
+
+const REQUEST_QUEUE_ID = 'pFCvCasdvsyvyZdfD'; // Replace with your valid request queue ID
+const allRequests = [];
+let exclusiveStartId = null;
+// List all requests from the queue, we have to do it in a loop because the request queue list is paginated
+for (; ;) {
+    const { items: requests } = await Actor.apifyClient.requestQueue(REQUEST_QUEUE_ID).listRequests({ exclusiveStartId, limit: 1000 });
+    for (const request of requests) {
+        allRequests.push(requests);
+    }
```
AndreyBykov, line 32: Or alternatively, if we want to remove some extra lines :)
```suggestion
    allRequests.push(...requests);
```

sources/academy/tutorials/api/retry_failed_requests.md
```diff
@@ -0,0 +1,61 @@
...
+import { Actor } from 'apify';
+
+const REQUEST_QUEUE_ID = 'pFCvCasdvsyvyZdfD'; // Replace with your valid request queue ID
+const allRequests = [];
+let exclusiveStartId = null;
+// List all requests from the queue, we have to do it in a loop because the request queue list is paginated
+for (; ;) {
+    const { items: requests } = await Actor.apifyClient.requestQueue(REQUEST_QUEUE_ID).listRequests({ exclusiveStartId, limit: 1000 });
+    for (const request of requests) {
+        allRequests.push(requests);
```
AndreyBykov, line 31: Either one `request`, not all of them :)
```suggestion
        allRequests.push(request);
```

sources/academy/tutorials/api/retry_failed_requests.md
```diff
@@ -0,0 +1,61 @@
...
+```typescript
+// The code is similar for both Crawlee-only but uses a different API
+import { Actor } from 'apify';
+
+const REQUEST_QUEUE_ID = 'pFCvCasdvsyvyZdfD'; // Replace with your valid request queue ID
+const allRequests = [];
+let exclusiveStartId = null;
+// List all requests from the queue, we have to do it in a loop because the request queue list is paginated
+for (; ;) {
+    const { items: requests } = await Actor.apifyClient.requestQueue(REQUEST_QUEUE_ID).listRequests({ exclusiveStartId, limit: 1000 });
```
AndreyBykov, line 29: Just a suggestion to split it between the lines to avoid the long lines and the need for scrolling or smaller screens
```suggestion
    const { items: requests } = await Actor.apifyClient
        .requestQueue(REQUEST_QUEUE_ID)
        .listRequests({ exclusiveStartId, limit: 1000 });
```

sources/academy/tutorials/api/retry_failed_requests.md
```diff
@@ -0,0 +1,61 @@
...
+    for (const request of requests) {
+        allRequests.push(requests);
+    }
+    // If we didn't get full 1000 requests, we have all and can finish the loop
+    if (requests.length < 1000) {
+        break;
+    }
```
AndreyBykov, line 36: Not sure if want to do it, but I would also inline it there.
```suggestion
    if (requests.length < 1000) break;
```

sources/academy/tutorials/api/retry_failed_requests.md
```diff
@@ -0,0 +1,61 @@
...
+    request.handledAt = null;
+    await Actor.apifyClient.requestQueue(REQUEST_QUEUE_ID).updateRequest(request);
+}
+
+// And now we can resurrect our scraper again, it will only process the failed requests
+```
+
+## Resurrect automatically with a free public Actor {#resurrect-automatically-with-a-free-public-actor}
+
+Fortunately, you don't need to implement this code into your workflow. [Apify Store](https://apify.com/store) provides the [Rebirth Failed [Requests](https://apify.com/lukaskrivka/rebirth-failed-requests) actor ((that is [open-source](https://github.com/metalwarrior665/rebirth-failed-requests))) that does this and more. The actor can automatically scan multiple runs of your Actors based on filters like date started. It can also automatically resurrect the runs after renewing the failed requests. That means you will finish your scrape into the final successful state with a single click on the Run button.
```
AndreyBykov, line 61:
```suggestion
Fortunately, you don't need to implement this code into your workflow. [Apify Store](https://apify.com/store) provides the [Rebirth Failed Requests](https://apify.com/lukaskrivka/rebirth-failed-requests) Actor (that is [open-source](https://github.com/metalwarrior665/rebirth-failed-requests)) that does this and more. The Actor can automatically scan multiple runs of your Actors based on filters like `date started`. It can also automatically resurrect the runs after renewing the failed requests. That means you will finish your scrape into the final successful state with a single click on the Run button.
```

AndreyBykov (approved):

sources/academy/tutorials/api/retry_failed_requests.md
TheoVasilis, line 12:
```suggestion
There are many reasons why requests for a scraper could fail. The most common causes are different page layouts or proxy blocking issues ([check here on how to effectively analyze errors](https://docs.apify.com/academy/node-js/analyzing-pages-and-fixing-errors)). Both [Apify](https://apify.com) and [Crawlee](https://crawlee.dev/) allow you to restart your scraper run from the point where it ended, but there is no native functionality to re-scrape only failed requests. Usually, you also want to first analyze the problem, update the code, and build it before trying again.
```

sources/academy/tutorials/api/retry_failed_requests.md
TheoVasilis, line 14:
```suggestion
If you attempt to restart an already finished run, it will likely immediately finish because all the requests in the [request queue](https://crawlee.dev/docs/3.7/guides/request-storage) are marked as handled. So, you need to update the failed requests in the queue to be marked as pending again.
```

sources/academy/tutorials/api/retry_failed_requests.md
```diff
@@ -0,0 +1,61 @@
...
+
+If you attempt to re-run an already finished run, it will likely immediately finish because all the requests in the [request queue](https://crawlee.dev/docs/3.7/guides/request-storage) are marked as handled. So you need to update the failed requests in the queue to be marked as pending again.
+
+The additional complication is that the [Request](https://crawlee.dev/api/core/class/Request) object doesn't have anything like `isFailed` property, we have to approximate it using other fields. Fortunately, we can use the `errorMessages` and `retryCount` properties to identify failed requests. Unless the user explicitly overrode these properties, we can identify failed requests by having a larger amount of `errorMessages` than `retryCount`. That happens because the last error that doesn't cause a retry anymore is added to `errorMessages`.
```
TheoVasilis, line 16:
```suggestion
The additional complication is that the [Request](https://crawlee.dev/api/core/class/Request) object doesn't have anything like the `isFailed` property. We have to approximate it using other fields. Fortunately, we can use the `errorMessages` and `retryCount` properties to identify failed requests. Unless the user explicitly has overridden these properties, we can identify failed requests with a larger amount of `errorMessages` than `retryCount`. That happens because the last error that doesn't cause a retry anymore is added to `errorMessages`.
```

sources/academy/tutorials/api/retry_failed_requests.md
```diff
@@ -0,0 +1,61 @@
...
+const REQUEST_QUEUE_ID = 'pFCvCasdvsyvyZdfD'; // Replace with your valid request queue ID
+const allRequests = [];
+let exclusiveStartId = null;
+// List all requests from the queue, we have to do it in a loop because the request queue list is paginated
+for (; ;) {
+    const { items: requests } = await Actor.apifyClient.requestQueue(REQUEST_QUEUE_ID).listRequests({ exclusiveStartId, limit: 1000 });
+    for (const request of requests) {
+        allRequests.push(requests);
+    }
+    // If we didn't get full 1000 requests, we have all and can finish the loop
```
TheoVasilis, line 33:
```suggestion
    // If we didn't get the full 1,000 requests, we have all and can finish the loop
```

TheoVasilis: I find the 'we have all' part confusing. Is it referring to a specific status in the code? Otherwise, it doesn't sound right, but as it seems to contradict the previous clause (we didn't get vs we have all), I'm not sure what alternative to suggest.

sources/academy/tutorials/api/retry_failed_requests.md
```diff
@@ -0,0 +1,61 @@
...
+// We need to update them 1 by 1 to the pristine state
+for (const request of failedRequests) {
+    request.retryCount = 0;
+    request.errorMessages = [];
+    // This tells the request queue to handle it again
+    request.handledAt = null;
+    await Actor.apifyClient.requestQueue(REQUEST_QUEUE_ID).updateRequest(request);
+}
+
+// And now we can resurrect our scraper again, it will only process the failed requests
```
TheoVasilis, line 56:
```suggestion
// And now we can resurrect our scraper again; it will only process the failed requests.
```

TheoVasilis (changes requested):

sources/academy/tutorials/api/retry_failed_requests.md
AndreyBykov: It goes in a loop while fetching batches of 1000 requests. If we get 1000 - it means there's more and we should continue the loop, if we have less than 1000 - it means that it was the last batch and we got the leftovers, so nothing else to fetch.

AndreyBykov: So the meaning here is something like:
// If we didn't get the full 1,000 requests of the current batch, we have fetched all the requests from the queue and can finish the loop

sources/academy/tutorials/api/retry_failed_requests.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+## Resurrect automatically with a free public Actor {#resurrect-automatically-with-a-free-public-actor}
```
TC-MO, line 59: I think you can forego in this case the hard coded anchor since it is the same as heading, and it's giving out Vale errors

sources/academy/tutorials/api/retry_failed_requests.md
TC-MO, line 61:
```suggestion
Fortunately, you don't need to implement this code into your workflow. [Apify Store](https://apify.com/store) provides the [Rebirth Failed Requests](https://apify.com/lukaskrivka/rebirth-failed-requests) Actor ((that is [open-source](https://github.com/metalwarrior665/rebirth-failed-requests))) that does this and more. The Actor can automatically scan multiple runs of your Actors based on filters like date started. It can also automatically resurrect the runs after renewing the failed requests. That means you will finish your scrape into the final successful state with a single click on the Run button.
```

sources/academy/tutorials/api/retry_failed_requests.md
TC-MO, line 12:
```suggestion
Request for a scraper could fail for many reasons. The most popular causes are different page layouts or proxy blocking issues ([check here on how to effectively analyze errors](https://docs.apify.com/academy/node-js/analyzing-pages-and-fixing-errors)). Both Apify and [Crawlee](https://crawlee.dev/) allow you to restart your scraper run from the point where it ended but there is no native functionality to re-scrape only failed requests. Usually, you also want to first analyze the problem, update the code and build it before trying again.
```

sources/academy/tutorials/api/retry_failed_requests.md
TC-MO, line 14:
```suggestion
If you attempt to re-run an already finished run, it will likely immediately finish because all the requests in the [request queue](https://crawlee.dev/docs/3.7/guides/request-storage) are marked as handled, necessitating an update of the failed requests in the queue to be marked as pending again..
```

sources/academy/tutorials/api/retry_failed_requests.md
metalwarrior665: I never inline `if` :)

TheoVasilis (approved):

PR #868, opened 2024-02-23
docs: update MonthlyUsage and AccountLimits in API reference

tobice:
Changes in `MonthlyUsage`:
- add `priceTiers`, which were missing
- remove `amountAfterVolumeDiscountUsd`, which is no longer served

Changes in `AccountLimits`:
- add recent new `maxConcurrentActorJobs` and `activeActorJobsCount`
- add `dataRetentionDays`, which were missing

I have checked both the actual response and API implementation for the latest state.


jirimoravcik (approved):
nice

PR #872, opened 2024-02-28
docs: rewrite + fixes for Vale & mdlint

TC-MO:
Slight rewrite to remove links to non existing actors and their KV Store 
fixes for Vale
fixes for mdlint

### DIFF
--- a/sources/academy/tutorials/node_js/block_requests_puppeteer.md
+++ b/sources/academy/tutorials/node_js/block_requests_puppeteer.md
@@ -5,27 +5,27 @@ sidebar_position: 16.2
 slug: /node-js/block-requests-puppeteer
 ---
 
-_Unfortunately, in the recent version of Puppeteer, request interception disables native cache and slows down the actor significantly. Therefore it is not recommended to follow examples shown in this article. Instead, use_ [_blockRequests_](/sdk/js/docs/api/puppeteer#puppeteerblockrequestspage-options-promise) _utility function from_ [_Apify SDK_](/sdk/js/)_. It works through different paths and doesn't slow down your process._
+:::caution Improve Performance: Use blockRequests`
+Unfortunately, in the recent version of Puppeteer, request interception disables native cache and slows down the Actor significantly. Therefore, it's not recommended to follow examples shown in this article. Instead, use [`blockRequests`](https://crawlee.dev/api/puppeteer-crawler/namespace/puppeteerUtils#BlockRequestsOptions) _utility function from_ [_Apify SDK_](/sdk/js/). It works through different paths and doesn't slow down your process.
+:::
 
 When using Puppeteer, often a webpage will load many resources that are not actually necessary for your use case. For example page could be loading many tracking libraries, that are completely unnecessary for most crawlers, but will cause the page to use more traffic and load slower.
 
 For example for this web page: <https://edition.cnn.com/>
-If we run an [Actor](https://www.apify.com/jaroslavhejlek/measure-downloaded-bytes) that measures extracted downloaded data from each response until the page is fully loaded, we get these results:
+If we run an Actor that measures extracted downloaded data from each response until the page is fully loaded, we get these results:
 
 ![Actor loading](./images/actor-load.png)
 
 
-And [this is how the website it looks](https://api.apify.com/v2/key-value-stores/sE2s9WmvoWFZhTff7/records/debug-screen.png?disableRedirect=true).
+Now if we want to optimize this to keep the webpage looking the same, but ignore unnecessary requests, then after
 
-Now if we want to optimise this to keep the webpage looking the same, but ignore unnecessary requests, then after
-
-```js
+```js showLineNumbers
 const page = await browser.newPage();
 ```
 
 we could can use this piece of code
 
-```js
+```js showLineNumbers
 await page.setRequestInterception(true);
 page.on('request', (request) => {
     if (someCondition) request.abort();
@@ -39,7 +39,7 @@ For our example we will only disable some tracking scripts and then check if eve
 
 Here is the code used:
 
-```js
+```js showLineNumbers
 await page.setRequestInterception(true);
 page.on('request', (request) => {
     const url = request.url();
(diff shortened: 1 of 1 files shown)


sources/academy/tutorials/node_js/block_requests_puppeteer.md
```diff
@@ -5,27 +5,27 @@ sidebar_position: 16.2
 slug: /node-js/block-requests-puppeteer
 ---
 
-_Unfortunately, in the recent version of Puppeteer, request interception disables native cache and slows down the actor significantly. Therefore it is not recommended to follow examples shown in this article. Instead, use_ [_blockRequests_](/sdk/js/docs/api/puppeteer#puppeteerblockrequestspage-options-promise) _utility function from_ [_Apify SDK_](/sdk/js/)_. It works through different paths and doesn't slow down your process._
+:::caution Improve Performance: Use blockRequests`
```
B4nan, line 8: have you tried this locally? i remember docusaurus admonitions required a blank line here and before the closing `:::`, otherwise it failed to parse it properly. but maybe something changed, or it was a different kind of problem (maybe even something prettier related)

B4nan: yeah its about prettier (which we currently dont use here currently, but we are planning to adopt biome in the upcoming weeks/months)

https://docusaurus.io/docs/next/markdown-features/admonitions#usage-with-prettier

TC-MO: Seems fine locally, TBH never had an issue without blank lines in admonitions
obraz

sources/academy/tutorials/node_js/block_requests_puppeteer.md
```diff
@@ -5,27 +5,27 @@ sidebar_position: 16.2
...
+Unfortunately, in the recent version of Puppeteer, request interception disables native cache and slows down the Actor significantly. Therefore, it's not recommended to follow examples shown in this article. Instead, use [<code>blockRequests</code>](https://crawlee.dev/api/puppeteer-crawler/namespace/puppeteerUtils#BlockRequestsOptions) _utility function from_ [_Apify SDK_](/sdk/js/). It works through different paths and doesn't slow down your process.
```
metalwarrior665, line 9:
```suggestion
Unfortunately, in the recent version of Puppeteer, request interception disables the native cache and slows down the Actor significantly. Therefore, it's not recommended to follow the examples shown in this article. Instead, use [<code>blockRequests</code>](https://crawlee.dev/api/puppeteer-crawler/namespace/puppeteerUtils#BlockRequestsOptions) _utility function from_ [_Crawlee_](https://crawlee.dev). It works through different paths and doesn't slow down your process.
```

sources/academy/tutorials/node_js/block_requests_puppeteer.md
B4nan: that screenshot actually shows another problem on this line i havent seen before, you got a trailing `` ` `` there :]

metalwarrior665 (changes requested):
Just this small thing and you can merge

sources/academy/tutorials/node_js/block_requests_puppeteer.md
TC-MO: ah. good catch, thanks

sources/academy/tutorials/node_js/block_requests_puppeteer.md
B4nan: let's also add the blank lines, we use them everywhere already, lets be consistent (and lets prepare for the future, as we would be dealing with this the minute we try to add some code formatting)

sources/academy/tutorials/node_js/block_requests_puppeteer.md
```diff
@@ -5,27 +5,27 @@ sidebar_position: 16.2
 slug: /node-js/block-requests-puppeteer
 ---
 
-_Unfortunately, in the recent version of Puppeteer, request interception disables native cache and slows down the actor significantly. Therefore it is not recommended to follow examples shown in this article. Instead, use_ [_blockRequests_](/sdk/js/docs/api/puppeteer#puppeteerblockrequestspage-options-promise) _utility function from_ [_Apify SDK_](/sdk/js/)_. It works through different paths and doesn't slow down your process._
+:::caution Improve Performance: Use `blockRequests`
+Unfortunately, in the recent version of Puppeteer, request interception disables the native cache and slows down the Actor significantly. Therefore, it's not recommended to follow the examples shown in this article. Instead, use [<code>blockRequests</code>](https://crawlee.dev/api/puppeteer-crawler/namespace/puppeteerUtils#BlockRequestsOptions) _utility function from_ [_Crawlee_](https://crawlee.dev). It works through different paths and doesn't slow down your process.
+:::
```
B4nan, line 10: it needs to be at the bottom as well
```suggestion

:::
```

TC-MO: yup already fixed , thank you

metalwarrior665 (approved):

sources/academy/tutorials/node_js/block_requests_puppeteer.md
```diff
@@ -5,27 +5,29 @@ sidebar_position: 16.2
...
 ![Actor loading](./images/actor-load.png)
 
 
-And [this is how the website it looks](https://api.apify.com/v2/key-value-stores/sE2s9WmvoWFZhTff7/records/debug-screen.png?disableRedirect=true).
-
-Now if we want to optimise this to keep the webpage looking the same, but ignore unnecessary requests, then after
+Now if we want to optimize this to keep the webpage looking the same, but ignore unnecessary requests, then after
 
-```js
+```js showLineNumbers
```
B4nan, line 24: side note - do you think we should enable line numbers globally? docusaurus does not allow it right now, but we could try this [css hack](https://gist.github.com/uditalias/3969a20ef759bf3a8757499429b30063), or have a custom code block component override to add it (that could probably also allow opting out of them)

B4nan: fyi this is how it looks with the css solution

image

B4nan: here is a PR doing that globally, i'd probably go with that approach and remove the explicit marks you added here so we don't end up with double numbering 

#873

TC-MO: Generally speaking I'm a fan of having numbered lines. I think it just make the code samples look cleaner.
I think that is a good idea, I will remove those from here, and will search for others that may already be in the docs in separate PR and then we could go forward with merging the #873

PR #873, opened 2024-02-28
feat: add line numbers to code examples via CSS globally

B4nan:
example: 
image


TC-MO:
Seems that we do not have `showLineNumbers` anywhere so we can move forward with this change to add them through CSS

TC-MO:
Will do right now, as I am doing a PR with changes to style guide regarding admonitions

PR #879, opened 2024-03-06
docs: rewrite of flowise integration

TC-MO:
add new images
rewrite for clarity
add new headings


sources/platform/integrations/flowise.md
```diff
@@ -9,28 +9,50 @@ slug: /integrations/flowise
...
+    npx flowise start
+    ```
+
+It will be available on `https://localhost:3000`
 
-If you have any experience with Flowise, you need to download and run it locally.
-You can find the instructions [here](https://github.com/FlowiseAI/Flowise#quick-start).
+Other methods of using Flowise can be found in their [documentation](https://docs.flowiseai.com/getting-started#quick-start)
+
+### Building your flow
```
janbuchar, line 38:
```suggestion
### Building your Flow
```

This is kinda off topic and I'm not even sure what I suggest is the correct capitalization. Do we conform to some particular style here?

sources/platform/integrations/flowise.md
```diff
@@ -39,6 +61,8 @@ The final flow can answer questions about Apify docs.
 
 ![Flowise and Apify](images/flowise-2.png)
 
+For more information visit the Flowise's [documentation](https://flowiseai.com/).
```
janbuchar, line 64:
```suggestion
For more information visit the Flowise [documentation](https://flowiseai.com/).
```

Probably a matter of personal taste

janbuchar (approved):
Much better, thank you!

sources/platform/integrations/flowise.md
TC-MO: I think if Flowise ever referred to it with capitalization we would do it too, but I checked out their docs and haven't seen them referring to it as Flow so I would go without capitalization here

PR #883, opened 2024-03-14
docs: remove sharing info from the Schedules page

tobice:
Schedules cannot be shared at the moment.


TC-MO (approved):

tobice:
@drobnikj FYI went on with the merge anyway, I don't think your input is crucial here. Let me know in case you have concerns 😊

PR #888, opened 2024-03-21
fix: update og-image url

webrdaniel:
Changing the path to og-image because it was updated on the web.


jbartadev (approved):
Thanks! 🚀

barjin (approved):
lgtm, cheers :)

PR #890, opened 2024-03-21
feat: Setup Smartlook [internal]

HonzaTuron:
Closes apify/apify-web#3489

Setup Smartlook for Docusaurus. Smartlook token is imported from Doppler.

Not sure abou lychee check though cc @TC-MO

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -13,3 +13,4 @@ jobs:
         - run: gh workflow run deploy.yaml --repo apify/apify-docs-private
           env:
             GITHUB_TOKEN: ${{ secrets.APIFY_SERVICE_ACCOUNT_GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


TC-MO:
We are aware of issue, part of python docs is not being built at the moment. I'm also not aware of any ETA for the fix for this. If this is urgent I think it should be okay to just force merge without lychee passing

B4nan:
looks like this broke the build (or better say PR checks)

see https://github.com/apify/apify-docs/actions/runs/8435147064/job/23102478546?pr=896

B4nan:
it breaks the local development as well, will add a dummy fallback to the env var, looks like that's enough at least locally

B4nan:
next time please add me as a reviewer to such changes, this was really broken. another problem - the plugin dependency is only added to the apify-docs, but it needs to be in all the repositories (sdks, clients, cli)

HonzaTuron:
Will do, sorry and thanks

PR #891, opened 2024-03-22
fix: examples cards small texts

webrdaniel:
Fixing small headings and descriptions on link cards (e.g. in `/sdk/js/docs/examples`).
This font size comes from docosaurus and the class names have generated suffixes (e.g. `cardTitle_rnsV`) so special selectors had to be used. It was probably caused when changing the rem base size from 16px to 10px.

Reported here: https://apify.slack.com/archives/C0L33UM7Z/p1710937646009279


jbartadev (approved):
Cool, cheers!

PR #897, opened 2024-03-26
fix(academy): major complete cleanup of typos and code bugs

metalwarrior665:
Sorry for this PR being so massive. You don't really need to read it, there might be a few changes that are not really consistent and I wasn't refactoring everything to the same language style but it is a huge improvement over the current state.

1. This started by following @gullmar 's [onboarding feedback doc](https://docs.google.com/document/d/1Kp9XzD64CzFuqlS6-bauDODDWa-B4ihqNZo-92L86Kg/edit#heading=h.not6o9vhyzo) (which is excellent!). I marked the fixed parts green and partially fixed have extra green.
2. While doing that, I was horrified that I still found a lot of typos. I thought we already proofreaded everything... So I started fixing them as well and then went into the rabbit hole and actually went through all files in the Academy. I'm using Grammarly extension in VS Code. Most of the fixes are clear typos or stylistic issues but there are also more language-style changes. I was not consistent in those and later stopped doing it even if Grammarly suggested. There are probably tons of wrongly placed commas but I gave up on those later as well. 

I would probably leave the stylistic part once we refactor the Academy structure. We should unify the voice & tone as well.


honzajavorek:
I'm not afraid to go through the changes, but the diff isn't readable much through the vale comments right now, so I'll wait for @TC-MO. TIL [vale](https://vale.sh/) exists! 👀 😮

TC-MO:
Vale is cool, but I think I need to work setup here, as it seems to be showing me max of ~8 errors, but when parsed through CLI not as GH Action it shows 140+ errors, will try to fix them ASAP

metalwarrior665:
I would do the Vale in a separate PR. My changes were minimally stylistical so the Vale issues are related to the previous content (and I totally agree that the voice & tone was not good, too hyped, so it needs fixing!). And the sooner we get all the typos and code bugs merged, the better.

TC-MO:
Welp I went ahead and fixed Vale so it should be a bit clearer now what are the issues and it will show only errors now (tho it is still necessary to go over logs of the GH Action if more than 10 errors are present in PR., but this is GitHub limitation as far as I'm aware)

and went ahead and fixed all vale issues that were present in this PR

TC-MO (approved):

PR #896, opened 2024-03-26
docs: fix markdown to disable an unnecessary link

honzajavorek:
I suppose this shouldn't be a real link.

PR #907, opened 2024-04-05
docs: disable unnecessary link

honzajavorek:
Supersedes https://github.com/apify/apify-docs/pull/896:

- better commit names
- made on top of current `master`, properly using `git` and not a GitHub pencil 😅 
- using inline code per @B4nan's comment https://github.com/apify/apify-docs/pull/896#pullrequestreview-1960109697

### DIFF
--- a/sources/academy/webscraping/web_scraping_for_beginners/crawling/finding_links.md
+++ b/sources/academy/webscraping/web_scraping_for_beginners/crawling/finding_links.md
@@ -59,7 +59,7 @@ We'll start from a boilerplate that's very similar to the scraper we built in [B
 
 Aside from importing libraries and downloading HTML, we load the HTML into Cheerio and then use it to retrieve all the `<a>` elements. After that, we iterate over the collected links and print their `href` attributes, which we access using the [`.attr()`](https://cheerio.js.org/docs/api/classes/Cheerio#attr) method.
 
-When you run the above code, you'll see quite a lot of links in the terminal. Some of them may look wrong, because they don't start with the regular **https:\/\/** protocol. We'll learn what to do with them in the following lessons.
+When you run the above code, you'll see quite a lot of links in the terminal. Some of them may look wrong, because they don't start with the regular `https://` protocol. We'll learn what to do with them in the following lessons.
 
 ## Next Up {#next}
 
--- a/sources/academy/webscraping/web_scraping_for_beginners/crawling/finding_links.md
+++ b/sources/academy/webscraping/web_scraping_for_beginners/crawling/finding_links.md
@@ -59,7 +59,7 @@ We'll start from a boilerplate that's very similar to the scraper we built in [B
 
 Aside from importing libraries and downloading HTML, we load the HTML into Cheerio and then use it to retrieve all the `<a>` elements. After that, we iterate over the collected links and print their `href` attributes, which we access using the [`.attr()`](https://cheerio.js.org/docs/api/classes/Cheerio#attr) method.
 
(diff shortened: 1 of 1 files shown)


honzajavorek:
Still failing with

```
[INFO] [en] Creating an optimized production build...
Error:  Unable to build website for locale en.
Error:  Error: You specified the `smartlook` object in `themeConfig` but the `projectKey` field was missing. Please ensure this is not a mistake.
```

B4nan:
Hmm I think I know whats happening, my fix only handles not set env var, but maybe we are getting empty string in there instead. 

This should help I think 24ce0fbc8ebba92d926c862256a47e08b0cd0f3a

B4nan (approved):
yeah good, that was it.

btw one unrelated note, it does not matter much in this repository, but generally speaking, commits that don't change in the library itself shouldn't be marked as fix/feat, those should be reserved for actual library fixes/features, things that belong to changelog (we generate changelogs based on the git history). 

as said, it does not matter much here, as this is not a library repository, but it does matter for all the other projects like SDKs or clients, where commits like this should be marked as `ci:` or `docs:` or `chore:`. since this changes docs content only, i would go with `docs:`

B4nan:
also fixed the broken langchain link here 53bc0c6782d7b8d8405ab1335f59226d9f6eddb8

B4nan:
hmm and here is the last broken link, not sure why lychee didn't print both :/

f1d0234e19cefa89e959962d9b3b738e670d20b8

B4nan:
ok, finally a green light :D next time it should be simpler, be sure to update your fork. also sent you an invite to the project contributors so you can just create a branch instead of using a fork.

honzajavorek:
Thanks for resolving the issue and for merging the PR! And for the invite.

> btw one unrelated note, it does not matter much in this repository, but generally speaking, commits that don't change the library itself shouldn't be marked as fix/feat, those should be reserved for actual library fixes/features, things that belong to changelog (we generate changelogs based on the git history).
> 
> as said, it does not matter much here, as this is not a library repository, but it does matter for all the other projects like SDKs or clients, where commits like this should be marked as `ci:` or `docs:` or `chore:`. since this changes docs content only, i would go with `docs:`

I use the prefixes in the context of a _product_ at hand. E.g., if adding commits to apify/apify-sdk-python, the main product is a Python library, so fix in code would be `fix`, and updates to documentation would be `docs`. In the context of apify/apify-docs I felt like the main product of the repo is the docs, the text files, and most of the commits will be referring to changes in the docs, so I've used `fix` etc. accordingly. In the context of documentation site, changes to code feel more like `chore` to me.

Looking at the commit history, it seems like @TC-MO sticks to mark everything as `docs`, the same way you suggest, but @metalwarrior665 seems to do the same mental shift as me, e.g. https://github.com/apify/apify-docs/commit/74bee0500e1f34358797be004b86832f4bf93d1d or https://github.com/apify/apify-docs/commit/3e21f78185dbca0fe984314e1e3ba4690709fe30 I don't mind which it is, just wanted to provide my reasoning on why I did it. I will follow whatever convention you sirs agree upon.

B4nan:
> I use the prefixes in the context of a product at hand. E.g., if adding commits to apify/apify-sdk-python, the main product is a Python library, so fix in code would be fix, and updates to documentation would be docs. In the context of apify/apify-docs I felt like the main product of the repo is the docs, the text files, and most of the commits will be referring to changes in the docs, so I've used fix etc. accordingly. In the context of documentation site, changes to code feel more like chore to me.

True, anyway, it's not very important since there is no generated changelog, which is the main reason to adhere to those rules, feel free to do it this way here.

PR #911, opened 2024-04-05
feat: improvements to the contributing guide

honzajavorek:
Reading through the contributing guide, I suggest the following improvements:

- Added top-level overview of the docs so that it's clear what "platform" or "academy" means when they're mentioned later
- Some paragraphs under felt like they belong under headings right after them
- Mention of redirects and SEO juice in the README felt oddly specific

I just fixed what I felt like fixing, didn't do a thorough proofreading or any effort to make this document beautiful, as I don't think that's the focus right now.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,27 +2,33 @@
 
 ## Architecture
 
-Apify documentation consists of six different repositories:
+Apify documentation consists of the following top-level areas:
 
+- **Academy** - Collection of mostly platform-independent courses on scraping techniques.
+- **Platform** - The main docs on how to use Apify as a product.
+- **API** - API reference and API client libraries docs, i.e. interacting with the platform from outside.
+- **SDK** - SDK libraries docs, i.e. interacting with the platform from code which runs inside it.
+- **CLI** - Tool for building code for the platform, and for interacting with the platform from outside.
+
+Thus the source code of the docs is spread across six different repositories:
+
+- apify-docs (this repository)
 - apify-client-js
 - apify-client-python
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -2,35 +2,41 @@
 
 ## Architecture
 
-Apify documentation consists of six different repositories:
+Apify documentation consists of the following top-level areas:
 
+- **Academy** - Collection of mostly platform-independent courses on scraping techniques.
```
honzajavorek, line 7: Aspirational description 😅

README.md
```diff
@@ -41,20 +41,19 @@ Instead, try to be more descriptive, for example:
...
-    const docsAreCool = require('coolDocs'); <br/>
-    ...<br/>
-    return docsAreCool;<br/>
+  const docsAreCool = require('coolDocs'); <br/>
+  ...<br/>
+  return docsAreCool;<br/>
 
-    \`\`\`
-    ```
+  \`\`\`
```
honzajavorek, line 53: This renders with the backslashes, but I didn't figure out how to do it in a correct way, so I left it as it was.

B4nan: https://stackoverflow.com/questions/42319711/fenced-code-block-inside-another-fenced-code-block

honzajavorek: I see I gave up too early! With all the AI around one would forget about the good old StackOverflow 😀 Thanks! Interesting solution, lol, adding more backticks. And it seems to work! At least in VS Code Markdown preview.

TC-MO (approved):

Issue #849, opened 2024-02-07
Billing->Limits (Platform docs) is missing info about overage usage and how it's charged


TsvetaYankova:
Hey Michal,
Can you please adjust the wording as follows:
_The Limits tab showcases the limits of Apify platform usage based on your current subscription plan. You will receive a notification if your usage is approaching the limit. If the limit is exceeded, Apify platform services will be suspended to stop incurring additional charges.
You can adjust the platform limit and go into overage(Pay as you go). All extra usage will be charged as overage to your account. You will be invoiced earlier if the overage platform usage reaches $200. Otherwise, this usage will be added to your next subscription invoice._

![image](https://github.com/apify/apify-docs/assets/111116306/5df2b01b-8f6e-4473-bff0-c2ff62805748)

TC-MO:
Fixed in https://github.com/apify/apify-docs/pull/920

PR #920, opened 2024-04-15
docs: add usage overage info

TC-MO:
add usage overage info to billings doc
fix broken table on limits.md
add link to Limits section of documentation in Limits doc


TC-MO:
Added info on overages based on the https://github.com/apify/apify-docs/issues/849

PR #927, opened 2024-04-18
feat: Improve docs around rate limiting, mention higher rate limit for key-value store record endpoints

fnesveda:
The documentation around rate limits was a bit confusing, so I expanded it a bit, to make the distinction between global and per-resource rate limits clearer.

I've also added a mention of the increased rate limits on key-value store records.

### DIFF
--- a/sources/platform/api_v2/api_v2_reference.apib
+++ b/sources/platform/api_v2/api_v2_reference.apib
@@ -329,22 +329,22 @@ There are two kinds of rate limits - a global rate limit and a per-resource rate
 
 ### Global rate limit
 
-The global rate limit is set to 250 000 requests per minute.
+The global rate limit is set to _250 000 requests per minute_.
 For [authenticated]((#authentication) requests, it is counted per user,
 and for unauthenticated requests, it is counted per IP address.
 
 ### Per-resource rate limit
-The default per-resource rate limit is 30 requests per second per resource, which in this context means a single Actor, a single Actor run, a single dataset, single key-value store etc.
+The default per-resource rate limit is _30 requests per second per resource_, which in this context means a single Actor, a single Actor run, a single dataset, single key-value store etc.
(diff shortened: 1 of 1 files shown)


drobnikj (approved):
👍

sources/platform/api_v2/api_v2_reference.apib
```diff
@@ -324,9 +324,27 @@ Here is the table of the most common errors that can occur for many API endpoint
...
 
 All API endpoints limit the rate of requests in order to prevent overloading of Apify servers by misbehaving clients.
-The default rate limit is 30 requests per second and is applied to every API endpoint except the following list
-of endpoints that are rate limited to 200 requests per second:
 
+There are two kinds of rate limits - a global rate limit and a per-resource rate limit.
+
+### Global rate limit
+
+The global rate limit is set to 250 000 requests per minute.
```
TC-MO, line 332:
```suggestion
The global rate limit is set to _250 000 requests per minute_.
```

sources/platform/api_v2/api_v2_reference.apib
```diff
@@ -324,9 +324,27 @@ Here is the table of the most common errors that can occur for many API endpoint
...
+For [authenticated]((#authentication) requests, it is counted per user,
+and for unauthenticated requests, it is counted per IP address.
+
+### Per-resource rate limit
+The default per-resource rate limit is 30 requests per second per resource, which in this context means a single Actor, a single Actor run, a single dataset, single key-value store etc.
```
TC-MO, line 337:
```suggestion
The default per-resource rate limit is _30 requests per second per resource_, which in this context means a single Actor, a single Actor run, a single dataset, single key-value store etc.
```

TC-MO (approved):
Only suggestion would be to use emphasis on number values like the comments I left, other than that LGTM

TC-MO:
sure thing, apologies, it has been a force of habit for me, will do it this way in the future

PR #932, opened 2024-04-21
docs: add legal section

mnmkng:
I'm adding a new top level section to Docs: `docs.apify.com/legal` This section will house all of our public facing terms, policies and agreements.

At this time, I don't need it to be linked from the Docs homepage or from the top navigation bars, since the individual documents will be linked directly from elsewhere. In the future, we should find a reasonable place for a link to the whole section.

The individual terms and agreements will follow.


mnmkng:
Lychee is failing because the documents don't exist yet, so they're throwing 404s.

TC-MO:
Yes, this is expected behavior, due to the way we had to set Lychee up

sources/legal/latest/policies/acceptable-use-policy.md
```diff
@@ -0,0 +1,49 @@
...
+description: Apify's acceptable use policy describes activities that are prohibited on the Apify platform and on our websites.
+sidebar_label: Acceptable Use Policy
+sidebar_position: 1
+category: legal
+slug: /acceptable-use-policy
+---
+
+<!-- vale off -->
+
+# Apify Acceptable Use Policy
```
TC-MO, line 12: If h1 is the same as `title` it doesn't make sense to have both, `title`. I would suggest removing H1 especially since it seems to be rendering twice at the moment 

obraz

sources/legal/latest/policies/acceptable-use-policy.md
```diff
@@ -0,0 +1,49 @@
...
+
+Apify Technologies s.r.o., with its registered seat at Vodičkova 704/36, Nové Město, 110 00 Prague 1, Czech Republic, Company ID No.: 04788290, registered in the Commercial Register kept by the Municipal Court of Prague, File No.: C 253224 (hereinafter referred to as “**we**” or “**Apify**”), is committed to making sure that the Platform and the Website are being used only for legitimate and legal purposes.
+
+By agreeing to our General Terms and Conditions or simply by using the Platform or the Website, you also agree to be bound by this Acceptable Use Policy.
+
+## A. General Provisions
```
TC-MO, line 18: small nitpick , is there any reason why there are three style of headings?

With lettered lists, numbered lists and without lettering/numbering?

TC-MO (approved):

PR #934, opened 2024-04-22
docs: update langchain integration docs as the Document has moved to langchain_core

jirispilka:
- update langchain integration as the Document has moved to langchain_core and ApifyWrapper to langchain_community.
- change langchain URL links

### DIFF
--- a/sources/platform/integrations/langchain.md
+++ b/sources/platform/integrations/langchain.md
@@ -20,7 +20,7 @@ but if you prefer to use JavaScript, you can follow the same steps in the [JavaS
 
 Before we start with the integration, we need to install all dependencies:
 
-`pip install apify-client langchain langchain_community tiktoken`
+`pip install apify-client langchain langchain_community openai tiktoken`
 
 After successful installation of all dependencies, we can start writing code.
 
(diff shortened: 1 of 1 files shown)


TC-MO (approved):

TC-MO:
@B4nan Are we sure that is the issue? If I remember correctly our implementation even on PR's crawls whole website, and it seems they did their redirect was within lychee default redirect value so it did not view it as broken link.

B4nan:
Well, not having it in the master workflow means we tend to ignore it (as even with a red PR check, the master ends up green). When you look at the lychee workflow, it was apparenlty failing for the last two weeks:

https://github.com/apify/apify-docs/actions/workflows/lychee.yml

So I'd guess the problematic link was broken during those two weeks, for whatever reason - maybe this one is broken because of some changes in redirects, haven't touched that part myself for a while, but I see there were some changes ~2 weeks ago:

https://github.com/apify/apify-docs/blob/master/nginx.conf

TC-MO:
ah yes, I'm aware of this link. It's a link that has been published to a new tutorial that is yet to be released. 
I should have acted earlier and comment it out, until the tutorial has been published. There would be no issue then.

PR #935, opened 2024-04-22
feat: Improve Apify Proxy Usage docs

### LINKED ISSUE #740: Improve explanation of internal access to proxy from Actors
Users have issues accessing proxy from within Actors when not using `Actor.createProxyConfiguration()`. If they're free users, they don't have external proxy access, and if they access proxy using the `proxy.apify.com` hostname, it fails even when the call is done from inside an Actor, because it's routed through the internet instead of the internal endpoint. We don't have any explanation anywhere that for access to proxy to be considered internal, you have to use the `APIFY_PROXY_HOSTNAME` env var instead of `"proxy.apify.com"`.

fnesveda:
Every now and then, we have a free user that can't use Apify Proxy from inside of the Actors because they're trying to connect to `proxy.apify.com`, not internally through `$APIFY_PROXY_HOSTNAME`, which fails since it's considered an external connection, which is banned on the Free plan.

This improves the Apify Proxy connection instructions to hopefully better explain that when connecting to Apify Proxy from inside an Actor, they should either use the helpers in the `Actor` class in the SDK, or use the `APIFY_PROXY_HOSTNAME`, `APIFY_PROXY_PORT` and `APIFY_PROXY_PASSWORD` environment variables.

Closes #740

### DIFF
--- a/sources/platform/proxy/usage.md
+++ b/sources/platform/proxy/usage.md
@@ -21,7 +21,7 @@ The full connection string has the following format:
 http://<username>:<password>@<hostname>:<port>
 ```
 
-> **WARNING:** All usage of Apify Proxy with your password is charged towards your account. Do not share the password with untrusted parties or use it from insecure networks – **the password is sent unencrypted** due to the HTTP protocol's [limitations](https://www.guru99.com/difference-http-vs-https.html).
+> **WARNING:** All usage of Apify Proxy with your password is charged towards your account. Do not share the password with untrusted parties or use it from insecure networks—**the password is sent unencrypted** due to the HTTP protocol's [limitations](https://www.guru99.com/difference-http-vs-https.html).
 
 ### External connection
 
@@ -136,8 +136,8 @@ Web scrapers can rotate the IP addresses they use to access websites. They assig
 
 Depending on whether you use a [browser](https://apify.com/apify/web-scraper) or [HTTP requests](https://apify.com/apify/cheerio-scraper) for your scraping jobs, IP address rotation works differently.
 
-* Browser – a different IP address is used for each browser.
-* HTTP request – a different IP address is used for each request.
+* Browser—a different IP address is used for each browser.
+* HTTP request—a different IP address is used for each request.
 
 Use [sessions](#sessions) to controll how you rotate and [persist](#session-persistence) IP addresses. See our guide [Anti-scraping techniques](/academy/anti-scraping/techniques) to learn more about IP address rotation and our findings on how blocking works.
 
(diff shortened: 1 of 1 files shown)


jirimoravcik (approved):

sources/platform/proxy/usage.md
```diff
@@ -21,7 +21,9 @@ The full connection string has the following format:
 http://<username>:<password>@<hostname>:<port>
 ```
 
-> **WARNING:** All usage of Apify Proxy with your password is charged towards your account. Do not share the password with untrusted parties or use it from insecure networks—**the password is sent unencrypted** due to the HTTP protocol's [limitations](https://www.guru99.com/difference-http-vs-https.html).
+:::caution
+All usage of Apify Proxy with your password is charged towards your account. Do not share the password with untrusted parties or use it from insecure networks—**the password is sent unencrypted** due to the HTTP protocol's [limitations](https://www.guru99.com/difference-http-vs-https.html).
```
TC-MO, line 25:
```suggestion
All usage of Apify Proxy with your password is charged towards your account. Do not share the password with untrusted parties or use it from insecure networks, as **the password is sent unencrypted** due to the HTTP protocol's [limitations](https://www.guru99.com/difference-http-vs-https.html).
```

sources/platform/proxy/usage.md
```diff
@@ -134,12 +171,12 @@ To test that your requests are proxied and IP addresses are being [rotated](/aca
...
 ### A different approach to `502 Bad Gateway`
 
-There are times when the `502` status code is not comprehensive enough. Therefore, we have modified our server with `590-599` codes instead to provide more insight:
+Sometimes when the `502` status code is not comprehensive enough. Therefore, we have modified our server with `590-599` codes instead to provide more insight:
 
 * `590 Non Successful`: upstream responded with non-200 status code.
 * `591 RESERVED`: *this status code is reserved for further use.*
 * `592 Status Code Out Of Range`: upstream responded with status code different than 100-999.
-* `593 Not Found`: DNS lookup failed - [`EAI_NODATA`](https://github.com/libuv/libuv/blob/cdbba74d7a756587a696fb3545051f9a525b85ac/include/uv.h#L82) or [`EAI_NONAME`](https://github.com/libuv/libuv/blob/cdbba74d7a756587a696fb3545051f9a525b85ac/include/uv.h#L83).
+* `593 Not Found`: DNS lookup failed—[`EAI_NODATA`](https://github.com/libuv/libuv/blob/cdbba74d7a756587a696fb3545051f9a525b85ac/include/uv.h#L82) or [`EAI_NONAME`](https://github.com/libuv/libuv/blob/cdbba74d7a756587a696fb3545051f9a525b85ac/include/uv.h#L83).
```
TC-MO, line 179:
```suggestion
* `593 Not Found`: DNS lookup failed, indicating either [`EAI_NODATA`](https://github.com/libuv/libuv/blob/cdbba74d7a756587a696fb3545051f9a525b85ac/include/uv.h#L82) or [`EAI_NONAME`](https://github.com/libuv/libuv/blob/cdbba74d7a756587a696fb3545051f9a525b85ac/include/uv.h#L83).
```

sources/platform/proxy/usage.md
```diff
@@ -134,12 +171,12 @@ To test that your requests are proxied and IP addresses are being [rotated](/aca
 
 ### A different approach to `502 Bad Gateway`
 
-There are times when the `502` status code is not comprehensive enough. Therefore, we have modified our server with `590-599` codes instead to provide more insight:
+Sometimes when the `502` status code is not comprehensive enough. Therefore, we have modified our server with `590-599` codes instead to provide more insight:
 
 * `590 Non Successful`: upstream responded with non-200 status code.
 * `591 RESERVED`: *this status code is reserved for further use.*
 * `592 Status Code Out Of Range`: upstream responded with status code different than 100-999.
```
TC-MO, line 178:
```suggestion
* `592 Status Code Out Of Range`: upstream responded with status code different than 100–999.
```

TC-MO (changes requested):

TC-MO (approved):

Issue #909, opened 2024-04-05
Unable to install the project as an external contributor


honzajavorek:
The repository uses private npm packages:

```
npm ERR! code E401
npm ERR! 401 Unauthorized - GET https://npm.pkg.github.com/download/@apify-packages/ui-components/0.86.0/6927975c9bcef7f899f41292e4b3d2b675011690 - authentication token not provided
```

I'm filing this as an issue, because I think it would be vital of anyone could clone this repo and contribute to it. Even if only some components of the docs could run locally, in a similar sense like this:

https://github.com/apify/apify-docs/blob/82d6da569455bdc5a8e8dece260431959e04f5ef/CONTRIBUTING.md?plain=1#L83

Currently, it is not possible to even successfully finish `npm install` of the repo.

B4nan:
cc @jbartadev I lost track of this, do you still plan to make the package public or inline the necessary components here?

B4nan:
@honzajavorek here are docs on how to install the package, but can't add you to the project myself

https://www.notion.so/apify/How-to-use-apify-packages-npm-registry-dfc20d3691ff4a879ef30ad2d104db75

honzajavorek:
Thanks, but that just points to 404 😅

B4nan:
Yeah, it's a completely valid point, we've been struggling with this internally too. This started relatively recently when a new homepage was added which uses the shared (and private) packages. I thought the packages were already made public, but as said above, I kinda lock track of the current state and plans.

jbartadev:
The ui-components library is published as a public package, it should be possible to install it without any authorization

https://github.com/apify-packages/shared-components/pkgs/npm/ui-components

B4nan:
Maybe we just need to update its version, but I remember there were breaking changes. Let me have a look.

B4nan:
Build is passing with latest version, will update in a minute.

B4nan:
Btw @jbartadev you might want to ship v1 of those public packages, as right now, every `0.x` release is considered major/breaking, so renovate won't update it automatically for us. If you would ship v1 and start distinguishing feature releases from major/breaking ones, it would help downstream to stay up to date.

honzajavorek:
> The ui-components library is published as a public package, it should be possible to install it without any authorization
> 
> [https://github.com/apify-packages/shared-components/pkgs/npm/ui-components](https://github.com/apify-packages/shared-components/pkgs/npm/ui-components)
> 
[...]

Issue #936, opened 2024-04-22
Is Nginx still needed for local setup?


honzajavorek:
The contributing guide contains the following:

> All the links in the navbar and footer need to be absolute, and they will use a different hostname, configured to docs.apify.loc - to use that, follow the steps below and set up the Nginx server.
> 
> Alternatively, you can skip the Nginx part and navigate to http://localhost:3000/academy or http://localhost:3000/platform manually instead of using links in the navbar.

This seems to be somewhat outdated. When I `npm start` and open http://localhost:3000/ in my browser, the top navigation seems to be working out of the box for both the Academy and Platform.

Links to docs sourced from other repos give me 404s, so I presume if I needed to develop those, I'd still need the Nginx setup. But no manual navigation seems to be needed for the core parts of the docs.

B4nan:
You need nginx if you want to set the whole thing up - if you want to have the complete portal including the child projects like SDKs, as each of those projects will run on a separate port and you need a webserver (nginx) on top of them to have everything on a single domain/port. You won't need this if you only want to deal with the one repository (e.g. this one if you want to update platform/academy docs, or the JS SDK if you want to update some SDK guides).

honzajavorek:
So the sentences could be changed to something along the lines of:

> The parts of the docs sourced from other repositories will give you 404s. If you want to locally run the project with all the other repositories included, clone them all and setup an Nginx server according to the steps below.

I removed mentions of manual navigation, as I don't think it is needed. I have the project running locally and I don't need it. I also removed the initial mention of docs.apify.loc, because I don't see any absolute links to that domain. Does such change of the wording make sense? If so, I can add it to https://github.com/apify/apify-docs/pull/937

B4nan:
> I also removed the initial mention of docs.apify.loc, because I don't see any absolute links to that domain

That's because you don't run it with the `DEV` env var (so [via `npm run start:dev`](https://github.com/apify/apify-docs/blob/master/package.json#L20)). This is still valid for the nginx setup, it will be the domain you will use to access the whole portal (and you need to adjust the `/etc/hosts` file too). 

https://github.com/apify/apify-docs/blob/master/apify-docs-theme/src/config.js#L6

honzajavorek:
Ah, nice. That isn't documented anywhere I believe 😀 I can mention these in the new wording.

B4nan:
[...]

PR #918, opened 2024-04-10
feat: Added tutorial to scraper website from multiple Actor runs

drobnikj:
* Added tutorial about how to scrape a single website from multiple Actor runs
* Added example of code under `./examples`

I do it as short as possible on one side, plus cover all necessary problems.
Of course, it is not a production-ready solution, but it should give the reader an idea of how to create a scraper running from multiple Actor runs.

PR #937, opened 2024-04-22
feat: document how to install the docs

honzajavorek:
Documents @mnmkng's approach as mentioned in https://github.com/apify/apify-docs/issues/909#issuecomment-2068960909 and closes https://github.com/apify/apify-docs/issues/909. If https://github.com/apify/apify-docs/issues/936 gets some responses, I can add commits rewording some of the sentences too, but as of now I leave that part as it is.


honzajavorek:
The `request_queue.md` contains a link to a non-existing part of the Academy 🤔 

> A detailed tutorial on how to crawl one request queue from multiple Actor runs can be found in [Advanced web scraping academy](https://docs.apify.com/academy/advanced-web-scraping/multiple-runs-scrape).

Looks like the page links https://github.com/apify/apify-docs/pull/918, which isn't merged yet? And it fails under my Pull Request…

CONTRIBUTING.md
```diff
@@ -79,13 +79,35 @@ For example:
 
 Over time, we should track which items are useful and which don't get any traffic. Also, as Apify Docs expand, we may need to add more cards and update which articles we link to.
 
-## Local setup
+## Installation and setup
```
honzajavorek, line 82: @B4nan I'd appreciate if you could take a look whether I got these sections right.

B4nan: sure thing

TC-MO (approved):

TC-MO:
Yeah, you can disregard the failing lychee check, as it is pertaining only to the soon released tutorial.

CONTRIBUTING.md
```diff
@@ -79,13 +79,35 @@ For example:
...
+   npm login --scope=@apify-packages --registry=https://npm.pkg.github.com --auth-type=legacy
+   ```
 
-Clone all the repositories, and start the Docusaurus instances in them.
+   Read [#909](https://github.com/apify/apify-docs/issues/909) if you want to understand why this is needed.
+3. Run `npm install`.
+4. Run `npm start`. The website should open at <http://localhost:3000>.
+
+This is sufficient to work on the main content, i.e. the Academy and Platform docs.
+If you want to also work on the other parts of the docs, you may want to first join them all together using Nginx.
```
B4nan, line 100: nit: we dont use prose wrap in markdown, its 2024 and editors can wrap things for us properly if we want them to do so (and its irrelevant for the rendered version). this is actually making things worse when reviewing the git diff (in general, its not relevant to this particular place).

at some point i'd like to add some code formatter that will deal with this automatically, we actually have/had a prettier setup here but it was more of a one time thing during the big migration. maybe we could prioritize that since there are now two people working primarily on the docs, it should be simple to revive. we are experimenting with biome which is super fast so it would be even suitable to have that enable on save

CONTRIBUTING.md
```diff
@@ -125,12 +145,14 @@ server {
 }
 ```
 
-And add a record to `/etc/hosts` to map the `docs.apify.loc` hostname to localhost:
+Add a record to `/etc/hosts`, which maps the `docs.apify.loc` host to a localhost:
```
B4nan, line 148: nit: `hostname` was probably more correct, from what I read, `host` is also supposed to contain the port, while `hostname` is just the name. but maybe not, haven't been googling that hard (and was happy with the first result that confirmed what I thought)

B4nan (approved):
just a few notes, nothing blocking

CONTRIBUTING.md
honzajavorek: I do it for [better writing](https://sive.rs/1s), not so much for better git diffs. I'll do whatever you folks prefer, because as you say, it's two different things how I type characters to the file vs how they're committed to git.

Two people is a low number, I think we can agree on a convention just fine without tools policing us. But from what I've seen just for the past week, there's many people contributing to docs. If I'm the only one doing this tho, I still think it's easier to just tell me not to do it 😀

If you ever set up automation, I'll look over your shoulder 😀 I have Prettier on junior.guru, but I had to turn off formatting for many Markdown documents, because I have them mixed with Jinja2 etc., and that isn't something Prettier would understand.

honzajavorek: Fixed in [35f8595](https://github.com/apify/apify-docs/pull/937/commits/35f8595dc454d95b5f5b06bc2af4eea894775c9a)

CONTRIBUTING.md
```diff
@@ -79,13 +79,35 @@ For example:
 
 Over time, we should track which items are useful and which don't get any traffic. Also, as Apify Docs expand, we may need to add more cards and update which articles we link to.
 
-## Local setup
+## Installation and setup
 
-If you want to work only on the main documentation content, cloning this repository is enough. Once you run `npm install` to install all dependencies and run `npm start`, the main portal will open on <http://localhost:3000>. All the links in the navbar and footer need to be absolute, and they will use a different hostname, configured to `docs.apify.loc` - to use that, follow the steps below and set up the Nginx server.
+1. Use Git to clone this repository.
+2. If you work at Apify and can install packages prefixed with `@apify-packages`, you can skip this step.
```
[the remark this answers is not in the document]
honzajavorek: Fixed in [1c086c6](https://github.com/apify/apify-docs/pull/937/commits/1c086c636cc34f23b2cf1b012b23b0752667967f)

honzajavorek:
Even though they were not blocking, I addressed the comments. I'll merge this now, but if you find anything to correct, feel free to comment further - I can send fixes in a subsequent PR.

PR #939, opened 2024-04-24
feat: moved imports of all shared components to ui-library

jbartadev:
A part of our effort to move reusable logic to the public components library and use the ui-components packages as a private library for sharing code between Apify apps.


barjin (approved):
Lgtm, thanks!

Out of curiosity, does this mean we won't need to do the github keys in `.npmrc` anymore? Is the `ui-library` gonna be publicly on npm?

jbartadev:
> Lgtm, thanks!
> 
> Out of curiosity, does this mean we won't need to do the github keys in `.npmrc` anymore? Is the `ui-library` gonna be publicly on npm?

@barjin That's the plan. We can't do it yet but it's coming.

HonzaKirchner (approved):

PR #999, opened 2024-05-09
feat: update schedules docs to include notifications

m-murasovs:
This is related to https://github.com/apify/apify-core/pull/15693, which adds notifications to failed schedules. 

closes apify/apify-core/#15593


sources/platform/schedules.md
```diff
@@ -46,13 +46,13 @@ Your schedule's name should be 3–63 characters long.
 
 In [Apify Console](https://console.apify.com/schedules), click on the **Schedules** in the navigation menu, then click the **Create new** button.
 
-Click on the name (by default it is _My Schedule_), there you can change the name of Schedule, add a description, as well as check it's _Unique name_ or _ID_.
+Click on the name (by default it is _My Schedule_), there you can change its name, add a description, as well as check its _Unique name_ or _ID_.
```
TC-MO, line 49: Since these are all part of UI I would go with **My Schedule** instead of emphasis

TC-MO (approved):
LGTM , just one small note, otherwise approved

PR #1006, opened 2024-05-16
fix(docs): codeblock correct theme

baldasseva:
In page /sdk and /api expecially, now the color of the Codeblock texts is working with the color theme of the page


HonzaTuron (approved):
Good job 😀

webrdaniel (approved):

PR #1011, opened 2024-05-20
fix: improve stylistics and several explanations

honzajavorek:
The individual commit messages should be self-explanatory.


mnmkng (approved):
Looks good to me. IMO I don't need to be included in stylistic reviews and small updates. I trust @TC-MO more than myself on this.

PR #1017, opened 2024-05-21
fix: rename several chapters to improve orientation and fix a few small issues

honzajavorek:
Individual commit messages should be self-explanatory. Each solves a problem which is documented as an issue on this repo. These issues are always linked from the commit description.

### DIFF
--- a/sources/academy/webscraping/web_scraping_for_beginners/data_extraction/browser_devtools.md
+++ b/sources/academy/webscraping/web_scraping_for_beginners/data_extraction/browser_devtools.md
@@ -5,8 +5,6 @@ sidebar_position: 1
 slug: /web-scraping-for-beginners/data-extraction/browser-devtools
 ---
 
-# Starting with browser DevTools {#devtools}
-
 **Learn about browser DevTools, a valuable tool in the world of web scraping, and how you can use them to extract data from a website.**
 
 ---
--- a/sources/academy/webscraping/web_scraping_for_beginners/data_extraction/browser_devtools.md
+++ b/sources/academy/webscraping/web_scraping_for_beginners/data_extraction/browser_devtools.md
@@ -1,11 +1,11 @@
 ---
-title: Browser DevTools - I
+title: Starting with browser DevTools
 description: Learn about browser DevTools, a valuable tool in the world of web scraping, and how you can use them to extract data from a website.
 sidebar_position: 1
 slug: /web-scraping-for-beginners/data-extraction/browser-devtools
 ---
 
-# Browser DevTools {#devtools}
+# Starting with browser DevTools {#devtools}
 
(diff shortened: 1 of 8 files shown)


sources/academy/webscraping/web_scraping_for_beginners/data_extraction/browser_devtools.md
```diff
@@ -1,11 +1,11 @@
 ---
-title: Browser DevTools - I
+title: Starting with browser DevTools
 description: Learn about browser DevTools, a valuable tool in the world of web scraping, and how you can use them to extract data from a website.
 sidebar_position: 1
 slug: /web-scraping-for-beginners/data-extraction/browser-devtools
 ---
 
-# Browser DevTools {#devtools}
+# Starting with browser DevTools {#devtools}
```
TC-MO, line 8: If title & H1 are to be the same you can omit H1, Docusaurus will then use `title` as H1.

Also you can control anchor links to headings by using `{anchor link here}`, in this case `{devtools}`

(if you want them to be different than headings)

TC-MO (approved):
LGTM, small comment/explanation only

sources/academy/webscraping/web_scraping_for_beginners/data_extraction/browser_devtools.md
honzajavorek: I wanted to keep anchor as it was, not to break any links. So if I understand correctly, I can achieve the same with the following in the frontmatter?

```yaml
title: Starting with browser DevTools {devtools}
```

TC-MO: Oh yeah, this escaped my mind I don;t think this would work if we want to control all anchors then I this H1 is necessary

honzajavorek: The docs mention the `{#id}` syntax: https://docusaurus.io/docs/markdown-features/toc But testing locally, it looks like it doesn't work for H1 at all. Even if I override it, when I inspect the resulting code, there is no such ID and the anchor doesn't work. The code below

```markdown
# Starting with browser DevTools {#devtools}
```

renders simply as

```html
<h1>Starting with browser DevTools</h1>
```

And when I think about it, it doesn't make much sense to add anchor overrides for H1 anyway. Linking specifically to H1 is effectively the same as linking to the whole page. And if it never worked, I think we're okay to just remove all anchor overrides for H1s!

TC-MO: It will render as H1, but the anchor link will be what you have inside the {}

PR #1020, opened 2024-05-21
fix: three small corrections in markup and code

honzajavorek:


sources/academy/webscraping/puppeteer_playwright/executing_scripts/injecting_code.md
```diff
@@ -119,7 +119,7 @@ const msg = await page.evaluate(() => returnMessage());
 
 console.log(msg);
 
-await page.waitForTimeout(20000);
+await page.waitForTimeout(10000);
```
honzajavorek, line 122: I'll ping @metalwarrior665 here because I just want to be sure I'm correcting a typo here. This code example has different timeout for Puppeteer and Playwright. I don't thing there is a particular reason to it, so I assume it's just a typo. But maybe you'll remember if it is or is not intentional.

metalwarrior665: Definitely typo

honzajavorek:
@metalwarrior665 @TC-MO Is this okay to merge?

TC-MO (approved):
LGTM

PR #1024, opened 2024-05-21
docs: update output schema document

TC-MO:
completely rework the structure of the document
reformat markdown tables
simplify examples within code sample
simplify ToC
style changes
make document style guide complaint
disable markdownlint rule md001 for last section of the document


mtrunkat (commented):
Just sending quickly some structure feedback:
- I really like that it now contains examples and screenshot what it actually does
- IMHO interlinking these would be benefitial
CleanShot 2024-05-23 at 14 56 50
- these should be H3 and thanks to that they will appear in the TOC on the right

CleanShot 2024-05-23 at 14 57 52
as in input schema
CleanShot 2024-05-23 at 14 58 36
- It might be beneficial to separate it into 2 pages:
    - Output schema (explanation with screenshots)
    - Specification
  as in input schema. The beneficial is that we can have then multiple specification versions and also that specs is not surrounded by extra content.
CleanShot 2024-05-23 at 14 59 08
 - I'd consider adding small snippets into "Dataset schema structure definitions" to make it easier to visualize it

TC-MO:
>     * these should be H3 and thanks to that they will appear in the TOC on the right
> 
> 
> CleanShot 2024-05-23 at 14 57 52

Regarding this point, this was done on purpose,  I thought that it muddles the TOC a bit, and this way the user can navigate the TOC to get to overall part of document that deals with dataset specification and from there, find the one that he needs to check

gippy (approved):
Looks ok to me, much easier to read.

TC-MO:
@davidjohnbarton this is now post reviews and will be merged today , tomorrow at the latest

PR #1027, opened 2024-05-22
style: use consistent code example language markers

honzajavorek:
Some examples were marked as `js`, some as `javascript`, some as `ts`, some as `typescript`. I chose the verbose variant, but in fact have no strong opinion, just wanted it to be consistent.


B4nan:
We use the short variants everywhere (at least I do), imo the long ones are mostly from the past.

TC-MO:
I have no opinion as long as syntax highlighting works correctly, but consistency is always welcome.

honzajavorek:
Eslint punishing me for touching old stuff 😅

TC-MO:
apparently if we want to go for the shortest one we could change all `python` to `py`, but not sure if there is a point to it.

honzajavorek:
@TC-MO I'm okay with it! I added [a663e40](https://github.com/apify/apify-docs/pull/1027/commits/a663e4064dd4b141793a0a2174aad829d0d814fe) 🐍 

Apart from one randomly timeouting link and ESlint I think all should be fine. I don't feel like I should fix lint errors in code as part of this PR just because I've touched the files, I'd like to keep the PR focused only on changing the code block markers.

Is it okay to merge like this?

TC-MO:
From lychee standpoint I'm ok, this link has been timing out multiple times on multiple pages. Not sure about ESlint maybe @B4nan can weigh in?

TC-MO (approved):
LGTM

honzajavorek:
I'll do it. Found `npm run lint:code:fix`, that already did half of the job.

honzajavorek:
Okay, so I did my best, but I couldn't fix it all. I'm not proficient in TypeScript, React, or the JavaScript Apify SDK to fix stuff like "Unexpected any. Specify a different type" or "React Hook "Actor.useState" cannot be called at the top level. React Hooks must be called in a React function component or a custom React Hook function".

I'll let the build to finish, see if only these errors are left, and then I'll merge this PR.

honzajavorek:
All green except those three linter errors. Merging!

honzajavorek:
Apparently it fails on `master` too 🙀 I thought it only lints changes! Filed https://github.com/apify/apify-docs/issues/1068

PR #1028, opened 2024-05-22
style: let the reader decide if something is simple or easy

honzajavorek:
We should refrain from using words like "easy!" or "simply" or "just" or "extremely easy" or "to show you how easy it really is", etc. Let the reader assess how hard a particular step is for them and not stress them out that it _should have been easy_.

This is something which bothered me quite a lot when reading through the academy for the first time, hence I decided that toning the whole academy down in this regard will be one of my first proposed changes.

This PR is rather large and I'm sorry for that. Most of the changes just remove one word from a sentence, as the sentence works completely the same without it. In some cases it was necessary to rewrite the sentence.


honzajavorek:
Getting huuuuge merge conflict with https://github.com/apify/apify-docs/pull/1067. Will re-create this PR after rebase.

PR #1029, opened 2024-05-22
docs: correct LlamaIndex examples

jirispilka:
Correct Python code examples as the LlamaIndex has changed package structure.
Add one more example of how to load the Apify dataset.

### DIFF
--- a/sources/platform/integrations/llama.md
+++ b/sources/platform/integrations/llama.md
@@ -9,7 +9,7 @@ slug: /integrations/llama
 
 ---
 
-> For more information on LlamaIndex, visit its [documentation](https://gpt-index.readthedocs.io/en/stable/).
+> For more information on LlamaIndex, visit its [documentation](https://docs.llamaindex.ai/en/stable/).
 
 ## What is LlamaIndex?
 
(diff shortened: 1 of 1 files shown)


sources/platform/integrations/llama.md
```diff
@@ -26,36 +26,58 @@ This example focuses on integrating Apify with LlamaIndex using the Apify Actor.
 
 Before we start with the integration, we need to install all dependencies:
 
-`pip install apify-client llama-index`
+`pip install llama-index-readers-apify`
```
drobnikj, line 29: You need to install `llama-index-core` as well.

sources/platform/integrations/llama.md
```diff
@@ -26,36 +26,58 @@ This example focuses on integrating Apify with LlamaIndex using the Apify Actor.
...
 
 After successfully installing all dependencies, we can start writing Python code.
-To use the Apify Actor loader, import `download_loader`, `Document`.
-Then, we need to import the loader using the method `download_loader`.
+
+### Apify Actor
+
+To use the Apify Actor, import `ApifyActor` and `Document`, and set your Apify API token in the code.
```
drobnikj, line 35: You can link how to find API token https://docs.apify.com/platform/integrations/api#api-token

drobnikj (approved):
Looks good.
Just two improvement otherwise fine, I leave checking grammar and texts to @TC-MO .

sources/platform/integrations/llama.md
jirispilka: It is actually not required as `llama-index-core` is a dependency of `llama-index-readers-apify`. However, on second thought, it might be better to install it explicitly along with `apify-client`.

TC-MO (approved):

PR #1036, opened 2024-05-30
docs: Added redirects for removed api clients pages

jbartadev:
Pages were removed in https://github.com/apify/apify-client-python/pull/232 and https://github.com/apify/apify-client-js/pull/542


B4nan (approved):
hmm, i will handle the openapi errors later, apparently my ignore rule wasn't enough

HonzaTuron (approved):

PR #1037, opened 2024-05-31
fix: split the `@font-face` rule to make it valid

barjin:
According to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-family#formal_syntax), the `@font-face { font-family }` descriptor can only be a single `string | custom ident`.

Compare to [`font-family` CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#formal_syntax), which can be a priority list.

This discrepancy is causing the tagline in the subprojects to look like ass (fig. 1)

obrazek

### DIFF
--- a/apify-docs-theme/src/theme/custom.css
+++ b/apify-docs-theme/src/theme/custom.css
@@ -340,11 +340,86 @@ body {
 }
 
 @font-face {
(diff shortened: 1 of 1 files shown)


B4nan (approved):

barjin:
obrazek

obrazek

PR #1059, opened 2024-06-10
docs: Fix broken link to flowise-apify-api.png for flowise integration doc

jirispilka:


TC-MO (approved):
LGTM, thanks for the catch!

PR #1067, opened 2024-06-17
fix(academy): capitalize 'actors' and 'actor'

honzajavorek:
- Using trial and error, with the help of VS Code's search and replace, I invented robust regular expressions to match lowercase actors and actor in the Markdown source code.
- I made all replacements manually by clicking the replace button in VS Code, one by one, always visually checking if it's inside prose or not.
- If there was a wrong match, I edited the regular expression to avoid such cases and resumed replacing.
- After I replaced all finds, I've performed another search, this time with a match-all regular expression `\bactors\b` or `\bactor\b`, to check whether there are no leftovers.
- In both `\bactors\b` or `\bactor\b` cases I found lowercase instances only inside URLs, identifiers, etc., so it all looks okay!
- I've saved the regular expressions that I've invented as Vale rules. For better visibility, this is committed separately as [a8f3a27](https://github.com/apify/apify-docs/pull/1067/commits/a8f3a279e7b42ed4c3c9ccc0a63605c9b2b69b5d). I did my best to explain the regular expressions in YAML comments.
- I made a few checks whether the rules work. I opened a random file and wrote a sentence with the words not capitalized, in various forms, and it seems that the Vale rules work correctly, I've got red underlines in the right cases.
- The changes in [9eecadf](https://github.com/apify/apify-docs/pull/1067/commits/9eecadf480be81e4c094c7afb7bf046b37118fe9) are unrelated, but I had to add it to make Vale happy as I touched several files with outstanding Vale errors.


honzajavorek:
Another random timeout 🙄 Lychee punishing me with red noise for no reason 🔴

TC-MO (approved):
LGTM! and thanks for updating Vale!

PR #1073, opened 2024-06-17
fix: fine-tune Vale and fix all outstanding errors

honzajavorek:
I removed `TokenIgnores`, as it clashes with rules being applied to link texts.

Screenshot 2024-06-17 at 13 28 59

Let's observe which annoying situations we have without it. For those we want to solve, let's create a separate GitHub Issue and I'll do my best with my regexp-fu to fine-tune Vale so that it's less annoying.

When I was at it, I accidentally fixed all outstanding errors. Just errors, not warnings 😅

### DIFF
--- a/sources/platform/storage/usage.md
+++ b/sources/platform/storage/usage.md
@@ -31,7 +31,7 @@ The [key-value store](./key_value_store.md) is ideal for saving data records suc
 
 ## Basic usage {#basic-usage}
 
-There are several ways to access your storage:
+You can access your storage in several ways:
 
 * [Apify Console](https://console.apify.com/storage) - provides an easy-to-use interface.
 * [Apify API](/api/v2#/reference/key-value-stores) - to access your storages programmatically.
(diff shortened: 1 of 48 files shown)


sources/platform/storage/usage.md
```diff
@@ -52,7 +52,9 @@ Additionally, you can quickly share the contents and details of your storage by
 
 ![Storage API](./images/overview-api.png)
 
+<!-- vale Microsoft.Dashes = NO -->
 These URLs link to API _endpoints_—the places where your data is stored. Endpoints that allow you to _read_ stored information do not require an [authentication token](/api/v2#/introduction/authentication). Calls are authenticated using a hard-to-guess ID, allowing for secure sharing. However, operations such as _update_ or _delete_ require the authentication token.
```
honzajavorek, line 56: There is a correct m-dash, but `Microsoft.Dashes` apparently isn't able to overcome the fact that the left side is marked up as italics.

TC-MO (approved):

PR #1074, opened 2024-06-17
docs: add new admonition to index.md within integrations/actors

TC-MO:
added new admonition with link to Store presenting Apify's catalogue of Integration Actors


jancurn (approved):
thanks

PR #1079, opened 2024-06-18
feat: add API documentation for dataset validation

gippy:
Hey all, this is a proposal of how the dataset validation could be handled in API.

We have decided, that no matter how many items in a request you send, if at least one is not valid the whole request fails with `400 Bad request`.

My proposal is to return our normal API error, but add "data" property to the error object returned from API and in the data return "invalidItems" which will list all validation errors for all invalid items by index in the items array.

So for example if you send a request with 1000 items and 2 of them do not have a required field. The response will look like this:

```json
{
  "error": {
    "type": "schema-validation-error",
    "message": "Schema validation failed",
    "data": {
        "invalidItems": {
          "4": ["name is required"],
          "124": ["name is required"],
         },
     }
  }
}
```


fnesveda (approved):
Nice, I really like that the whole request fails, it's easier to work with.

sources/platform/api_v2/api_v2_reference.apib
```diff
@@ -3081,6 +3081,9 @@ The POST payload is a JSON object or a JSON array of objects to save into the da
 **IMPORTANT:** The limit of request payload size for the dataset is 5 MB. If the array exceeds the size,
 you'll need to split it into a number of smaller arrays.
 
+If the dataset has fields schema defined, the push request can potentialy fail with `400 Bad Request` if any item does not match the schema.
```
tobice, line 3084: It'd be worth mentioning what happens with the valid items. I guess none will be inserted, right?

sources/platform/api_v2/api_v2_reference.apib
tobice, line 1: Btw what about our OpenAPI documentation? :sweat_smile: Shouldn't we be adding this there as well?

tobice (approved):
Two small points but approving :+1:

sources/platform/api_v2/api_v2_reference.apib
gippy: That is generated from this I think

B4nan: Yes, that was a one-off thing, I'd say that every change to the old specs also needs to be added to the openapi specs here: https://github.com/apify/openapi

We did quite a lot of polishment on what was converted from the old specs already (and more are coming next week), I don't think we want to do it all over again.

gippy: Thanks for the link. I will update it there too.

netmilk:
@tobice, thank you for the carbon copy. @gippy, I'm super sorry, I know I might be late to the party, but I have an issue with the structure of the validation payload.

For future extensibility, would it be possible to return the validation output (`invalidaItems` property) as an array of objects? I foresee some additional metadata that might be useful in the future for each validation output, such as `severity`, `jsonPointer`, `sourceMap` etc...

Can you still fix that @gippy ? I'd love to be sure we don't shoot ourselves in the foot.

Example:
```
...
invalidItems: {
  0: [
    {"text": "should have required property 'name'",
    {"text": "'bar' should be string"}
  ],
  1: [
    {"text":"'bar' should be string"}
  ],
},
...
```

gippy:
> @tobice, thank you for the carbon copy. @gippy, I'm super sorry, I know I might be late to the party, but I have an issue with the structure of the validation payload.
> 
> For future extensibility, would it be possible to return the validation output (`invalidaItems` property) as an array of objects? I foresee some additional metadata that might be useful in the future for each validation output, such as `severity`, `jsonPointer`, `sourceMap` etc...
> 
> Can you still fix that @gippy ? I'd love to be sure we don't shoot ourselves in the foot.
> 
> Example:
> 
> ```
> ...
> invalidItems: {
>   0: [
>     {"text": "should have required property 'name'"},
>     {"text": "'bar' should be string"}
>   ],
>   1: [
>     {"text":"'bar' should be string"}
>   ],
> },
> ...
> ```

No problem, this is exactly why this PR exists so that we can come up with the best way to do it. I think it's fine to return objects instead of array of strings. And you are right that this way it will be easier for us to add any other logic later.

fnesveda:
BTW wouldn't the `invalidItems` be better as an array, for easier iteration?
```
"invalidItems": [
    {
        "itemPosition": 0,
        "validationErrors": [
            {"text": "should have required property 'name'"},
            {"text": "'bar' should be string"}
        ]
    },
    {
        "itemPositon": 1,
        "validationErrors": [
            {"text":"'bar' should be string"}
        ]
    }
]
```

fnesveda:
Or why not just forward the raw errors from AJV? They seem to be good enough, we wouldn't have to reinvent the wheel.

```js
import Ajv from 'ajv';
const ajv = new Ajv({ allErrors: true })

const schema = {
    type: "array",
    items: {
        type: "object",
        properties: {
            stringField: { type: "string" },
            integerField: { type: "integer" }
        }
    }
}


const data = [
    { stringField: "abc", integerField: 12345 },
    { stringField: 12345, integerField: 12345 },
    { stringField: "abc", integerField: "abc" },
];

const valid = ajv.validate(schema, data)
if (!valid) console.log(ajv.errors)
```

Gives:
```js
[
  {
    instancePath: '/1/stringField',
    schemaPath: '#/items/properties/stringField/type',
    keyword: 'type',
    params: { type: 'string' },
    message: 'must be string'
  },
  {
    instancePath: '/2/integerField',
    schemaPath: '#/items/properties/integerField/type',
    keyword: 'type',
    params: { type: 'integer' },
    message: 'must be integer'
  }
]
```

gippy:
> Or why not just forward the raw errors from AJV? They seem to be good enough, we wouldn't have to reinvent the wheel.

I have reduced the output from AJV to a single message per error because it's huge and can be very hard to parse (logically, not in code) for developers. But except for huge response in API there is no real issue with returning the whole AJV validation error. The only other thing I can think of is that if we return full AJV object then if developer stores the object as is in dataset which is then visible to user, the user will have to read through the huge object too. But maybe giving developers the whole object makes more sense. Not sure. 

But we will have to always support it that way forever after, keeping it minimal for now would allow us to extend it later/switch to other validation library if AJV is slow.

netmilk:
> BTW wouldn't the `invalidItems` be better as an _array_, for easier iteration?

I would prefer this, too. The index in the key coupling the data with the structure can also harm us in the future.

> But maybe giving developers the whole object makes more sense.

Having the errors compatible with AJV, is a benefit for the tooling. For instance, there might already be some open-source library for highlighting the validation errors in the code, etc...

gippy:
@netmilk @fnesveda Ok, so let's go with this structure:
```json
{
    "error": {
        "type": "schema-validation-error",
        "message": "Schema validation failed",
        "data": {
            "invalidItems": [
                {
                    "itemPosition": 0,
                    "validationErrors": [
                        {
                            "instancePath": "/1/stringField",
                            "schemaPath": "#/items/properties/stringField/type",
                            "keyword": "type",
                            "params": {
                                "type": "string"
                            },
                            "message": "must be string"
                        }
                    ]
                }
            ]
        }
    }
}
```

I do not mind either way. Also if you think the naming should change I'm open to suggestions too :)

fnesveda:
Looks good to me 👍

mtrunkat:
I also prefere the variant of array with `itemPosition` as it's ugly to use strings for number indexes:

CleanShot 2024-06-25 at 14 15 55@2x

One suggestion to this - wouldn't the `index` be better than `itemPosition`? The "item position" seems a bit strange to me but I am not very opinionated on this.

Second - I am afraid a bit that if we return the whole AJV error object then the response will be often larger then request :). Maybe it's too much information?

gippy:
> ```json
> {
>     "error": {
>         "type": "schema-validation-error",
>         "message": "Schema validation failed",
>         "data": {
>             "invalidItems": [
>                 {
>                     "itemPosition": 0,
>                     "validationErrors": [
>                         {
>                             "instancePath": "/1/stringField",
>                             "schemaPath": "#/items/properties/stringField/type",
>                             "keyword": "type",
>                             "params": {
>                                 "type": "string"
>                             },
>                             "message": "must be string"
>                         }
>                     ]
>                 }
>             ]
>         }
>     }
> }
> ```

Hmm I think it should be fine. It can increase our outgoing traffic for dataset insert calls, but it shouldn't be that crazy (as long as someone does not create a huge dataset schema and then starts sending thousands of empty items to it).

I like the name `index` instead of `itemPosition` more. What do you think @fnesveda @netmilk

netmilk:
> I like the name `index` instead of `itemPosition` more. What do you think ...

@gippy I'd prefer it if we could use vocabulary specific to our domain to avoid conflict with a generic `index` property that might appear natively in the programming language consuming the API. What about `itemIndex`? (in reference to SDK [`items` method](https://docs.apify.com/sdk/js/reference/interface/DatasetContent#items) and API [`Get Items` collection of resources](https://docs.apify.com/api/v2#/reference/datasets/item-collection/get-items) )

gippy:
So I decided to go with the original proposal before Mara's comment.
I'm merging this, then will implement the changes in API and the update also the OpenAPI spec.

PR #1081, opened 2024-06-21
feat: update input schema docs

m-murasovs:
Updates the docs to include the new `prefill` property on Boolean fields in the input schema. When the field has a `default` value set now, we display the `default` chip, to differentiate between them and regular/prefilled fields.

Related to https://github.com/apify/apify-core/pull/16390/

Also, polite request - please merge this when the above PR is out - I won't be able to since I won't be at Apify anymore :)


TC-MO (approved):
LGTM

gippy:
> FYI: @gippy , this is out right?

Yes, merging

PR #1091, opened 2024-07-09
feat: publish the new openapi docs on `/api/v2` and move apiary docs to `/api/v2-old`

B4nan:


netmilk (approved):
Looking good - as far as I can tell.

barjin:
I'm no Redocusaurus expert - are the fragments in the URLs the same as in the current docs?

e.g. `https://docs.apify.com/api/v2#/introduction/rate-limiting/per-resource-rate-limit` targets the page about the rate limiting?

barjin:
Ah, cool, it seems to work. lgtm then 👍🏽

PR #1096, opened 2024-07-15
feat: improve standby docs

jirimoravcik:
This PR moves development part of Actor Standby into the corresponding docs section. We also mention standby port in env vars and `usesStandbyMode` in `actor.json` section.


fnesveda (approved):
Looks good, let's just move the "Can I monetize..." section too

TC-MO (approved):
LGTM

Issue #967, opened 2024-04-26
Academy newsletter


honzajavorek:
The [Generating fingerprints](https://docs.apify.com/academy/anti-scraping/mitigation/generating-fingerprints) page contains the following text:

> That's it for the **Mitigation** course for now, but be on the lookout for future lessons! We release lessons as we write them, and will be updating the Academy frequently, so be sure to check back every once in a while for new content! Alternatively, you can subscribe to our mailing list to get periodic updates on the Academy, as well as what Apify is up to.

The words _you can subscribe to our mailing list_ don't link anywhere. Should they? Is there Apify newsletter I should know about? Should I report changes in the Academy to someone from the marketing team? Should the newsletter be marketed like this inside lessons, or should the Docusaurus template take care of it, offering a sign up on every page, somewhere in the footer? Perhaps Academy homepage should have a component for newsletter sign ups?

Alternatively, is it something that never materialized and I should just delete the mention?

honzajavorek:
@mnmkng Do you know about any mailing lists?

mnmkng:
I vaguely remember we had an email form there one day. But we weren't able to regularly create meaningful developer-centric content, so we deleted it.

I would remove the reference.

PR #1106, opened 2024-07-22
fix: remove newsletter mention

honzajavorek:
Close https://github.com/apify/apify-docs/issues/967


honzajavorek:
The link check failure is unrelated.

TC-MO:
Yeah, not sure about this link, it periodically fails the check and when checked manually works... not sure what would be the best course of action in this case, but I guess it's nor relevant to the PR

TC-MO (approved):

honzajavorek:
The link timeouted for me locally as well 🤔 Maybe their site just isn't reliable.

PR #1110, opened 2024-07-23
fix: make anchors in API docs lead to correct position

B4nan:


barjin (approved):
Not that it matters too much, but the other `font-size: 10px` related fixes by @jbartadev seem to define the sizes in pixels (probably to disambiguate in case there is another switch).

PR #1117, opened 2024-07-26
feat: make pay-per-result pricing explicit

mhamas:
Explicit pricing is easier to understand and it better reflects the fact that it is fixed and will not be  necessarily tied to default platform pricing in the future.

CleanShot 2024-07-26 at 14 25 05


TC-MO (approved):
LGTM, I've added note explaining what are `GB-hours` and fixed lint issues

PR #1118, opened 2024-07-28
docs: improve basic commands document

TC-MO:
streamline the prose
restructure the document
use tabs component to render different set of prose for each language

### DIFF
--- a/sources/platform/actors/development/programming_interface/basic_commands.md
+++ b/sources/platform/actors/development/programming_interface/basic_commands.md
@@ -12,17 +12,18 @@ import TabItem from '@theme/TabItem';
 
 ---
 
-The commands described in this section are expected to be called from within a context of a running Actor, both in a local environment or on the Apify platform.
+This page covers essential commands for the Apify SDK in JavaScript & Python. These commands are designed to be used within a running Actor, either in a local environment or on the Apify platform.
 
-## Actor initialization
+## Initialize your Actor
 
-First, the Actor should be initialized. During initialization, it prepares to receive events from the Apify platform. The Actor determines the machine and storage configuration and optionally purges the previous state from local storage. It will also create a default instance of the Actor class.
+Before using any Apify SDK methods, initialize your Actor. This step prepares the Actor to receive events from the Apify paltform, sets up machine and storage configurations, and optionally clears previous local storage states.
 
-It is not required to perform the initialization explicitly because the Actor will initialize on the execution of any Actor method, but we strongly recommend it to prevent race conditions.
+<Tabs groupId="main">
+<TabItem value="JavaScript" label="JavaScript">
 
-### Using the JavaScript SDK
+Use the `init()` method to initialize your Actor. Pair it with `exit()` to properly terminate the Actor
 
-The Actor is initialized by calling the `init()` method. It should be paired with an `exit()` method, which terminates the Actor. The use of `exit()` is not required but recommended. For more information, go to [Exit Actor](#exit-actor).
+The use of `exit()` is not required but recommended. For more information, go to [Exit Actor](#exit-actor).
 
 ```js
 import { Actor } from 'apify';
@@ -33,7 +34,8 @@ console.log('Actor starting...');
 await Actor.exit();
 ```
 
-An alternative way of initializing the Actor is with a `main()` function. This is useful in environments where the latest JavaScript syntax and top-level awaits are not supported. The main function is only syntax-sugar for `init()` and `exit()`. It will call `init()` before it executes its callback and `exit()` after the callback resolves.
+Alternatively, use the `main()` function for environments that don't support top-level awaits:
+
 
 ```js
 import { Actor } from 'apify';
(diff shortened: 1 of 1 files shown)


sources/platform/actors/development/programming_interface/basic_commands.md
```diff
@@ -332,7 +360,6 @@ await Actor.exit();
 
 ```python
 # 😔 Custom handlers are not supported in the Python SDK yet.
```
mtrunkat, line 362: We miss the Python here, is it correct? @vdusek / @janbuchar

sources/platform/actors/development/programming_interface/basic_commands.md
```diff
@@ -33,7 +34,8 @@ console.log('Actor starting...');
 await Actor.exit();
 ```
 
-An alternative way of initializing the Actor is with a `main()` function. This is useful in environments where the latest JavaScript syntax and top-level awaits are not supported. The main function is only syntax-sugar for `init()` and `exit()`. It will call `init()` before it executes its callback and `exit()` after the callback resolves.
+Alternatively, use the `main()` function for environments that don't support top-level awaits:
```
drobnikj, line 37: I would keep info about calling init() and exit().

drobnikj (approved):

sources/platform/actors/development/programming_interface/basic_commands.md
janbuchar: It is, you cannot subscribe to `exit` in Python SDK

sources/platform/actors/development/programming_interface/basic_commands.md
```diff
@@ -12,17 +12,18 @@ import TabItem from '@theme/TabItem';
...
 ---
 
-The commands described in this section are expected to be called from within a context of a running Actor, both in a local environment or on the Apify platform.
+This page covers essential commands for the Apify SDK in JavaScript & Python. These commands are designed to be used within a running Actor, either in a local environment or on the Apify platform.
 
-## Actor initialization
+## Initialize your Actor
 
-First, the Actor should be initialized. During initialization, it prepares to receive events from the Apify platform. The Actor determines the machine and storage configuration and optionally purges the previous state from local storage. It will also create a default instance of the Actor class.
+Before using any Apify SDK methods, initialize your Actor. This step prepares the Actor to receive events from the Apify platform, sets up machine and storage configurations, and optionally clears previous local storage states.
```
B4nan, line 19: the default storage clearing is opt-out, the default behavior is to clear them. this is something we changed with v3, and we are still missing this in the apify cli, there it is still opt-in - but we will be changing this to align the behavior of apify cli with the rest of the tooling in the next major version of the CLI (which is likely to be v1). currently, when you use `apify run`, the CLI effectively disables the auto purge that is otherwise handled by SDK

sources/platform/actors/development/programming_interface/basic_commands.md
```diff
@@ -12,17 +12,18 @@ import TabItem from '@theme/TabItem';
...
 
-It is not required to perform the initialization explicitly because the Actor will initialize on the execution of any Actor method, but we strongly recommend it to prevent race conditions.
+<Tabs groupId="main">
+<TabItem value="JavaScript" label="JavaScript">
 
-### Using the JavaScript SDK
+Use the `init()` method to initialize your Actor. Pair it with `exit()` to properly terminate the Actor
```
B4nan, line 24:
```suggestion
Use the `init()` method to initialize your Actor. Pair it with `exit()` to properly terminate the Actor.
```

sources/platform/actors/development/programming_interface/basic_commands.md
```diff
@@ -12,17 +12,18 @@ import TabItem from '@theme/TabItem';
...
 
-The Actor is initialized by calling the `init()` method. It should be paired with an `exit()` method, which terminates the Actor. The use of `exit()` is not required but recommended. For more information, go to [Exit Actor](#exit-actor).
+The use of `exit()` is not required but recommended. For more information, go to [Exit Actor](#exit-actor).
```
B4nan, line 26: the exit call is kinda required, if you use `init`, you open websocket connection, and without the `exit` call your program would not finish. at least that's how it used to work and i am not aware of any changes.

sources/platform/actors/development/programming_interface/basic_commands.md
TC-MO: so just removing `optionally` here  will be sufficient as right now this is default behaviour to purge the local states? But in the future once apify-cli moves to v1 it will become optional?

B4nan: it depends on what you are describing here, if its apify CLI usage, then right now, purging is still explicit, you need `-p`. but the behavior of SDK and crawlee are auto-purge, the CLI controls this explicitly via env vars. in CLI v1 we will unify the behavior so `apify run` will also auto-purge (and you will have an opt-out flag like `--no-purge`, with an alias `--resurrect`)

TC-MO: This is SDK behaviour so it would so just removing `optionally` seems like easiest choice as it should be auto-purge

PR #1120, opened 2024-07-28
docs: improve status message doc

TC-MO:
rework logical flow
add admonitions within code tabs relevant to respective language rework structure of the doc

### DIFF
--- a/sources/platform/actors/development/programming_interface/status_messages.md
+++ b/sources/platform/actors/development/programming_interface/status_messages.md
@@ -74,11 +74,5 @@ async def main():
         # INFO  [Status message]: Crawled 45 of 100 pages
 ```
 
-:::note Update frequency
-
-You can call the `set_status_message` function as often as needed. The SDK only invokes the API if the status changes, simplifying usage.
-
-:::
-
 </TabItem>
 </Tabs>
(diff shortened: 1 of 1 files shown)


sources/platform/actors/development/programming_interface/status_messages.md
```diff
@@ -47,6 +56,12 @@ await Actor.setStatusMessage('Crawled 45 of 100 pages');
 await Actor.exit();
 ```
 
+:::note Update frequency
+
+You can call the `setStatusMessage` function as often as needed. The SDK only invokes the API if the status changes, simplifying usage.
```
B4nan, line 61: I don't think this is accurate, we don't check this anywhere in the SDK, we just call the API on each method invocation, at least in the JS version. 

https://github.com/apify/apify-sdk-js/blob/master/packages/apify/src/actor.ts#L555

TC-MO: So it would make the most sense to just remove this note? Or just a rewording removing the second part would be better as in:

```suggestion
You can call the `setStatusMessage` function as often as needed.
```

B4nan: I recall someone told me there is a 10s interval on apify platform that refreshes the state, so technically it makes no sense to call it more often than that - we respect this in crawlee and call this periodically every 10s from the basic crawler. Not sure if it's still valid (or if the interval changed). I would maybe just remove this note completely.

TC-MO: that makes sense , thank you

B4nan (approved):

PR #1122, opened 2024-07-28
docs: improve container web server doc

TC-MO:
rework doc structure
streamline the prose


drobnikj:
The changes look good. I have just an idea about standby mode. It is worth mentioning here that we allow this option, which is better suited for web servers in Actor. Not sure who handle docs about standby cc @fnesveda .

drobnikj (approved):

fnesveda:
@drobnikj Good idea, let's merge this one and then I can send a PR with some mention of the Standby mode in the updated docs

PR #1126, opened 2024-07-30
chore: typo fixes, proxy IP availability

seyhello:


sources/platform/proxy/usage.md
```diff
@@ -32,7 +32,7 @@ If you need to test Apify Proxy before you subscribe, please [contact our suppor
 
 | Parameter           | Value / explanation |
 |---------------------|---------------------|
-| Hostname            | `proxy.apify.com`, alternatively you can use static IP addresses `18.208.102.16`, `35.171.134.41`. |
+| Hostname            | `proxy.apify.com`, alternatively you can use static IP addresses `18.208.102.16`, `35.171.134.41`. However, it is **strongly recommended to use the DNS name** (`proxy.apify.com`) as IP addresses are only available based on the current backend targets. |
```
fnesveda, line 35: Is there any reason someone might want to connect directly through the IP addresses? I would be decisive and not list the direct connection option at all, and list only the hostname here, and then list the IP addresses in a separate paragraph outside the table. E.g. "the proxy runs on these two IP addresses, if you need to whitelist them in your firewall".

fnesveda (changes requested):
Looks good, one note about the IP addresses 🙂

sources/platform/proxy/usage.md
seyhello: please, check last commit

fnesveda (approved):
Cool 👏

TC-MO (approved):
LGTM

PR #1140, opened 2024-08-07
feat: change link checking to periodic

TC-MO:
Given the amount of false positives we run into at every time when we are adding a new page to docs, it makes sense to change the link checking job to periodic instead of at every PR.

### DIFF
--- a/.github/workflows/lychee.yml
+++ b/.github/workflows/lychee.yml
@@ -1,55 +1,55 @@
 name: Periodic Link Checker
 
 on:
-  schedule:
-    - cron: '0 0 * * *'  # Run daily at midnight UTC
-  workflow_dispatch:  # Allow manual triggering
+    schedule:
+        - cron: '0 0 * * *'  # Run daily at midnight UTC
+    workflow_dispatch:  # Allow manual triggering
 
 jobs:
-  link-check:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
+    link-check:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
 
-      - name: Use Node.js 20
-        uses: actions/setup-node@v4
-        with:
-          node-version: 20
-          cache: 'npm'
-          cache-dependency-path: 'package-lock.json'
-          always-auth: 'true'
-          registry-url: 'https://npm.pkg.github.com/'
-          scope: '@apify-packages'
+            - name: Use Node.js 20
+              uses: actions/setup-node@v4
+              with:
+                  node-version: 20
+                  cache: 'npm'
+                  cache-dependency-path: 'package-lock.json'
+                  always-auth: 'true'
+                  registry-url: 'https://npm.pkg.github.com/'
+                  scope: '@apify-packages'
 
-      - name: Build docs
-        run: |
-          npm ci --force
-          npm update @apify/openapi
-          npm run build
-        env:
-          APIFY_SIGNING_TOKEN: ${{ secrets.APIFY_SIGNING_TOKEN }}
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-          SMARTLOOK_PROJECT_KEY: ${{ secrets.SMARTLOOK_PROJECT_KEY }}
+            - name: Build docs
+              run: |
+                  npm ci --force
+                  npm update @apify/openapi
+                  npm run build
+              env:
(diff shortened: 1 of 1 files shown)


B4nan (changes requested):
Lets revert the whitespace changes unless the other workflows use 2 spaces for indents

.github/workflows/lychee.yml
```diff
@@ -1,36 +1,55 @@
...
+                      --verbose
+                      --no-progress
+                      --accept '100..=103,200..=299,403..=403,429'
+                      --output ./lychee/out.md
+                      --format markdown
+                      './build/**/*.html'
+              continue-on-error: true
+
+            - name: Add job summary
+              if: always()
```
B4nan, line 54: whats the point of this?

.github/workflows/lychee.yml
```diff
@@ -1,36 +1,55 @@
...
+    schedule:
+        - cron: '0 0 * * *'  # Run daily at midnight UTC
+    workflow_dispatch:  # Allow manual triggering
 
 jobs:
     link-check:
         runs-on: ubuntu-latest
         steps:
-            -   uses: actions/checkout@v4
+            - uses: actions/checkout@v4
```
B4nan, line 12: please revert those whitespace changes too, the indent should be 4 spaces, not 4 and sometimes 2 at the end...

B4nan: idk how you format this, webstorm handles this correctly out of box, here is the full formatted file:

```yml
name: Periodic Link Checker

on:
    schedule:
        -   cron: '0 0 * * *'  # Run daily at midnight UTC
    workflow_dispatch:  # Allow manual triggering

jobs:
    link-check:
        runs-on: ubuntu-latest
        steps:
            -   uses: actions/checkout@v4

            -   name: Use Node.js 20
                uses: actions/setup-node@v4
                with:
                    node-version: 20
                    cache: 'npm'
                    cache-dependency-path: 'package-lock.json'
                    always-auth: 'true'
                    registry-url: 'https://npm.pkg.github.com/'
                    scope: '@apify-packages'

            -   name: Build docs
                run: |
                    npm ci --force
                    npm update @apify/openapi
                    npm run build
                env:
                    APIFY_SIGNING_TOKEN: ${{ secrets.APIFY_SIGNING_TOKEN }}
                    NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
                    SMARTLOOK_PROJECT_KEY: ${{ secrets.SMARTLOOK_PROJECT_KEY }}

            -   name: Run Lychee Link Checker
                id: lychee
                uses: lycheeverse/lychee-action@v1.10.0
                env:
                    GITHUB_TOKEN: ${{ secrets.APIFY_SERVICE_ACCOUNT_GITHUB_TOKEN }}
                with:
                    fail: false
                    args: >
                        --base https://docs.apify.com
                        --exclude-path 'build/versions.html'
                        --max-retries 6
                        --verbose
                        --no-progress
                        --accept '100..=103,200..=299,403..=403,429'
                        --output ./lychee/out.md
                        --format markdown
                        './build/**/*.html'
                continue-on-error: true

            -   name: Add job summary
                if: always()
                run: cat ./lychee/out.md >> $GITHUB_STEP_SUMMARY
```

B4nan (changes requested):
i

.github/workflows/lychee.yml
TC-MO: My understanding is that this creates job summary that can be later found within Actions UI based on the .md file created during the run.

TC-MO:
Thank you, not sure what is the issue with VSC and formatting ¯\_(ツ)_/¯

.github/workflows/lychee.yml
B4nan: i mean what is the point of using `if: always()`? removing that line should have the same effect unless i am missing something :]

TC-MO: Ah fair enough, this was supposed to make sure job summary gets created even if previous steps fail, but not sure if this is correct way to do it and if you think it should be removed , let's remove it

B4nan: oh i see, never saw it before. i dont mind keeping it there if it actually does something, up to you really, maybe keep it and remove once you see its not adding value for the broke builds?

TC-MO: that makes sense to me

TC-MO:
@B4nan are any other changes needed or can we move forward with merging thit change?

B4nan (approved):

PR #1141, opened 2024-08-07
feat: add new typos PR check

TC-MO:
add PR check for new typos action
add _typos.toml to configure ignores for the new action

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,0 +1,8 @@
+[default]
+extend-ignore-re = [
+    '`[^`\n]+`',
+    '```[\s\S]*?```',
+]
+
+[default.extend-words]
+SER = "SER"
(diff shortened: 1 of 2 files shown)


B4nan (approved):
nice

_typos.toml
```diff
@@ -0,0 +1,8 @@
+[default]
+extend-ignore-re = [
```
honzajavorek, line 2: Would you mind dropping a comment here about what these REs are supposed to ignore? Anyone who comes next to add or debug something will love you (including future you, I think 😄). In TOML, you can do

```toml
extend-ignore-re = [
    '`[^`\n]+`',  # skip ...?
    '```[\s\S]*?```',  # skip code blocks
]
```

B4nan: i believe the first one is an inline code block

honzajavorek: Aah, yes, that would be it 💡

PR #1157, opened 2024-08-21
fix: Update main.py

alperkaya:
There is a change in BBC weather which made the previous script obsolete.


vdusek (approved):
LGTM, thanks.

PR #1158, opened 2024-08-22
docs(platform): update run data retention info

katacek:
Change Free Retention "microcopy" -> following the update in console (apify-core repo), link to zenhub issue [here](https://app.zenhub.com/workspaces/platform-team-5f6454160d9f82000fa6733f/issues/gh/apify/apify-core/15694)

I have updated 2 places -> platform: runs and builds, and tutorials: tasks. Please, let me know if there is some other page where it suits well, thanks!

Issue:

> Right now we are (supposed to be) preserving the last 10 runs that a free-user made beyond 30 days. This is for reactivation and debugging purposes. We don't explain this to the user at all, so we should do a better job of that.

Updated sections in docs ⬇️ 
Screenshot 2024-08-22 at 10 46 57
Screenshot 2024-08-22 at 10 45 13

### DIFF
--- a/sources/academy/tutorials/apify_scrapers/getting_started.md
+++ b/sources/academy/tutorials/apify_scrapers/getting_started.md
@@ -73,7 +73,7 @@ In the settings tab, you can set options that are common to all tasks and not di
 
 ### [](#runs) Runs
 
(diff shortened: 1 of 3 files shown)


katacek:
> It looks pretty good to me. I think I would just add the note in usage>data retention as well.
> 
> https://docs.apify.com/platform/storage/usage#data-retention

thanks, updated as well

Screenshot 2024-08-23 at 12 43 39

TC-MO (approved):

PR #1167, opened 2024-08-26
docs: fix outdated docs re: create new actor button

TC-MO:
add new screenshot with correct UI
change prose to reflect the current state of UI


mtrunkat (approved):

honzajavorek:
The original `create-new-actor.png` is linked from other pages, or could it be removed now? I don't see it removed it the PR.

PR #1170, opened 2024-08-26
fix: Fix mobile screen issues

### LINKED ISSUE #1128: The line numbers are overlapping the code in codeblock when horizontally scrolling
see the video in the report

reported here:
https://apify.slack.com/archives/C0L33UM7Z/p1722258775097099

### LINKED ISSUE #1129: Videos are too wide on mobile screens
The video should be aligned with the text content to avoid creating an empty side area. It will also make it easier to watch it.
reported here:
https://apify.slack.com/archives/C0L33UM7Z/p1722259148276879

baldasseva:
Closes #1128
Closes #1129

### DIFF
--- a/apify-docs-theme/src/theme/custom.css
+++ b/apify-docs-theme/src/theme/custom.css
@@ -1579,14 +1579,9 @@ html[data-theme='dark'] .runnable-code-block svg .apify-logo {
     top: -13rem;
 }
 
-.youtubeVideoPlayer {
-    position: relative;
-    padding-bottom: 56.25%; /* 16:9 */
-}
-.youtubeVideoPlayer iframe {
-    position: absolute;
-    top: 0;
-    left: 0;
+iframe[src*="youtube"] {
     width: 100%;
-    height: 100%;
+    aspect-ratio: 16/9;
+    height: auto;
+    margin-bottom: 1.6rem;
 }
--- a/apify-docs-theme/src/theme/custom.css
+++ b/apify-docs-theme/src/theme/custom.css
@@ -1560,7 +1560,6 @@ html[data-theme='dark'] .runnable-code-block svg .apify-logo {
     min-width: 1.5rem;
     display: inline-block;
     opacity: .3;
-    position: sticky;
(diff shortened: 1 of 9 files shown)


apify-docs-theme/src/theme/custom.css
```diff
@@ -1560,7 +1560,6 @@ html[data-theme='dark'] .runnable-code-block svg .apify-logo {
     min-width: 1.5rem;
     display: inline-block;
     opacity: .3;
-    position: sticky;
```
baldasseva, line 1563: Now it's the same as on web, the numbers just scroll with the code (e.g. here https://apify.com/compass/google-maps-extractor/api/client/nodejs )

B4nan:
FYI you need to use separate `closes` keyword, once for each issue.

```
// this will close only issue 1128
Closes #1128 , #1129

// this will close both issues
Closes #1128
Closes #1129
```

apify-docs-theme/src/theme/custom.css
```diff
@@ -1579,3 +1578,15 @@ html[data-theme='dark'] .runnable-code-block svg .apify-logo {
 .apiPage .tsd-panel-header .tsd-anchor-id {
     top: -13rem;
 }
+
+.youtubeVideoPlayer {
```
webrdaniel, line 1582: If it is possible I would prefer a solution without the wrapper. Because someone can easily forget to use it when adding a new video. Maybe we can use this selector to target just the youtube frames `iframe[src*="youtube"]`

baldasseva: Forget about this comment. It is in fact possible haha

webrdaniel (approved):
🆒

apify-docs-theme/src/theme/custom.css
B4nan: keep in mind youtube also uses other domains like youtu.be

B4nan (approved):
otherwise it looks good to me, havent tried running it locally tho

PR #1175, opened 2024-09-02
feat: Remove smartlook and add intercom messenger

webrdaniel:
- Remove Smartlook as we no longer use it
- Adding Intercom messenger script to allow AI chat
 - added a new variable to the Intercom for this

The growth team has to config properly the messenger window for docs in Intercom and give us a green light to deploy this.

### DIFF
--- a/apify-docs-theme/src/theme.js
+++ b/apify-docs-theme/src/theme.js
@@ -132,21 +132,6 @@ This either means that your Docusaurus setup is misconfigured, or that your GitH
             o.plugins.push(postcssPreset); // allow newest CSS syntax
             return o;
         },
-        injectHtmlTags() {
-            return {
-                headTags: [
-                    process.env.INTERCOM_APP_ID && {
-                        tagName: 'script',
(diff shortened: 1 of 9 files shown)


apify-docs-theme/src/theme.js
```diff
@@ -132,6 +132,21 @@ This either means that your Docusaurus setup is misconfigured, or that your GitH
             o.plugins.push(postcssPreset); // allow newest CSS syntax
             return o;
         },
+        injectHtmlTags() {
```
jbartadev, line 135: Questing: @webrdaniel can't we use the https://www.npmjs.com/package/@gracefullight/docusaurus-plugin-intercom plugin for this?

webrdaniel: I was kinda discouraged by only 12 weekly downloads but yeah we can use this 👍

webrdaniel: I have some issues getting it working with the plugin. It is producing an invalid script. So I guess we will stick with the manual implementation
Screenshot 2024-09-02 at 11 53 41

B4nan (approved):

jbartadev (approved):
I'm cool with that if it's going to work 👍

PR #1176, opened 2024-09-02
chore: bump Docusaurus to v3

barjin:
Following the bump of `@apify/docs-search-modal` to Docusaurus v3, this PR bumps the Docusaurus dependency version in `@apify/docs-theme` and the Apify Docs themselves. 

This is required for Docusaurus version bump in all the subprojects (SDK docs, Client docs etc.)

The changes follow the [official v3 migration guide](https://docusaurus.io/docs/migration/v3).


sources/platform/integrations/ai/langchain.md
```diff
@@ -126,7 +126,7 @@ LangChain is a standard interface through which you can interact with a variety
 
 ## Resources
 
-- <https://python.langchain.com/docs/get_started/introduction>
-- <https://python.langchain.com/docs/integrations/providers/apify>
-- <https://python.langchain.com/docs/integrations/tools/apify>
-- <https://python.langchain.com/docs/modules/model_io/llms/>
+- [https://python.langchain.com/docs/get_started/introduction](https://python.langchain.com/docs/get_started/introduction)
```
B4nan, line 129: do we really need to do this? i thought this is enough
```suggestion
- https://python.langchain.com/docs/get_started/introduction
```

sources/academy/platform/get_most_of_actors/naming_your_actor.md
```diff
@@ -21,23 +21,23 @@ For Actors such as [YouTube Scraper](https://apify.com/bernardo/youtube-scraper)
 
 GOOD:
 
-- Technical name (Actor's name in the [Apify Console](https://console.apify.com)): **${domain}-scraper**, e.g. **youtube-scraper**.
-- Publication title for the Apify Store: **${Domain} Scraper**, e.g. **YouTube Scraper**.
-- Name of the GitHub repository: **actor-${domain}-scraper**, e.g. **actor-youtube-scraper**.
+- Technical name (Actor's name in the [Apify Console](https://console.apify.com)): **$\{domain\}-scraper**, e.g. **youtube-scraper**.
+- Publication title for the Apify Store: **$\{Domain\} Scraper**, e.g. **YouTube Scraper**.
```
B4nan, line 25: iirc we only need to escape the opening bracket, the upgrading guide kinda confirms this

https://docusaurus.io/docs/migration/v3#bad-usage-of-

sources/platform/integrations/ai/langchain.md
barjin: Uh oh, apparently the markdown linter in the CI checks [doesn't like bare URLs](https://github.com/apify/apify-docs/actions/runs/10679542720/job/29599005818?pr=1176). 

And the apparent fix is to [enclose it in angle brackets](https://github.com/updownpress/markdown-lint/blob/master/rules/034-no-bare-urls.md), isn't that convenient 🙃 

I'm all for bare URLs; you're right that writing the same URL twice makes it less readable (i.e., not really what using linter should do). I'll check what we can do to disable this rule in markdownlint - is that okay with you @TC-MO ?

B4nan:
we will also need to wait for https://github.com/milesj/docusaurus-plugin-typedoc-api/pull/156 to be merged/released, or alter those imports via patch package, some projects like the js SDK use versioning and this will pop up in those

barjin:
Right, I already fixed this (for Python Crawlee) in our fork of that package, but you're right that we're using the original package in the JS projects.

Maybe it's time to switch the JS projects to our fork too? Once we want to make some changes to how the rendered docs look like (you know, the ones that I was promising since 6 months ago), we can deploy it over all our projects at once (Python and JS alike).

B4nan:
yeah, if its still compatible with js projects, lets do that

apify-docs-theme/src/theme/DocBreadcrumbs/index.js
```diff
@@ -1,9 +1,9 @@
 import Link from '@docusaurus/Link';
-import { ThemeClassNames } from '@docusaurus/theme-common';
 import {
     useSidebarBreadcrumbs,
-    useHomePageRoute,
-} from '@docusaurus/theme-common/internal';
+} from '@docusaurus/plugin-content-docs/client';
+import { ThemeClassNames } from '@docusaurus/theme-common';
+import { useHomePageRoute } from '@docusaurus/theme-common/internal';
```
B4nan, line 6: i thought the `/internal` imports are no longer allowed?

sources/academy/tutorials/apify_scrapers/puppeteer_scraper.md
```diff
@@ -11,7 +11,7 @@ slug: /apify-scrapers/puppeteer-scraper
 
 #
```
B4nan, line 12: isn't this weird? does it do something?

B4nan (approved):
lgtm

apify-docs-theme/src/theme/DocBreadcrumbs/index.js
barjin: From what I understood, the Docusaurus team removed some `theme-common` package dependencies, as they didn't make sense from the "big-picture" POV. The `theme-common/internal` module was reexporting some of the symbols from those dependencies, which then indeed stopped working.

The `useHomePageRoute` hook is defined inside of `theme-common`, though, so it still works. Honestly, I don't really like importing something called `internal` either, but... as long as it works, I guess I'm fine?

B4nan: yeah sure, that's why I gave you the checkmark right ahead, build is passing, so it's fine.

PR #1180, opened 2024-09-03
fix: `publish-theme` action uses `secrets.NPM_TOKEN`

barjin:
Wherever applicable, this PR changes the NPM token to `secrets.NPM_TOKEN` to fix the issue with the `ui-packages` [STILL NOT BEING PUBLIC](https://github.com/apify/apify-docs/actions/runs/10681779569/job/29606166939) (I will never stop hating this.)

This does not replace the last occurrence of the token since I believe we need the old token (`APIFY_SERVICE_ACCOUNT_NPM_TOKEN`) to publish packages in the @apify workspace. This might be a question for @fnesveda - are any of the mentioned GH secrets organization-wide (and do you think we can publish packages with `secrets.NPM_TOKEN`?)

### DIFF
--- a/.github/workflows/publish-theme.yaml
+++ b/.github/workflows/publish-theme.yaml
@@ -55,7 +55,7 @@ jobs:
                     echo "access=public" > ~/.npmrc
                     echo "@apify-packages:registry=https://npm.pkg.github.com/" >> ~/.npmrc
                     echo "//registry.npmjs.org/:_authToken=${{ secrets.APIFY_SERVICE_ACCOUNT_NPM_TOKEN }}" >> ~/.npmrc
-                    echo "//npm.pkg.github.com/:_authToken=${{ secrets.NPM_TOKEN }}" >> ~/.npmrc
+                    echo "//npm.pkg.github.com/:_authToken=${{ secrets.APIFY_SERVICE_ACCOUNT_GITHUB_TOKEN }}" >> ~/.npmrc
 
             -   name: Bump the theme version
                 run: |
--- a/.github/workflows/publish-theme.yaml
+++ b/.github/workflows/publish-theme.yaml
@@ -32,7 +32,7 @@ jobs:
 
     publish:
         needs: look_for_change
-        if: needs.look_for_change.outputs.theme_changed == 'true'
+        if: ${{ needs.look_for_change.outputs.theme_changed == 'true' || github.event_name == 'workflow_dispatch' }}
         runs-on: ubuntu-latest
         steps:
             -   uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


barjin:
As discussed with @fnesveda yesterday, the NPM / Github token management is a total mess. 

The latest commit puts all the token-related stuff in the `~/.npmrc` file and expects all the scripts to take it from there (to avoid passing different-named envvars with various values to different places).

The commit also adds `workflow_dispatch` trigger for the publishing workflow so we can test this better.

.github/workflows/publish-theme.yaml
```diff
@@ -45,32 +46,27 @@ jobs:
...
                 run: |
                     git config --global user.name "Apify Release Bot"
                     git config --global user.email "noreply@apify.com"
 
-                    echo "access=public" > .npmrc
-                    echo "//registry.npmjs.org/:_authToken=${{ secrets.APIFY_SERVICE_ACCOUNT_NPM_TOKEN }}" >> .npmrc
+                    echo "access=public" > ~/.npmrc
+                    echo "@apify-packages:registry=https://npm.pkg.github.com/" >> ~/.npmrc
+                    echo "//registry.npmjs.org/:_authToken=${{ secrets.APIFY_SERVICE_ACCOUNT_NPM_TOKEN }}" >> ~/.npmrc
+                    echo "//npm.pkg.github.com/:_authToken=${{ secrets.NPM_TOKEN }}" >> ~/.npmrc
```
fnesveda, line 58: Please use this one instead, to make it more explicit which token it is. We'll eventually get rid of all the badly named tokens across the organization.

```suggestion
                    echo "//npm.pkg.github.com/:_authToken=${{ secrets.APIFY_SERVICE_ACCOUNT_GITHUB_TOKEN }}" >> ~/.npmrc
```

fnesveda (changes requested):
One more note, otherwise 🤞

.github/workflows/publish-theme.yaml
B4nan: Note that the `NPM_TOKEN` is a repo override, its not even the one from apify org, I think it might be mine, I recall things got broken with the token that was in the org some time ago. We should clear this up finally and remove the repo override from here.

fnesveda (approved):

.github/workflows/publish-theme.yaml
barjin: Yeah, the "if it works, don't touch it" rule went a long way here... this token is now used in all the workflows in this repo that access NPM in some way 😬  I'm glad it's changing though.

B4nan (approved):

PR #1184, opened 2024-09-04
fix: escape MDX characters in changelog (Docusaurus v3)

barjin:
`<` and `{` now have special meaning in Docusaurus v3 MDX syntax. In case those characters appear in `changelog.md` file, this breaks the documentation builds.


janbuchar (approved):
LGTM

PR #1186, opened 2024-09-05
fix: remove race condition from the changelog generation

barjin:
Removes the reintroduced changelog generator race condition. 

Fixes the `options.changelogFromRoot` bug when the default changelog is never overwritten (default changelog is always newer than the source changelog) by tracking which changelogs are default and can be overwritten.


janbuchar (approved):
LGTM

PR #1187, opened 2024-09-05
fix: double wrapping codeblocks in `MDXPre`

### LINKED ISSUE #1181: Code blocks look like they are zoomed out, wide code needs scrolling for copy button
Code blocks inside docs seem like they are zoomed out? or maybe there is some weird margin set there? (see screen #1 & #2)
Also when we have really wide code example the `copy code` button only shows when scrolled to the end (see screen #3)
image
image
image

Reported on Slack https://apify.slack.com/archives/C0L33UM7Z/p1725366285517509

barjin:
Closes #1181 by removing the custom `MDXPre` component implementation. The underlying issue this component was addressing was likely fixed in Docusaurus v3.


janbuchar (approved):
Seems legit, if I understand how docusaurus themes work correctly

PR #1189, opened 2024-09-06
fix: make TOC rendering consistent regardless of the page content

### LINKED ISSUE #1188: Issues with broken ToC layout
There are a few issues connected to the ToC in the right column on the page.
1. on some pages, it gets displayed under the main article
	- e.g. https://docs.apify.com/platform/actors/development/actor-definition/input-schema/specification/v1
	- it's probably because of some weird interaction between our docs theme, `display: flex` on the main layout and `overflow: auto` on code blocks
	- if you delete long lines in code blocks (e.g. lines 43 and 406 [here](https://github.com/apify/apify-docs/blob/2b1e1eeaeb65ebe493189fb37fbf8dec5d91ed42/sources/platform/actors/development/actor_definition/input_schema/specification.md?plain=1)), the layout gets fixed
2. on some pages, scrolling through the page makes the laout wider / narrower, resulting in the article jumping around
	- it might be connected to long subheadings being marked as current in the ToC

https://github.com/user-attachments/assets/4590a167-a0d7-426b-9648-e3ef87a8948b

Reported here: https://apify.slack.com/archives/C0L33UM7Z/p1725462627933989
cc @barjin

barjin:
Sets `flex-wrap: nowrap` to allow for "overflowing" TOC when too wide. 

Swizzles the `TOCItems` component into `@apify/docs-theme`. Uses a [hack](https://stackoverflow.com/questions/556153/inline-elements-shifting-when-made-bold-on-hover) with `::before` pseudoclass to reserve space for the bold TOC items, so the element width doesn't change on selection.

Closes #1188


apify-docs-theme/src/theme/TOCItems/styles.module.css
```diff
@@ -0,0 +1,12 @@
...
+}
+
+.apifyTocLink::before {
+    display: block;
+    content: attr(data-label);
+    font-weight: 700;
+    height: 0;
+    overflow: hidden;
+    visibility: hidden;
+}
```
B4nan, line 12:
```suggestion
}

```

src/theme/DocItem/Layout/styles.module.css
```diff
@@ -21,4 +21,8 @@
   .docItemCol {
     max-width: 820px !important;
   }
+}
+
+.row {
+  flex-wrap: nowrap;
 }
```
B4nan, line 28:
```suggestion
}

```

B4nan (approved):

PR #1190, opened 2024-09-09
chore: throw on build with broken anchors

### LINKED ISSUE #952: Check broken anchors
This is more of a suggestion or an idea for future. We already check whether links are broken. We could check if anchors we point to in our documents actually exist. Those links are valid as in that the page exists and will load in reader's browser, but if the link contains an anchor, the intention was to point to a specific paragraph.

This is something I think Docusaurus could check out of the box (MkDocs has similar features) or that could be achieved by a relatively simple script in our CI, which goes through the resulting HTML and fails if something is off. In the past I wrote a similar script for myself in Python.

barjin:
Docusaurus v3 adds the [`onBrokenAnchors`](https://docusaurus.io/docs/api/docusaurus-config#onBrokenAnchors) setting that allows us to fail the build when Docusaurus finds broken anchor links (internal fragment links). 

Closes #952


TC-MO (approved):
Ready for merge, 
Changed to `warn` it still shows a lot of API ref links but they work when clicked manually, all other broken anchors have been fixed

barjin:
Looking good to me, thank you and feel free to merge @TC-MO 👍🏽  Given that this fixes broken links, I don't think anybody's gonna object 😄

TC-MO:
Yeah I doubt anyone would object to 100+ fixed links :D

PR #1201, opened 2024-09-10
style: let the reader decide if something is simple or easy (upstream branch)

honzajavorek:
Closes https://github.com/apify/apify-docs/pull/1071. It's the same PR, but from an upstream branch, not from my fork. As discussed on Slack, this works around unauthorized npm token on CI.

I also rebased the original branch on top of current master, hopefully without mistakes. Otherwise the changes should be the same.


honzajavorek:
So I went through the endless diff myself now, once again, just to be super sure I didn't mess up during the rebase. All seems good 🙏 @TC-MO can I have one more rubber stump here, please? (I know you've already approved the previous one, but GitHub won't allow me to merge this without explicit review.)

PR #1202, opened 2024-09-10
fix: avoid permanent redirects (upstream branch)

honzajavorek:
Closes https://github.com/apify/apify-docs/pull/1101. It's the same PR, but from an upstream branch, not from my fork. As discussed on Slack, this works around unauthorized npm token on CI.

I also rebased the original branch on top of current master, hopefully without mistakes. Otherwise the changes should be the same.


honzajavorek:
@TC-MO One more rubber stamp, please 🙏 I went through the diffs after rebasing, caught some leftovers and fixed them in d7894d6, now it looks okay.

TC-MO (approved):

PR #1203, opened 2024-09-10
docs: Add documentation for Milvus/Zilliz database integration

jirispilka:
This integration is similar to Pinecone/Qdrant. I'm writing a tutorial on how to use Apify with Milvus/Zilliz, and it would be good to reference our documentation in the tutorial.

Just a note: Milvus is an open-source vector database, and Zilliz offers a managed solution based on Milvus.

@TC-MO, could you please review the English? I haven't included any screenshots, as I believe the description should suffice to get started. Also, maintain and update screenshot is a bit painful. 

Additionally, I haven't added this integration to the [cards](https://docs.apify.com/platform/integrations#data-pipelines-etls-and-aillm-tools). Do you think it should be included there?


jirispilka:
> I don't see any reason to not add it to the cards, so we can move forward with this

Thanks 👍🏻  added.

TC-MO (approved):
Content-wise LGTM, I was wondering whether we should somehow inform readers that this will be done through Zilliz through admonitions or something, but I guess we've set the precedent with Telegram integration that is done the same way through Zapier. Are ther plans to create guide that shows direct Milvus integration without 3rd party solutions?

PR #1207, opened 2024-09-11
fix: revert details component rendering to pre-v3 state

### LINKED ISSUE #1199: The <Details /> component doesn't present summary very well
The text of the `<summary>` tag is indistinguishable from the rest of the contents. It also isn't visible when the component is collapsed. There's just a placeholder text _Details_, which isn't helpful.

Previously we were using native HTML `<details>` elements, just styled with CSS. After the Docusaurus v3 upgrade or perhaps with other changes, the docs now use a custom `<Details />` component. Is there a reason why we do so?

## Collapsed

Screenshot 2024-09-10 at 10 31 54

Screenshot 2024-09-10 at 10 32 15

## Open

Screenshot 2024-09-10 at 10 32 33

Screenshot 2024-09-10 at 10 32 51

## How it should work

Screenshot 2024-09-10 at 10 35 25

barjin:
Fixes the mismatch between the `<details>` element pre- and post- the Docusaurus v3 update. Caused by Docusaurus expecting (some?) MDX component names being capitalized. 

Closes #1199

### DIFF
--- a/apify-docs-theme/src/theme/MDXComponents/Details.js
+++ b/apify-docs-theme/src/theme/MDXComponents/Details.js
@@ -6,7 +6,7 @@ export default function MDXDetails(props) {
     // Split summary item from the rest to pass it as a separate prop to the
     // Details theme component
     const summary = items.find(
-        (item) => React.isValidElement(item) && item.props?.mdxType === 'summary',
+        (item) => React.isValidElement(item) && item.type === 'summary',
     );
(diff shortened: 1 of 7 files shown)


barjin:
Also, the chevron on the left doesn't seem to be the right size (and position)... Maybe yet another `<html style="font-size: 10px">` collateral @jbartadev ?

obrazek

PR #1209, opened 2024-09-11
chore: remove patching

honzajavorek:
Previously there has been a patch to modify front matter in Docusaurus v2, but with migration to Docusaurus v3, this has been moved to a config file and patching is not necessary anymore. The patch itself has been removed from the codebase, but I noticed there are leftovers in form of a dependency and a postinstall script.

Not sure who can review this, so I'm sorry if I asked the wrong people 🙏

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -58,7 +58,6 @@
                 "globby": "^14.0.0",
                 "markdownlint": "^0.35.0",
                 "markdownlint-cli": "^0.41.0",
-                "patch-package": "^8.0.0",
                 "path-browserify": "^1.0.1",
                 "rimraf": "^6.0.0",
                 "typescript": "^5.1.3"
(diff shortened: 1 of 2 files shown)


honzajavorek:
I noticed it after seeing this in my terminal:

```
❯ npm ci

> apify-docs@2.0.0 postinstall
> patch-package

patch-package 8.0.0
Applying patches...
No patch files found

added 2734 packages, and audited 2736 packages in 4m
```

I didn't measure it, but I thought - oh, perhaps this can be a few ms faster if we skip the moot patching. Ofc I have no idea what are your plans further down the roadmap, so if you expect more patching to happen soon, it doesn't make sense to remove it.

B4nan:
yeah, a few milliseconds maybe, but the install part took 4 minutes, this wont be measurable :] 

but let's remove it, its not a big deal adding it back when needed.

honzajavorek:
That was a coffee place wifi install 😅 But yeah 😀

PR #1210, opened 2024-09-11
docs: update-privacy-policy.md

lenkab151:
new version of the privacy policy


TC-MO:
I think in this case we don't need to follow markdownlint guidelines as here we value proper formatting and readability more than set of rules set for the rest of documentation. I will go ahead and try to disable markdownlint for any content within `/legal`

mnmkng (approved):

PR #1222, opened 2024-09-20
fix: landing pages syntax highlighting [internal]

webrdaniel:
reported here: https://apify.slack.com/archives/C0L33UM7Z/p1726736971578239

I've found out that it started with this commit https://github.com/apify/apify-docs/pull/1176

<s>The issue was with `package-lock`, so I've tried to use the previous `package-lock` in the current master and update it with `npm i`. Looks like it is fixed but there are quite a lot of changes, so I hope it didn't break anything. </s>

We decided to use the native Docosaurus CodeBlock component. It looks quite different than our component, so it needs some restyling to make it look more similar.


B4nan (changes requested):
reverting dependency updates like this is not really a fix, this will get broken again quite easily. if something, you should find what dependency caused this and pin its version instead of altering lockfile by hand. even better, find a way to make it work with the latest versions of our dependencies - that is the proper fix to me.

barjin:
The new changes are reusing the `CodeBlock` from the Docusaurus theme that has the code highlight working. The design is slightly different.

|before|after|
|---|---|
|obrazek|obrazek|


The reason for the broken highlight is still unknown; this also broke the `ReactSyntaxHighlighter` in the search modal preview panel (right side).

obrazek

webrdaniel:
@barjin Thanks for looking into it
@hanatsai Is it OK if we use the Docusaurus CodeBlock that has a different design (demonstrated in the "after" screenshot), to make the highlight work?

B4nan:
> Is it OK if we use the Docusaurus CodeBlock that has a different design (demonstrated in the "after" screenshot), to make the highlight work?

We agreed with Jindra that we will to make it look the same. But we have a better question - do we want all the other code blocks (outside of such landing pages, so in the actual docs content) to look the same as those? Since we will be restyling it to look the same on the landing page, we can as well do it globally for all the codeblocks.

hanatsai:
Hi, would be nice to make it look the same globally, if you are facing some codeblock that our design system doesn't support or doesn't have an explanation from [here](https://www.figma.com/design/duSsGnk84UMYav8mg8QNgR/Shared-library?node-id=2131-1006&t=Bxzekp8g0SqSTOTJ-1), please ask the design team or me, we'll update the design! Thank you :)

B4nan:
Ok good, I also thought it would be nice to have it unified (and in such case it makes a lot of sense to just restyle the native docusaurus component instead of using a custom one).

jbartadev:
@B4nan Cool, our team can take care of the unification of the native docusaurus component with the shared one if you need to - you have a lot of your agenda already ✌️

webrdaniel:
We will try to apply some styling to the native component to make it look more similar to ours in the next sprint.

webrdaniel:
Added themes that are close to the ones that we use in the console and on the web. (The light one is the default Prism theme, and it was missing in the `prism-react-renderer`, so I had to define it manually. The dark one is night-owl with an edited background color).
Also updated the styles of the code block to be similar to the one we have. Although there are still some differences, it's difficult to change those only by rewriting CSS.

HonzaTuron (approved):

B4nan (approved):

hanatsai (approved):
Looks perfect!!!!

jbartadev (approved):

apify-docs-theme/src/theme/custom.css
```diff
@@ -1562,8 +1562,29 @@ html[data-theme='dark'] .runnable-code-block svg .apify-logo {
...
+}
+
+.theme-code-block {
+    border-radius: 12px !important;
+    overflow: hidden;
+    border: 1px solid var(--color-neutral-border);
+    box-shadow: none !important;
+}
+
+.codeBlockTitle_node_modules-\@docusaurus-theme-classic-lib-theme-CodeBlock-Content-styles-module {
```
B4nan, line 1576: i wonder if this actually works on a production build, probably not

webrdaniel: Good catch. It's hard to select the classes but this should do the trick.

B4nan (approved):

PR #1229, opened 2024-09-26
docs: change milvusApiKey to milvusToken

jirispilka:


sources/platform/integrations/ai/milvus.md
```diff
@@ -22,7 +22,7 @@ It can also be connected to a managed Milvus instance on [Zilliz Cloud](https://
 
 Before you begin, ensure that you have the following:
 
-- A Milvus database URL and API token. Optionally, you can use a username and password. You can run Milvus on Docker or Kubernetes, but in this example, we'll use the hosted Milvus service at [Zilliz Cloud](https://cloud.zilliz.com).
+- A Milvus database URI and API token. Optionally, you can use a username and password in the URI. You can run Milvus on Docker or Kubernetes, but in this example, we'll use the hosted Milvus service at [Zilliz Cloud](https://cloud.zilliz.com).
```
TC-MO, line 25: Is URI common enough that it doesn't need to be first written out fully? (It does pass google test)

TC-MO (approved):
LGTM, one question but otherwise looks good

sources/platform/integrations/ai/milvus.md
jirispilka: I think it is known, but we can write it fully first, there is no harm in that

PR #1239, opened 2024-10-08
docs: rewrite runs section

TC-MO:
Restructed the content with headings & subheadings
Simplified language for clarity
Changed passive voice to active voice where appropriate

### DIFF
--- a/sources/platform/actors/development/builds_and_runs/runs.md
+++ b/sources/platform/actors/development/builds_and_runs/runs.md
@@ -27,35 +27,3 @@ The run receives input via the `INPUT` record of its default [key-value store](.
 ## Run duration and timeout
 
 Actor runs can be short or long-running. To prevent infinite runs, you can set a timeout. The timeout is specified in seconds, and the default timeout varies based on the template from which you create your Actor. If the run doesn't finish within the timeout, it's automatically stopped, and its status is set to `TIMED-OUT`.
-
-Depending on the template used for your Actor, the default timeout values may differ.
-
-| Template name | Default timeout value in seconds |
-|:---|:---|
-| Scrapy | 3600 |
-| Start with Python | 3600 |
-| BeautifulSoup | 3600 |
-| Playwright + Chrome | 3600 |
-| Selenium + Chrome | 3600 |
-| Empty Python project | 3600 |
-| Standby Python project | 3600 |
-| Crawlee + BeautifulSoup | 3600 |
-| Crawlee + Playwright + Chrome | 3600 |
-| Start with JavaScript | 360 |
-| Start with TypeScript | 360 |
-| Crawlee + Cheerio | 3600 |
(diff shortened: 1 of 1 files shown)


sources/platform/actors/development/builds_and_runs/runs.md
```diff
@@ -9,15 +9,53 @@ slug: /actors/development/builds-and-runs/runs
...
+- Using the [Scheduler](../../../schedules.md) provided by the Apify platform
+- By one of the available [integrations](../../../integrations/index.mdx)
+
+## Input and environment variables
+
+The run receives input via the `INPUT` record of its default [key-value store](../../../storage/key_value_store.md). Environment variables are also passed to the run. For more information about environment variables check the [Environment variables](../programming_interface/environment_variables.md) section.
+
+## Run duration and timeout
+
+Actor runs can be short or long-running. To prevent an indefinite runs, you can set a timeout. The timeout is specified in seconds and varies between Actor templates. If the run doesn't finish within the timeout, it's automatically stopped, and its status is set to `TIMED-OUT`.
```
fnesveda, line 29: I clarified it a bit:
```suggestion
Actor runs can be short or long-running. To prevent infinite runs, you can set a timeout. The timeout is specified in seconds, and the default timeout varies based on the template from which you create your Actor. If the run doesn't finish within the timeout, it's automatically stopped, and its status is set to `TIMED-OUT`.
```

sources/platform/actors/development/builds_and_runs/runs.md
```diff
@@ -9,15 +9,53 @@ slug: /actors/development/builds-and-runs/runs
...
+
+Depending on the template used for your Actor, the default timeout values may differ.
+
+| Template name | Default timeout value in seconds |
```
fnesveda, line 33: I wonder if this table is really necessary, I don't think it's important to have the timeouts per template listed here. It will also be hard to maintain, as we'll have to update it every time we add, update or rename a template, and people will forget the templates are listed in this place.

fnesveda (changes requested):
Looks good, just a few suggestions surrounding the timeout

sources/platform/actors/development/builds_and_runs/runs.md
TC-MO: Yeah on one hand I think it would make sense to somehow give this information within docs, but on the other this table seems extremely unwieldy, I'll try to tinker with some other approaches

TC-MO: maybe just something within prose like 

```
The default values for timeout can be:

* 3600 seconds
* 360 seconds

depending on the template you've used for your Actor
```

fnesveda: It doesn't necessarily have to be just these two values, it just so happened that all templates have either 360 or 3600 seconds. I would just skip mentioning the values altogether.

PR #1245, opened 2024-10-09
docs: Add open assistants integration

jirispilka:
The documentation has two parts: 
- Real-time search data for OpenAI Assistant
- Save data into OpenAI Vector Store and use it in the assistant

I know it's a bit dense, but I wanted to avoid splitting it into two separate pages

@TC-MO any idea where to add a full code example? 
What about provide a link to github repository? Gist?


TC-MO:
I think the best way even if it's huge would be to put the code example within the doc. As kind of a summary? If you are afraid that it super big maybe we can hide it in <details> tag?

sources/platform/integrations/ai/openai_assistants.md
```diff
@@ -0,0 +1,432 @@
...
+description: Learn how to integrate Apify with OpenAI Assistants to provide real-time search data and to save them into OpenAI Vector Store
+sidebar_position: 1
+slug: /integrations/openai-assistants
+---
+
+**Learn how to integrate Apify with OpenAI Assistants to provide real-time search data and to save them into OpenAI Vector Store.**
+
+---
+
+[OpenAI Assistants API](https://platform.openai.com/docs/assistants/overview) allows you to build your own AI applications such as such as chatbots, virtual assistants, and more.
```
TC-MO, line 13:
```suggestion
[OpenAI Assistants API](https://platform.openai.com/docs/assistants/overview) allows you to build your own AI applications such as chatbots, virtual assistants, and more.
```

TC-MO (approved):
Just one repetition that I found, otherwise looks good!

PR #1261, opened 2024-10-22
docs: add resourceType property to support storage picker

MFori:
Proposal of new input schema property `resourceType` (based on https://apify.slack.com/archives/C01VBUV81UZ/p1729081102142439?thread_ts=1728031818.871689&cid=C01VBUV81UZ) which will solve the needs of storages picker (https://github.com/apify/apify-core/issues/12871)

As I already verified, in ajv schema there would be two new top level properties `resourceProperty` and `resourceArrayProperty`, ajv can distinguish between these and `stringProperty` with `arrayProperty` based on the required `resourceType` property. The `editor` property is optional with default value of `resourcePicker`.
Input schema PR will come when this one is approved.

The description is tightly connected with storages even though later the `resourceType` should support other types of resources, but for now I think it's ok like this.

Preview of docs:

Screenshot 2024-10-22 at 19 47 38
Screenshot 2024-10-22 at 19 47 51


sources/platform/actors/development/actor_definition/input_schema/specification.md
```diff
@@ -409,3 +409,41 @@ Editor type `select` allows the user to pick items from a select, providing mult
 ```
 
 To correctly define options for multiselect, you need to define the `items` property and then provide values and (optionally) labels in `enum` and `enumTitles` properties.
+
+### Resource
```
jancurn, line 413: Why not "Resource type" ?

sources/platform/actors/development/actor_definition/input_schema/specification.md
```diff
@@ -409,3 +409,41 @@ Editor type `select` allows the user to pick items from a select, providing mult
...
+
+The Resource property enables Actor users to select different types of resources.
```
jancurn, line 415: I'd say that:

Resource type identifies what kind of object is referred to in the input field. For example, a the Key-value store resource type can be referred to using a string ID.

mtrunkat (changes requested):
I am missing info what is the returned value in a case of an array vs string. There is screenshot only for string. I'd also improve the wording there mentioning that the return value is the resource ID itself.

Otherwise good to go after the other comments are fixed.

MFori:
> I am missing info what is the returned value in a case of an array vs string. There is screenshot only for string. I'd also improve the wording there mentioning that the return value is the resource ID itself.
> 
> Otherwise good to go after the other comments are fixed.

I update it based on @jancurn comments and try to better mention that the return value is the resource ID:
I can also include screenshot for array input once we have it implemented.

Screenshot 2024-10-23 at 12 05 23

gippy (approved):
Just one question. Since these docs already mention, that we support arrays too. Are you planning to update the resource picker in UI to support arrays?

If not then we should not mention arrays until we actually decide that we want to implement them.

MFori:
> Just one question. Since these docs already mention, that we support arrays too. Are you planning to update the resource picker in UI to support arrays?
> 
> If not then we should not mention arrays until we actually decide that we want to implement them.

Yeah I have almost done it in UI too

jancurn:
I assume we don't allowed both string and array at the same time, that the input type can be just one of them. Right?

MFori:
> I assume we don't allowed both string and array at the same time, that the input type can be just one of them. Right?

Exactly, there is still the `type` property with value of `string` or `array`:

Screenshot 2024-10-24 at 8 17 34

mtrunkat (approved):

sources/platform/actors/development/actor_definition/input_schema/specification.md
```diff
@@ -409,3 +409,41 @@ Editor type `select` allows the user to pick items from a select, providing mult
 ```
 
 To correctly define options for multiselect, you need to define the `items` property and then provide values and (optionally) labels in `enum` and `enumTitles` properties.
+
+### Resource type
+
+Resource type identifies what kind of object is referred to in the input field. For example, the Key-value store resource type can be referred to using a string ID.
```
netmilk, line 415:
```suggestion
Resource type identifies what kind of Apify Platfrom object is referred to in the input field. For example, the Key-value store resource type can be referred to using a string ID.
```

sources/platform/actors/development/actor_definition/input_schema/specification.md
```diff
@@ -409,3 +409,41 @@ Editor type `select` allows the user to pick items from a select, providing mult
...
+
+![Apify Actor input schema dataset](./images/input-schema-dataset-json.png)
+
+Properties:
+
+| Property       | Value                                                                             | Required | Description                                                                        |
+|----------------|-----------------------------------------------------------------------------------|----------|------------------------------------------------------------------------------------|
+| `type`         | One of <ul><li>`string`</li><li>`array`</li></ul>                                 | Yes      | Specifies the type of input - string for single value or array for multiple values |
+| `editor`       | One of <ul><li>`resourcePicker`</li><li>`hidden`</li></ul>                        | No       | Visual editor used for <br/>the input field. Defaults to `resourcePicker`.         |
+| `resourceType` | One of <ul><li>`dataset`</li><li>`keyValueStore`</li><li>`requestQueue`</li></ul> | Yes      | Type of resource                                                                   |
```
netmilk, line 447:
```suggestion
| `resourceType` | One of <ul><li>`dataset`</li><li>`keyValueStore`</li><li>`requestQueue`</li></ul> | Yes      | Type of Apify Platform resource                                                                   |
```

netmilk (approved):
It looks all ok to me! 

One small thing to consider: adding context to the terms "object" and "resource" where possible makes it easier to understand they are related to Apify's architecture and not just a "resource" or "object" from the generic programming or integration/API domain language.

PR #1273, opened 2024-11-05
chore: add redirect for crawlee-python upgrading page

vdusek:


B4nan:
@fnesveda could you take a peek at this? we don't have any redirects for crawlee here just yet, so not sure if its fine like this

fnesveda (approved):
Looks good. Maybe I would move the `rewrite` after the `location`s to keep it consistent with the first `server` block, it should have no functional difference.

PR #1279, opened 2024-11-11
docs: restructure content and enhance model descriptions

TC-MO:
Restructured the content for better flow and clarity
Expanded the description of each pricing model
Added Pay per Event pricing model description
Formatted the content using Markdown syntax

PR #1281, opened 2024-11-11
docs: add info about PPE pricing model to Academy

TC-MO:
Added new section about PPE pricing model
Added screenshot reflecting UI in the Console for PPE pricing model Added tip about naming of events in the PPE Actors Added TODOs


mhamas (changes requested):
Before the monetization is available to all creators, i.e. all creators can set it up themselves, I wouldn't add it to docs, or at least I'd frame it in a way that if they want to have this feature enabled, they need to contact us. What do u think?

TC-MO:
Is there any concrete ETA for releasing it to all creators? I'm asking because I can see few solutions

1. If it is close ETA then add note about availability as you suggested and release docs with such note, to be removed once PPE is generally available
2. If it not a close ETA then remove mentions of PPE untli it is GA

TC-MO:
After further thinking I would propose to:

* Shelve this PR untli PPE is GA and then fill in the blanks regarding the payout & best practices
* Move forward with the https://github.com/apify/apify-docs/pull/1279 since the goal there is to just have all possible pricing models that user might encounter in one place (and we have working PPE Actors in Store)

Thoughts?

mhamas:
> Is there any concrete ETA for releasing it to all creators?
No exact ETA, but I think general availability is a couple of months away at this point.

Agree with your suggestion

TC-MO:
@mhamas I guess with all the restructuring and PPE already living within platform docs, this PR can be closed?

PR #1286, opened 2024-11-13
fix: space between header and flags added

HonzaKirchner:
CleanShot 2024-11-13 at 09 56 26@2x

### DIFF
--- a/apify-docs-theme/src/theme/custom.css
+++ b/apify-docs-theme/src/theme/custom.css
@@ -607,7 +607,7 @@ article .card h2 {
 }
 
 .tsd-flag-group {
-    margin-right: 0 .5rem;
+    margin: 0 .5rem;
 }
 
 .tsd-panel .tsd-flag {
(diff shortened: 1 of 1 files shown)


apify-docs-theme/src/theme/custom.css
```diff
@@ -606,6 +606,10 @@ article .card h2 {
     user-select: none;
 }
 
+.tsd-flag-group {
```
B4nan, line 609: this doesnt look right, we already have margin on the group, its missing on `tsd-flag`

image

image

for `tsd-flag` we want just `margin-right`, it has a negative top margin and you would override it this way

B4nan: hmm the screen above is parameters, but the one in OP is on higher level (method name). this will be probably more tricky, you will need to target this better most likely.

try to debug this directly on the SDK docs so you can see how things end up visually. you can run them separately.

HonzaKirchner: Ahh.. I see.
Should be fixed now ✅ 
CleanShot 2024-11-18 at 13 00 30@2x

B4nan (approved):
I still feel like we will also need the `tsd-flag` margin which we have in the other repos, but let's deal with that once we see it.

PR #1287, opened 2024-11-13
fix: ensure correct position of old anchors in API docs

B4nan:


TC-MO (approved):
Should there be a bigger audit to double check all API links?

janbuchar (approved):

PR #1294, opened 2024-11-18
chore: Add info about omitted user fields

novotnyj:
We remove these fields when the endpoint is accessed from the actor run. This is done to avoid leaking sensitive fields to public actors. It's not mentioned anywhere.


mhamas (approved):

TC-MO (approved):

TC-MO:
Shouldn't that update openapi repo not the apib?

TC-MO:
I think this PR can be closed without merging since we will be retiring `apib` soon and OpenAPI is supposed to be the default API docs

novotnyj:
Closing in favour of https://github.com/apify/openapi/pull/107

PR #1300, opened 2024-11-22
feat: better openapi docs

B4nan:
This PR overhauls our OpenAPI setup completely:
- new API docs are available on https://docs.apify.com/api/v2-new/apify-api
- use https://github.com/PaloAltoNetworks/docusaurus-openapi-docs instead of the redoc plugin
- embed the OpenAPI specs to this repository (`apify-api` folder)
	- both the specs yaml file and the MDX files are git ignored
	- the spec include (not only) corrections from https://github.com/apify/openapi/pull/108
- provide NPM scripts to work with the OpenAPI specs, namely
	- `npm start` will generate the specs as well as produce the MDX files out of them
	- `npm run api:rebuild` will rebuild the yaml spec file and recreate the MDX files
- update docusaurus to v3.6.3 and enable `faster` mode (rust based tooling)
- cache Run on Apify code hashes locally

Things to discuss:
- ~index page now only redirects, maybe we should move the content from the introduction section (so outside of OpenAPI specs)~
- ~folder structure~
- OpenAPI tests and schemathesis
- nested readmes in openapi folder


TC-MO:
I took a look and I think the easiest way will be to just exclude /source/api from mdlint, vale & typos

B4nan:
That's exactly what I am trying to do, it works for typos, still battling with the other two.

TC-MO (approved):

PR #1303, opened 2024-11-25
feat: lesson about using a framework

honzajavorek:
This PR introduces a new lesson to the Python course for beginners in scraping. The lesson is about working with a framework. Decisions I made:

- I opted not to use type hints to make the examples less cluttered and to avoid the need to explain type hints to people who didn't ever use them
- The logging section serves two purposes - first, it adds logging :) and second, it conveniently provides code for the whole program at the end of the lesson
- I had a hard time to come up with exercises, because most of the simple stuff I came up with was too simple and would result in shorter and simpler code without the framework 😅
    - I decided to have one classic scenario (listing & detail) just to let the student write their first Crawlee program. It's a bit challenging regarding traversal over the HTML to get the data, but it shouldn't be challenging regarding Crawlee.
    - I introduced one scenario where the scraper needs to jump through several pages (even domains) to get the result. Such program would be hard or at least very annoying to write without framework.
- As always, I focused on having the examples based on real world sites which are somewhat known and popular globally, but also don't feature extensive anti-scraping protections.

## Crawlee feedback

Regarding Crawlee, I didn't have much trouble to write this lesson, apart from the part where I wanted to provide hints on how to do this:

```py
requests = []
for ... in context.soup.select(...):
    ...
    requests.append(Request.from_url(imdb_search_url, label="..."))
await context.add_requests(requests)
```

I couldn't find good example in the docs, and I was afraid that even if I provided pointers to all the individual pieces, the student wouldn't be able to figure it out.

Also, I wanted to link to docs when pointing out the fact that `enqueue_links()` has a `limit` argument, but I couldn't find `enqueue_links()` in the docs. I found [this](https://crawlee.dev/python/api/class/EnqueueLinksFunction#Methods) which is weird. It's not clear what object is documented, or what it is, feels like some internals, not as regular docs of a method. I probably know how come it's this way, but I don't think it's useful this way and I decided I don't want to send people from the course to that page.

One more thing: I do think that Crawlee should log some "progress" information about requests made or - especially - items scraped. It's so weird to run the program and then just look at the program as if it hanged, waiting if something happens or not. E.g. Scrapy logs how many items per minute I scraped, which I personally find super useful.


sources/academy/webscraping/scraping_basics_python/12_framework.md
```diff
@@ -6,25 +6,587 @@ sidebar_position: 12
...
+
+From the two main open-source options for Python, [Scrapy](https://scrapy.org/) and [Crawlee](https://crawlee.dev/python/), we chose the latter—not just because we're the company financing its development.
+
+We genuinely believe beginners to scraping will like it more, since it allows to create a scraper with less code and less time spent reading docs. Scrapy's long history ensures it's battle-tested, but it also means its code relies on technologies that aren't really necessary today. Crawlee, on the other hand, builds on modern Python features like asyncio and type hints.
+
+:::
+
+## Installing Crawlee
+
+When starting with the Crawlee framework, we first need to decide which approach to downloading and parsing we prefer. We want the one based on BeautifulSoup, so let's install the `crawlee` package with the `beautifulsoup` extra specified in brackets. The framework has a lot of dependencies, so expect the installation to take a while.
```
vdusek, line 37: Maybe we can add link to BeautifulSoup here?

sources/academy/webscraping/scraping_basics_python/12_framework.md
```diff
@@ -6,25 +6,587 @@ sidebar_position: 12
...
+- **No error handling:** The scraper stops if it encounters issues. It should allow for skipping problematic products with warnings or retrying downloads when the website returns temporary errors.
+
+In this lesson, we'll tackle all the above issues while keeping the code concise thanks to a scraping framework.
+
+:::info Why Crawlee and not Scrapy
+
+From the two main open-source options for Python, [Scrapy](https://scrapy.org/) and [Crawlee](https://crawlee.dev/python/), we chose the latter—not just because we're the company financing its development.
+
+We genuinely believe beginners to scraping will like it more, since it allows to create a scraper with less code and less time spent reading docs. Scrapy's long history ensures it's battle-tested, but it also means its code relies on technologies that aren't really necessary today. Crawlee, on the other hand, builds on modern Python features like asyncio and type hints.
```
vdusek, line 31: I think this is fine, but if you would want more reasons you can check out [this PR](https://github.com/apify/crawlee-python/pull/772).

vdusek (approved):
I got your point for avoiding type hints. However, in the case of the handler:

```python
    @crawler.router.default_handler
    async def handle_listing(context):
        ...
```        

It leaves the reader without any possibility for code completions or static analysis when working with the `context` object.

In my opinion, type hints should be included here. We have been using them across all docs & examples.

Just a suggestion for you to reconsider, not a request.

Other than that, good job :slightly_smiling_face:, and the code seems to be working.

sources/academy/webscraping/scraping_basics_python/12_framework.md
honzajavorek: The reader should know about BS4 from the previous lessons which work with it extensively.

honzajavorek:
Thanks for the review! I see your point and I will indeed reconsider adding the type hint, at least for the context. It would be easier decision if the type name wasn't [28 characters long](https://crawlee.dev/python/api/class/BeautifulSoupCrawlingContext), but you're right about the benefits for people with editors like VS Code, where we could assume some level of automatic code completions.

sources/academy/webscraping/scraping_basics_python/12_framework.md
```diff
@@ -6,25 +6,587 @@ sidebar_position: 12
...
+    asyncio.run(main())
+```
+
+In the code, we do the following:
+
+1. We perform imports and specify an asynchronous `main()` function.
+1. Inside, we first create a crawler. The crawler objects control the scraping. This particular crawler is of the BeautifulSoup flavor.
+1. In the middle, we give the crawler a nested asynchronous function `handle_listing()`. Using a Python decorator (that line starting with `@`), we tell it to treat it as a default handler. Handlers take care of processing HTTP responses. This one finds the title of the page in `soup` and prints its text without whitespace.
+1. The function ends with running the crawler with the product listing URL. We await the crawler to finish its work.
+1. The last two lines ensure that if we run the file as a standalone program, Python's asynchronous machinery will run our `main()` function.
```
mnmkng, line 72:
```suggestion
1. We perform imports and specify an asynchronous `main()` function.
2. Inside, we first create a crawler. The crawler objects control the scraping. This particular crawler is of the BeautifulSoup flavor.
3. In the middle, we give the crawler a nested asynchronous function `handle_listing()`. Using a Python decorator (that line starting with `@`), we tell it to treat it as a default handler. Handlers take care of processing HTTP responses. This one finds the title of the page in `soup` and prints its text without whitespace.
4. The function ends with running the crawler with the product listing URL. We await the crawler to finish its work.
5. The last two lines ensure that if we run the file as a standalone program, Python's asynchronous machinery will run our `main()` function.
```

mnmkng (approved):

sources/academy/webscraping/scraping_basics_python/12_framework.md
honzajavorek: Ha! In most Markdown flavors, numbering ordered lists with 1. is a trick how to trigger automatic numbering. While rendered correctly 1. 2. 3. etc., in the source you don't have to think about the order, and it's easy to add, remove, or re-order the points without the necessity to re-number them.

However, I didn't verify Docusaurus supports this and I don't know what Apify's conventions are regarding this. @TC-MO, do you have an opinion on this?

TC-MO: Docusaurus supports it with no issues. I don't think we have specific conventions for this tbh, I think we have regular numbering because MDlint sometimes goes crazy.

honzajavorek:
@vdusek I added reconsidering the type hints to https://github.com/apify/apify-docs/issues/1319, thanks!

@TC-MO I think we could merge this now, but I'd appreciate if you could take a look at what I did, at least in https://github.com/apify/apify-docs/pull/1303/commits/f0c604103e10f7388d841b42dcbf207c736111e0. I randomly added words to Vale's spelling dictionary, but even then, I had to turn it off for one code block. According to my testing it seems that Vale isn't able to identify and skip the block as a code block if it has `title="newmain.py"` (which is a non-standard Markdown extension of the syntax, I suppose). Do we have a solution for that, or do I have to comment it out, like I did? Thanks!

I asked about the dictionary also here https://github.com/apify/apify-docs/pull/1345/files#r1922397733

TC-MO:
not sure about titling the codeblock and vale issues, never encountered it before, let me investigate, but for a quick fix I guess turning vale off & on after codeblock will do. If we find a proper solution it will be an easy change

honzajavorek:
Thanks! I'm surprised, because I think this isn't the first time it's used in the docs, but it's also possible that Vale works incrementally, so it errors only when I'm adding it, and ignores whatever is already there. Not sure.

TC-MO:
The way Vale is set up is , it doesn't check whole docuemntation each time it runs, it just checks the changed files, so that is why some other files may have had title in codeblock, but since they have not been changed in this PR it isn't caught. ALso this might be the result of Vale Spelling being added.

honzajavorek:
That's what I think as well.

PR #1305, opened 2024-11-27
chore: copy & paste friendly curl API example

netmilk:
image

- the commands are restructured in a way that the editable part is always at the end of the snippet (especially annoying in multi-line snippets)
- added the saving of the response of the Run to a `run.json `
- added convenience `jq` command to find the dataset in the Actor run response
- removed (imo) unnecessary quotes
- squished the per command line breaks to save lines

Please, try it yourself before merging:

```
# Prepare Actor input
echo '{ "searchStringsArray": ["Apify"] }' > input.json

# Run the Actor and retreive its default dataset id.
curl -X POST -d @input.json -H 'Content-Type: application/json' \
  -s -o >(tee run.json) \
  https://api.apify.com/v2/acts/compass~crawler-google-places/runs?token=<YOUR_API_TOKEN>

# Find the defaultDatasetId in the API response 
cat run.json | jq -r .data.defaultDatasetId

# And pass it instead of <DATASET_ID>
curl https://api.apify.com/v2/datasets/<DATASET>/items?token=<YOUR_API_TOKEN>
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -73,7 +73,7 @@
         },
         "apify-docs-theme": {
             "name": "@apify/docs-theme",
-            "version": "1.0.161",
+            "version": "1.0.162",
             "license": "ISC",
             "dependencies": {
                 "@apify/docs-search-modal": "^1.2.0",
--- a/src/pages/api/index.tsx
+++ b/src/pages/api/index.tsx
@@ -138,20 +138,18 @@ export default function Api() {
                     <ClientCodeWrapper>
                         <CodeBlock title="cURL" language='bash'>
                             {`# Prepare Actor input
-cat > input.json <<'EOF'
-{
- // Define the input in JSON here
-}
-EOF
+echo '{ "searchStringsArray": ["Apify"] }' > input.json
 
-# Run the Actor
-curl "https://api.apify.com/v2/acts/username~actorname/runs?token=<YOUR_API_TOKEN>" \\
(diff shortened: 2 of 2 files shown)


jancurn:
Thanks for looking into it. 

- Please can we make the code not depending on manually-entered `<DATASET_ID>` ? Basically so that you can run it and and see results right away
- I'd consider using an Actor that is not PPR, e.g. Website Content Crawler

netmilk:
> Please can we make the code not depending on manually-entered <DATASET_ID> ? Basically so that you can run it and and see results right away

Sure, we can do that, but I would consider the learning and communication value of the code example on the Docs API landing page. I thought about having the API key exported at the very top and the dataset ID extracted with jq into another constant, but it seemed harder to read compared to having the reader manually replace the placeholders and understand what they’re doing.

If we want a super simple one-liner `curl` example to demo Apify (API), I'd consider adding a convenience API endpoint that: 
- calls the actor
- waits for its start
- it streams the default dataset content. 

(if that's even possible)

netmilk:
@jancurn  What about now?  The [Actor](https://apify.com/compass/crawler-google-places) is "pay per usage" if I'm not mistaken.

image

jancurn:
@netmilk looks good cheers

janbuchar (approved):

TC-MO (approved):

PR #1307, opened 2024-11-27
docs: Update input schema specs

jancurn:


sources/platform/actors/development/actor_definition/input_schema/specification.md
```diff
@@ -11,20 +11,17 @@ slug: /actors/development/actor-definition/input-schema/specification/v1
...
 - It determines the user interface components rendered in the Apify Console for configuring the Actor's input.
 
-By defining an input schema, you can provide a user-friendly interface for configuring your Actor while ensuring that the input data supplied by users adheres to the specified requirements and constraints.
+By defining an input schema, you can provide a user-friendly interface for configuring your Actor while ensuring that the input data supplied by users adheres to the specified requirements and constraints. Also, the input schema make it easy to call and integrate your Actor from external systems.
 
-You can specify input schema for an Actor in multiple ways:
+To define an input schema for an Actor, set `input` field in the `.actor/actor.json` file to an input schema object as described below, or path to a JSON file containing the input schema.
+For backwards compability, if the `input` field is omitted, the system looks for an `INPUT_SCHEMA.json` file in the `.actor` directory or for an `INPUT_SCHEMA.json` file in the Actor's root directory - but do not depend on this behavior as it might be removed in the future.
```
TC-MO, line 22:
```suggestion
To define an input schema for an Actor, set `input` field in the `.actor/actor.json` file to an input schema object as described below, or path to a JSON file containing the input schema.

For backwards compability, if the `input` field is omitted, the system checks these locations:

  1. `INPUT_SCHEMA.json` in the `.actor` directory
  2. `INPUT_SCHEMA.json` in the Actor's root directory
 ```
 
 I would omit any information about what might or might not happen, if it does then we can update the documentation with a note. If we want to steer users in one direction we can add that some info about preferred or recommended way of doing things

**Review by TC-MO — CHANGES_REQUESTED** [review]

File: sources/platform/actors/development/actor_definition/input_schema/specification.md (hunk 1 above)
Comment (jancurn) [reply sources/platform/actors/development/actor_definition/input_schema/specification.md:22]: IMO it's quite normal in docs to say that some behavior might be deprecetated in the future, so that people don't use it - this is exactly the case

Comment (jancurn) [reply sources/platform/actors/development/actor_definition/input_schema/specification.md:22]: Really? I see it all the time in docs that some feature is deprecated and might be removed in the future. See https://www.google.com/search?q=deprecated+and+might+be+removed+in+the+future

Comment (TC-MO) [reply sources/platform/actors/development/actor_definition/input_schema/specification.md:22]: But none of the results are from official documentation but rather community resources. 

Furthermore
`deprecated & might be removed in the future` =/= `might be removed in the future`,
if the intention is that these two ways of defining input schema are deprecated and & will be removed in the future, then let's mark them explicitly as deprecated and then mentioning that they might be removed in the future is a-ok

Comment (jancurn) [reply sources/platform/actors/development/actor_definition/input_schema/specification.md:22]: Oh right, fair enough, I'll mention it's deprecated

**Review by TC-MO — APPROVED** [review]

File: sources/platform/actors/development/actor_definition/input_schema/specification.md [hunk 2]
```diff
@@ -11,20 +11,17 @@ slug: /actors/development/actor-definition/input-schema/specification/v1
...
-- It determines the user interface components rendered in the Apify Console for configuring the Actor's input.
+- It ensures the input data supplied to the Actor adhere specified requirements and validation rules.
+- It is used by the Apify platform to generate a user-friendly interface for configuring and running your Actor.
+- It simplifies invoking your Actors from external systems by generating calling code and connectors for integrations.
 
-By defining an input schema, you can provide a user-friendly interface for configuring your Actor while ensuring that the input data supplied by users adheres to the specified requirements and constraints.
+To define an input schema for an Actor, set `input` field in the `.actor/actor.json` file to an input schema object (described below), or path to a JSON file containing the input schema object.
+For backwards compability, if the `input` field is omitted, the system looks for an `INPUT_SCHEMA.json` file either in the `.actor` directory or the Actor's top-level directory - but note that this functionality is deprececated and might be removed in the future. The maximum allowed size for the input schema file is 500 kB.
```
TC-MO, line 21: typo fix

```suggestion
For backwards compatibility, if the `input` field is omitted, the system looks for an `INPUT_SCHEMA.json` file either in the `.actor` directory or the Actor's top-level directory - but note that this functionality is deprececated and might be removed in the future. The maximum allowed size for the input schema file is 500 kB. 
```

PR #1313, opened 2024-11-29
fix: Fixing the broken YAML multi-string in opanapi

netmilk:
This was caused by the manual edits after the migration, improper use of YAML folded block scalar, and fixed the same way as https://github.com/apify/openapi/pull/111

**After:**
image

**Before**
image (9)

### DIFF
--- a/apify-api/openapi/paths/actors/acts@{actorId}@runs@last.yaml
+++ b/apify-api/openapi/paths/actors/acts@{actorId}@runs@last.yaml
@@ -3,117 +3,65 @@ get:
     - Actors/Last run object and its storages
   summary: Get last run
   description: |
-    This is not a single endpoint, but an entire group of endpoints that lets
-    you to
-
-    retrieve and manage the last run of given Actor or any of its default
-    storages.
-
+    This is not a single endpoint, but an entire group of endpoints that lets you to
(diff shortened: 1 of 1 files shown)


apify-api/openapi/paths/actors/acts@{actorId}@runs@last.yaml
```diff
@@ -3,117 +3,65 @@ get:
...
 
-
     Additionally, each of the above API endpoints supports all sub-endpoints
-
     of the original one:
 
-
     #### Key-value store
 
+    * `/v2/acts/{actorId}/runs/last/key-value-store/keys{?token,status}` [Key collection](#/reference/key-value-stores/key-collection)
```
B4nan, line 42: we dont need this blank line either, same with the other lists

not sure it it is rendered as two separate lists with it?

B4nan (approved):

apify-api/openapi/paths/actors/acts@{actorId}@runs@last.yaml
netmilk: Well, you were right. The Markdown parser in your head is quite admirable. 🫡

Addressed, and now it looks like this:

image

PR #1317, opened 2024-12-04
feat: second lesson about DevTools

honzajavorek:
- [ ] text
- [ ] exercises
- [ ] proofreading


sources/academy/webscraping/scraping_basics_python/02_devtools_locating_elements.md
```diff
@@ -6,12 +6,210 @@ sidebar_position: 2
...
+
+It will return the HTML element for the first product card in the listing:
+
+![Using querySelector() in DevTools Console](./images/devtools-queryselector.png)
+
+:::note About the missing semicolon
+
+In the screenshot, there is a missing semicolon `;` at the end of the line. In JavaScript, semicolons are optional, so it doesn't make a difference here.
+
+:::
```
mnmkng, line 75: Since semicolons are optional in Python as well, wouldn't it be easier to just delete the semicolon from the code example - to align it with the screenshot? Rather than writing this long explanation? 😅

mnmkng (approved):

vdusek (approved):

sources/academy/webscraping/scraping_basics_python/02_devtools_locating_elements.md
honzajavorek: The issue is that all the other JS examples do have the semicolon, so it's inconsistency and the reader might wonder why suddenly the semicolon isn't there. The most correct solution would be to create a new screenshot with the semicolon in place, so that there's nothing to think about, and then the note can be removed. But this was faster for now. I can file it as a low-priority improvement.

honzajavorek:
@TC-MO Do you want to take a look at this, or are we good to go?

TC-MO (approved):

sources/academy/webscraping/scraping_basics_python/02_devtools_locating_elements.md
honzajavorek: Filed as https://github.com/apify/apify-docs/issues/1417

honzajavorek:
Thanks everyone!

PR #1318, opened 2024-12-04
feat: add code samples for actor collection

m-murasovs:
Here are the first two code samples for the API docs. These are for the JS client and are a tentative proposal. When we've agreed on how they should look, I'll also add the ones for the Python client.

**Note**: The Palo Alto plugin currently doesn't support TypeScript and there's no straightforward way to add multiple custom code samples for the same language (I thought we could have JS and TS client samples under the JavaScript bit).
I have created an issue asking about this on the plugin's repo: https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/issues/1035

**Note 2**: For the POST request, as mentioned by @netmilk , it would be nice to have the request body present, but that would be complicated, since the schema isn't available in the examples, and would make the testing harder. I think it's not so bad, though, since the request body is shown right under the code example, so it's quite easy to find.

So, on the new docs it will look like this:
CleanShot 2024-12-04 at 16 06 42@2x

And in the old/current docs like this: (I had to fix the background color)
CleanShot 2024-12-04 at 16 27 02@2x


janbuchar (approved):
If it works, it is perfect :slightly_smiling_face:

PR #1321, opened 2024-12-06
feat: third lesson about DevTools

honzajavorek:
Porting [this lesson](https://docs.apify.com/academy/web-scraping-for-beginners/data-extraction/using-devtools) and [this lesson](https://docs.apify.com/academy/web-scraping-for-beginners/data-extraction/devtools-continued):

- I deliberately didn't include anything about cleaning data or iterating over the data, because it would be unnecessary load of JavaScript for a Python dev, for no obvious reason. They'll learn everything in the following lessons. I focused only on stuff which can be considered as ad-hoc inspection of the website, which is useful even to Python devs.
- I tried to unify the style with the rest of the course. I wrote the stuff with my own words, and added standard polishing I usually do (dictionary, proofreading by AI, etc.)
- I added several screenshots.
- I added exercises.
- There's no Python, so I'm not bothering Vláďa with this one.

image


honzajavorek:
In https://github.com/apify/apify-docs/pull/1321/commits/94b26f5e75b43fc75a0050a80319c1511392c3e0 I added `ul` to the dictionary, because Vale wasn't happy otherwise, but I think we should set it up so that it's able to skip inline code blocks. It didn't have problems with any other inline code, so I'm not sure what's the problem - perhaps it's even a bug in Vale we should report back?

Screenshot 2025-01-21 at 15 01 49

mnmkng:
Hey guys, I'm sorry, but I find it really hard to find the time for the reviews. If @metalwarrior665 can't do them reasonably fast, then we need to find someone else to chip in.

Perhaps @souravjain540?

It's a shame that this great content gets stuck here for so long.

metalwarrior665:
I didn't know we were even blocking these PRs on reviews. I have done some work in the JS space recently and I would definitely want to review any change there. For Python I think @souravjain540 would make sense

honzajavorek:
As of now I'm in no rush with this one, I got stuck on other stuff.

TC-MO:
> In [94b26f5](https://github.com/apify/apify-docs/commit/94b26f5e75b43fc75a0050a80319c1511392c3e0) I added `ul` to the dictionary, because Vale wasn't happy otherwise, but I think we should set it up so that it's able to skip inline code blocks. It didn't have problems with any other inline code, so I'm not sure what's the problem - perhaps it's even a bug in Vale we should report back?
> Screenshot 2025-01-21 at 15 01 49

I could've sworn that it was already set up like this. But it also should omit frontmatter and in one document it insisted to check frontmatter and throw error so ¯\\\_(ツ)_/¯

sources/academy/webscraping/scraping_basics_python/03_devtools_extracting_data.md
```diff
@@ -6,12 +6,131 @@ sidebar_position: 3
...
+
+![Extracting product title](./images/devtools-extracting-title.png)
+
+## Extracting price
+
+To figure out how to get the price, we'll use the **Elements** tab of DevTools again. We notice there are two prices, a regular price and a sale price. For the purposes of watching prices we'll need the sale price. Both are `span` elements with the `price` class.
+
+![Finding child elements](./images/devtools-product-details.png)
+
+We could either rely on the fact that the sale price is likely to be always the one which is highlighted, or that it's always the first price. For now we'll rely on the later and we'll let `querySelector()` to simply return the first result:
```
TC-MO, line 61:
```suggestion
We could either rely on the fact that the sale price is likely to be always the one which is highlighted, or that it's always the first price. For now we'll rely on the latter and we'll let `querySelector()` to simply return the first result:
```

sources/academy/webscraping/scraping_basics_python/03_devtools_extracting_data.md
```diff
@@ -6,12 +6,131 @@ sidebar_position: 3
...
+
+```js
+price = subwoofer.querySelector('.price');
+price.textContent;
+```
+
+It works, but the price isn't alone in the result. Before we'd use such data, we'd need to do some **data cleaning**:
```
TC-MO, line 68: This sentence `Before we'd...` somehow feels unnatural, perhaps something like this? Also within docs I  try to keep bold text only to indicate UI elements and cursive as emphasis. And it seems like here bold is used as both emphasis & to highlight UI elements, isn't that a bit confusing?

```suggestion
It works, but the price isn't alone in the result. Before using such data, we'd need to do some _data cleaning_:
```

TC-MO (approved):
One typo and one suggestion otherwise looks good from my point of view

sources/academy/webscraping/scraping_basics_python/03_devtools_extracting_data.md
honzajavorek: Thanks! I'm trying to keep bold for UI elements, so this is my mistake. Did you see it elsewhere or just here?

honzajavorek: I scanned through the text and I think I see it only here. Phew! Thank you.

TC-MO: In this lesson only here :D

PR #1322, opened 2024-12-06
fix(request-queue): Update request queue locking docs

drobnikj:
I made some changes in docs regarding the API changes.

* stress out that locking works on the client level and the same as on the run level.
* update code example by separating code into two actors using code tabs

### DIFF
--- a/sources/platform/storage/request_queue.md
+++ b/sources/platform/storage/request_queue.md
@@ -435,12 +435,12 @@ const client = new ApifyClient({
 const requestQueue = await client.requestQueues().getOrCreate('example-queue');
 
 // Creates two clients with different keys for the same request queue.
-const requestQueueClientOne = client.requestQueue(requestQueue.id, {
+const requestQueueClient = client.requestQueue(requestQueue.id, {
     clientKey: 'requestqueueone',
 });
 
 // Adds multiple requests to the queue.
-await requestQueueClientOne.batchAddRequests([
+await requestQueueClient.batchAddRequests([
     {
         url: 'http://example.com/foo',
         uniqueKey: 'http://example.com/foo',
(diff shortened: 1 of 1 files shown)


sources/platform/storage/request_queue.md
```diff
@@ -457,23 +467,71 @@ await requestQueueClientOne.batchAddRequests([
...
     {
         limit: 2,
-        lockSecs: 60,
+        lockSecs: 120,
     },
 );
 
+// Checks when the lock will expire. The locked request will have a lockExpiresAt attribute.
+const theFirstRequestLockedByClientOne = processingRequestsClientOne.items[0];
+const requestLockedByClientOne = await requestQueueClientOne.getRequest(
```
fnesveda, line 476: The naming of these two variables is pretty confusing, I don't really know what's the difference between `theFirstRequestLockedByClientOne` and `requestLockedByClientOne`, especially by name. Could you come up with some better names?

sources/platform/storage/request_queue.md
```diff
@@ -457,23 +467,71 @@ await requestQueueClientOne.batchAddRequests([
...
+        limit: 10,
         lockSecs: 60,
     },
 );
-
-// Checks when the lock will expire. The locked request will have a lockExpiresAt attribute.
-const theFirstRequestLockedByClientOne = processingRequestsClientOne.items[0];
-const requestLockedByClientOne = await requestQueueClientOne.getRequest(
-    theFirstRequestLockedByClientOne.id,
+const wasTheClientTwoLockedSameRequest = !!processingRequestsClientTwo.items.find(
```
fnesveda, line 530: Here I'm also not sure what the varaible means, could you rename it, or explain better in the comment?

fnesveda (changes requested):
Looks good, but the variable naming is a bit confusing, could you improve it?

fnesveda (approved):

TC-MO (changes requested):

TC-MO (approved):

PR #1331, opened 2024-12-09
docs: fix Actor capitalization throughout open api docs

TC-MO:


apify-api/openapi/openapi.yaml
```diff
@@ -45,8 +45,8 @@ info:
 
     To use your token in a request, either:
 
-    - Add the token to your request's `Authorization` header as `Bearer
-    <token>`.
+    - Add the token to your request's `Authorization` header as <code>Bearer
+    &lt;token&gt;</code>.
```
netmilk, line 49: Is this written by hand? Are you aware of the HTML entities and the HTML tags instead of Markdown backticks?

apify-api/openapi/paths/actors/acts@{actorId}@builds.yaml
```diff
@@ -75,8 +75,8 @@ get:
   x-py-doc-url: https://docs.apify.com/api/client/python/reference/class/BuildCollectionClientAsync#list
 post:
   tags:
-    - Actors/Build collection
-  summary: Build actor
+    - Builds
```
netmilk, line 78: Are you concisely changing the tag here? It might result in a broken link.

netmilk (changes requested):

apify-api/openapi/openapi.yaml
TC-MO: yes, because for some reason regular backticks are not working in new API docs check the screenshot from current version

it looks fine on Redoc version
image


but this is how it looks currently on the new API docs
image


I tested this locally and it changed to backticks locally so I hope the same will happen on prod

apify-api/openapi/paths/actors/acts@{actorId}@builds.yaml
TC-MO: good catch, that is remnant of some testing, no intention to change any tags, will fix it

apify-api/openapi/openapi.yaml
B4nan: I'll have a look, next time ping me with such issues instead of going with some workaround, I hope we can find a way to make this work without any changes, since the workaround is ugly at best :]

TC-MO: hmm but on redoc API it is rederered correctly, the issue was in new API docs?

B4nan: No, its their quirk, it makes no sense to escape anything inside code blocks. I think they just solve things on wrong level.

B4nan (approved):

netmilk (approved):

PR #1336, opened 2024-12-10
feat: add code samples for Actor object

m-murasovs:
More code samples for the API docs. Please merge after approving 🙏


netmilk:
I feel like we should approach the substitutions the same way the other generated examples do that. 

The users won't read the comment line and the parseable example won't throw an exception, but it will result in non-`2xx`. A placeholder e.g. `<ACTOR ID>` would throw. Other approach might be using obvious placeholders e.g. `johndoe/foo-bar-actor`.

Also, please try to prevent horizontal scrolling wherever possible.
e.g.
```
const { items } = await apifyClient
  .actors()
  .list();
```


image

netmilk (approved):

PR #1337, opened 2024-12-10
docs: update monetize your Actor section

TC-MO:
rework structure
rephrase for better understanding & conciseness
add details tag to handle examples
add new admonitions to highlight important information


metalwarrior665 (approved):
Lot of these things now live in docs, right? So it makes sense to mostly remove stuff. We should try to honor the split of docs = what, academy = how. 

This will need another rewrite once PPE is officially released. We will probably deprecate PPR and discourage rental. And then we have to agree where we will list the best practices. If there is enough in docs and SDK handles lot, then maybe we don't need any tutorial.

sources/academy/platform/get_most_of_actors/monetizing_your_actor.md
```diff
@@ -5,169 +5,130 @@ sidebar_position: 5
...
-| Key-value store - lists         | **$0.05** / 1,000 lists    |
-| Request queue - timed storage   | **$4.00** / 1,000 GB-hours |
-| Request queue - reads           | **$0.004** / 1,000 reads   |
-| Request queue - writes          | **$0.02** / 1,000 writes   |
+| Key-value store - reads | **$0.005** / 1,000 reads |
+| Key-value store - writes | **$0.05** / 1,000 writes |
+| Key-value store - lists | **$0.05** / 1,000 lists |
+| Request queue - timed storage | **$4.00** / 1,000 GB-hours |
+| Request queue - reads | **$0.004** / 1,000 reads |
+| Request queue - writes | **$0.02** / 1,000 writes |
```
mhamas, line 57: Nit: I find this harder to read in source code

mhamas (approved):

sources/academy/platform/get_most_of_actors/monetizing_your_actor.md
TC-MO: I added more spacing to make it more readable in the source

PR #1339, opened 2024-12-10
docs: rewrite SEO and promotion document

TC-MO:
new structure
added details & summary tags for examples
added admonitions
rewrites for clarity & conciseness

### DIFF
--- a/sources/academy/platform/get_most_of_actors/seo_and_promotion.md
+++ b/sources/academy/platform/get_most_of_actors/seo_and_promotion.md
@@ -5,121 +5,105 @@ sidebar_position: 4
 slug: /get-most-of-actors/seo-and-promotion
 ---
 
-# SEO and promotion
-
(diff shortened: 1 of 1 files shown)


metalwarrior665 (changes requested):
Thanks. We have now pretty [sophisticated marketing playbook](https://apify.notion.site/3fdc9fd4c8164649a2024c9ca7a2d0da?v=6d262c0b026d49bfa45771cd71f8c9ab) so I would link it as the top section and keep the rest as short as possible.

TC-MO:
I don't know why I assumed that it was internal document, not viewable outside of org (and we knows what happens when one assumes something...) Maybe instead of keeping this whole section we could make it so that when you click on it in the sidebar it takes you to the marketing playbook?

metalwarrior665:
That sounds good if you are able to do that :)

TC-MO:
So you can checkout the results of adding marketing playbook [here](https://github.com/apify/apify-docs/pull/1341) PR's that are related to this will be closed off, with the exception of monetization guide

PR #1344, opened 2024-12-11
fix: openapi docs improvements

B4nan:
- unify sidebar and content badges for method type (post/put/create/delete/...)
- render category index pages
- hide next/prev buttons from openapi docs pages
- remove descriptions from generated category index of openapi docs
- add missing syntax highlighting and use custom language order and labels 
- enable docusaurus 3.6 perf optimizations expect SWC loader (not working with styled components)


TC-MO (commented):
LGTM, just one nit, shouldn't we go with proper names of languages instead of making them lowercase? ie. JavaScript not javascript, PHP not php , etc. in the right side component?

image

TC-MO (approved):
LGTM

PR #1345, opened 2024-12-11
docs: add spell check capabilities to Vale linter

TC-MO:


B4nan (approved):
As long as the CI is happy, I am happy 🙃

TC-MO:
@fnesveda , @mtrunkat This should help with typos throughout PR's. While our previous tool started strong (catching over 100 typos), it eventually started missing some pretty obvious ones. 

Vale is much more reliable ,with much bigger dictionary,  and will actually block merges until typos are fixed. Plus, we can make it even smarter by adding our own regex-based rules to `accept.txt` & `reject.txt`.

.github/styles/config/vocabularies/Docs/accept.txt
```diff
@@ -1,9 +1,88 @@
...
+serverless
+[Bb]oolean
+node_modules
+[Rr]egex
+[Mm]onorepo
+[Gg]ist
+SDK
+Dockerfile
+Docker's
+
```
honzajavorek, line 23: Is there any special meaning to these empty lines? I mean, if I'm about to add to this list, should I keep some sections apart, should I sort the words alphabetically, or what's the system?

TC-MO: Since this is normal `.txt` file there is no way to create any sort of heading/title for section, so the spaces are used to group words into some kind of topics.

PR #1356, opened 2024-12-13
docs: Add Python variant example of catching dataset validation errors

Pijukatel:
Add Python variant example of catching dataset validation errors.

Documents this change: https://github.com/apify/apify-client-python/pull/314


TC-MO (approved):
LGTM

PR #1360, opened 2024-12-18
feat(api): dataset fields statistics

MFori:
Solves https://github.com/apify/apify-core/issues/18807 - proposal for new API endpoint `/v2/datasets/{datasetId}/field-statistics` which should return [dataset field statistics](https://docs.apify.com/platform/actors/development/actor-definition/dataset-schema/validation#dataset-field-statistics)


apify-api/openapi/components/schemas/datasets/GetDatasetFieldStatisticsResponse.yaml
```diff
@@ -0,0 +1,21 @@
...
+    type: object
+    required:
+      - fields
+      - statistics
+    properties:
+      fields:
+        type: array
+        items:
+          type: string
+        description: 'Keys of the fields for which the statistics are provided.'
```
mtrunkat, line 16: Is this the list of all fields from `dataset.fields` in DB or really the list of fields for which we have statistics? If the latter one then the question is if we need to be returning this redundant information as

```
response.data.fields === Object.keys(response.data.statistics)
```

MFori: It's the list of all fields defined in `dataset.schema.fields`. 

I was thinking the same but because we store it separately and in docs there is this sentence: `When you configure the dataset fields schema, we generate a field list and measure the following statistics`, I though it has some reason.

But from implementation it looks like there will be always all fields in `statistics`. (If some field defined in dataset schema is never in dataset itself, it will have `emptyCount=number_of_items`).

So I assume you are right and it's redundant.

mtrunkat: In such case, I'd remove it; we can always add new properties later, but we can't ever remove them as that would break existing integrations.

MFori: I already updated it, but now I wonder, whether it would be better like this having the fields right under `data` object. E.g. `{ "data": { "name": { "emptyCount": 100 } } }`

 Or keep the `data.statistics` object having the place to add other things under `data` in the future 🤔 such as
 `{ "data": { "statistics": { "name": { "emptyCount": 100 } } } }`

mtrunkat: @netmilk and @fnesveda , what do you think?

```json
{
   "data":{
      "statistics":{
         "someValue":{
            "emptyCount":100
         },
         "anotherValue":{
            "min":100,
            "max":200,
            "emptyCount":0
         }
      }
   }
}
```

vs simply

```json
{
   "data":{
      "someValue":{
         "emptyCount":100
      },
      "anotherValue":{
         "min":100,
         "max":200,
         "emptyCount":0
      }
   }
}
```

The former one is extensible and the latter one is simpler.

fnesveda: How about this?
```
/v2/datasets/{datasetId}/statistics
```
```
{
   "data":{
      "fieldStatistics":{
         "someValue":{
            "emptyCount":100
         },
         "anotherValue":{
            "min":100,
            "max":200,
            "emptyCount":0
         }
      }
   }
}
```
That way, if we want to add more statistics about datasets later on, we can do it in the same endpoint.

MFori: In the ticket https://github.com/apify/apify-core/issues/18807 was note about this:

> Basically we want the API to return the output of the data from the dataset statistics collection.
And the endpoint could potentially be /\<datasetId>/stats or /\<datasetId>/validation-statistics if it's the first one then we might want to add also the normal dataset statistics there, so that might be confusing...

netmilk: I prefer @fnesveda's proposal: The uri `.../statistics` + `fieldStatistics` objects under `fieldStatistics` property, especially if you foresee additional types of statistics in the future returned in the response. 

Naming the key just `statistics` doesn't provide any additional semantic meaning, it would now introduce just an additional nesting and it would lead to an overload of the term in the future.

MFori: Ok, I used @fnesveda's approach, please take a look @mtrunkat @netmilk

mtrunkat (approved):

apify-api/openapi/paths/datasets/datasets@{datasetId}@statistics.yaml
```diff
@@ -0,0 +1,49 @@
...
+        example: WkzbQMuFYuamGv3YF
+    -   name: token
+        in: query
+        description: |
+            API authentication token. It is required only when using the `username~dataset-name` format for `datasetId`.
+        style: form
+        explode: true
+        schema:
+            type: string
+            example: soSkq9ekdmfOslopH
```
MFori, line 27: @mtrunkat @netmilk - @drobnikj suggest in [this PR](https://github.com/apify/apify-core/pull/18862#discussion_r1899409785) to require the authentication always in this endpoint. Is it right?

> This is questionable as we plan to remove anonymous access for storage. We can do it for new endpoints already.
I would make this isAuthenticationRequired: true, and restrict this endpoint.

```suggestion
```

mtrunkat: Good point! Let's make it authenticated = less work in the next months. As we will turn all endpoints this way.

PR #1364, opened 2024-12-19
fix: Add redirect from crawlee.dev/python/docs

janbuchar:
https://crawlee.dev/python/docs now leads to a 404 page, which is not optimal. This should fix that, in a manner consistent with the JS version of the docs


B4nan (approved):

janbuchar:
@B4nan is there anyone else who should see this or do I just merge it and it will get auto-deployed?

B4nan:
I don't think so, let's hope it will just work, the deployment should be automated (it was for the docs itself, not entirely sure if it applies to crawlee docs too).

PR #1367, opened 2024-12-20
feat: auto-trigger code sample testing

m-murasovs:


.github/workflows/test-code-samples.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Get current branch name
+        id: get-branch-name
+        run: echo "branch_name=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_ENV
+
+      - name: Call Code Sample Tester Actor
```
m-murasovs, line 23: I couldn't figure out how to fail the workflow if the actor run fails

.github/workflows/test-code-samples.yaml
```diff
@@ -0,0 +1,31 @@
...
+        env:
+          APIFY_TOKEN: ${{ secrets.CODE_SAMPLE_TESTER_ACTOR_TOKEN }}
```
m-murasovs, line 25: We'll need to add a token for this once the actor is moved to the Apify account

.github/workflows/test-code-samples.yaml
```diff
@@ -0,0 +1,31 @@
...
+          BRANCH_NAME: ${{ env.branch_name }}
+        run: |
+          curl --fail-with-body -X POST "https://api.apify.com/v2/acts/X8D0L4wfpO8fCL1uL/run-sync" \
+          -H "Content-Type: application/json" \
+          -H "Authorization: Bearer $APIFY_TOKEN" \
+          -d '{ "branch": "'"$BRANCH_NAME"'" }'
```
janbuchar, line 31: Will this work on PRs from forks?

.github/workflows/test-code-samples.yaml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  run-code-sample-tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Get current branch name
+        id: get-branch-name
```
janbuchar, line 20: That shouldn't be necessary, right?

.github/workflows/test-code-samples.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  run-code-sample-tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Get current branch name
```
janbuchar, line 19: Also the branch should be accessible under `{{ github.ref_name }}`. Then you shouldn't even need to call `actions/checkout`.

.github/workflows/test-code-samples.yaml
m-murasovs: This PR is from a fork, so I think so

CleanShot 2024-12-20 at 11 38 15@2x

janbuchar: The run in the console suggests otherwise 
image

m-murasovs: Great catch! I've modified the action and the actor so that the owner of the triggering repo is passed

m-murasovs: Here's the last run with the test: https://console.apify.com/actors/X8D0L4wfpO8fCL1uL/runs/g4B6nhuQZVKlOc9dx#log

m-murasovs: Also, sorry I deleted that comment because I realised it was wrong, but hadn't seen your second one yet :)

janbuchar (approved):

PR #1371, opened 2025-01-02
fix: various OpenAPI plugin fixes

B4nan:
- fix HTML comments in descriptions
- group sidebar based on group tags
- enable redirects for redoc links (`#tag...`)
- scroll sidebar after redirect
- enable redirects for sidebar categories
- fix wrongly escaped inline code block
- redirect to search page if no match found


TC-MO (approved):
LGTM

PR #1372, opened 2025-01-03
feat: add Redoc legacy urls and a script for traversing and safe dumping the yamls

netmilk:


apify-api/openapi/components/tags.yaml
```diff
@@ -170,23 +185,27 @@
   x-trait: 'true'
   description: '**[DEPRECATED]** API endpoints related to run of the actor were moved
     under new namespace [`actor-runs`](#/reference/actor-runs).'
+  x-legacy-doc-urls: https://docs.apify.com/api/v2#tag/ActorsAbort-run
```
B4nan, line 188: this looks like the value is a string, while it should be an array

B4nan (commented):
also, should we get rid of the `https://docs.apify.com/api/v2` prefix everywhere? the redirection script I build doesn't need it, and I can imagine the decorators also don't, looking at their code, they just cut the part before the `#` sign

apify-api/scripts/add-redoc-legacy-urls-to-tags.py
```diff
@@ -0,0 +1,48 @@
...
+from slugify import slugify
+from ruamel.yaml.scalarstring import LiteralScalarString
+import re
+from functools import reduce
+
+yaml = ruamel.yaml.YAML(typ='rt')
+yaml.indent(mapping=2, sequence=2, offset=0)
+
+# https://stackoverflow.com/questions/78075152/preserve-line-wrapping-in-ruamel-yaml
+# needed to convert flow sclars (>-) to literal blocks (|)
```
B4nan, line 13:
```suggestion
# needed to convert flow scalars (>-) to literal blocks (|)
```

also since this is commented out, i guess we don't need it?

apify-api/scripts/add-redoc-legacy-urls-to-tags.py
```diff
@@ -0,0 +1,48 @@
...
+      except KeyError:
+          tag['x-legacy-doc-urls'] = [redoc_url]
+    
+    with open(tags_file_path, 'w') as file:
+        yaml.dump(tags, file)
+
+
+
+
+
```
B4nan, line 48:
```suggestion

```

apify-api/scripts/add-redoc-legacy-urls-to-tags.py
```diff
@@ -0,0 +1,48 @@
...
+#yaml.representer.add_representer(type(None), my_represent_none)
+
+
+# Load tags
+tags_file_path = '../openapi/components/tags.yaml'
+with open(tags_file_path, 'r') as file:
+    tags = yaml.load(file)
+    for tag in tags:
+      print(tag)
+      safe_tag = re.sub(r'\/', "", tag['name'])
```
B4nan, line 33: not sure how this works in python, i would expect a global modifier, since there could be multiple slashes. but looking at our tags, we probably always have only one (or none), so i guess its fine this way too

apify-api/scripts/add-redoc-legacy-urls-to-tags.py
```diff
@@ -0,0 +1,48 @@
...
+
+# https://stackoverflow.com/questions/42094599/preserving-quotes-in-ruamel-yaml
+yaml.preserve_quotes = True
+
+# non empty null
+# https://stackoverflow.com/questions/57172997/yaml-dumper-outputs-blank-instead-of-null-from-none-value-how-to-make-it-output
+#def my_represent_none(self, data):
+#    return self.represent_scalar(u'tag:yaml.org,2002:null', u'null')
+#
+#yaml.representer.add_representer(type(None), my_represent_none)
```
B4nan, line 24: same here, was it needed when its commented out?

apify-api/scripts/add-redoc-legacy-urls-to-operations.py
```diff
@@ -0,0 +1,67 @@
...
+from slugify import slugify
+from ruamel.yaml.scalarstring import LiteralScalarString
+import re
+from functools import reduce
+
+yaml = ruamel.yaml.YAML(typ='rt')
+yaml.indent(mapping=2, sequence=4, offset=2)
+
+# https://stackoverflow.com/questions/78075152/preserve-line-wrapping-in-ruamel-yaml
+# needed to convert flow sclars (>-) to literal blocks (|)
```
B4nan, line 13:
```suggestion
# needed to convert flow scalars (>-) to literal blocks (|)
```

apify-api/scripts/add-redoc-legacy-urls-to-operations.py
```diff
@@ -0,0 +1,67 @@
...
+        redoc_url = f'https://docs.apify.com/api/v2#tag/{safe_tag}/operation/{operation['operationId']}'
+        operation['x-legacy-doc-urls'].append(redoc_url)
+
+    with open(path_path, 'w') as file:
+        yaml.dump(path_yaml, file)
+
+
+
+
+
```
B4nan, line 67:
```suggestion

```

B4nan (approved):
just a few nits about the scripts, nothing blocking, checked a few URLs are all good

apify-api/scripts/add-redoc-legacy-urls-to-tags.py
netmilk: It's a one-off script. It worked for this use case. I'm leaving the scripts behind to document my thought process and have a paper trail of the patterns for manipulating the yaml for future reference.

PR #1393, opened 2025-01-08
fix: correct limits for dataset get items endpoint

metalwarrior665:
See test runs:
2000 columns - https://console.apify.com/view/runs/zQPSs9IcesvwlZfiw
unlimited items - https://console.apify.com/view/runs/mnPtQvpRJCLtFevg3

I think these changes are a few years old by now so it would be good if someone took a good look.


B4nan (approved):
fine by me, but let's also include @drobnikj or someone else from the platform team to confirm the changes

drobnikj (approved):
Thanks for the fix!

PR #1411, opened 2025-01-14
feat: switch new OpenAPI docs to `/api/v2`

B4nan:


static/robots.txt
```diff
@@ -1,4 +1,5 @@
 User-agent: *
 Sitemap: https://docs.apify.com/sitemap.xml
 Disallow: https://docs.apify.com/api/v2-old
-Disallow: https://docs.apify.com/api/v2-new
+Disallow: https://docs.apify.com/api/v2-redoc
+Disallow: https://docs.apify.com/api/v2
```
netmilk, line 5: We want the `/api/v2` to be indexed, and this would disallow that, wouldn't  it?

```suggestion
```

apify-docs-theme/static/js/custom.js
```diff
@@ -78,8 +78,7 @@ function scrollOpenApiSidebarItemIntoView() {
 function redirectOpenApiDocs() {
     const { hash, pathname, origin } = new URL(window.location.href);
 
-    // TODO change to '/api/v2'
-    if (pathname.replace(/\/$/, '') !== '/api/v2-new') {
+    if (pathname.startsWith('/api/v2') && pathname.startsWith('/api/v2-')) {
```
netmilk, line 81: isn't the `pathname.startsWith('/api/v2')` part redundant?

netmilk (changes requested):

apify-docs-theme/static/js/custom.js
B4nan: right, should be `if (!pathname.startsWith('/api/v2') || pathname.startsWith('/api/v2-')) {`

netmilk (approved):

PR #1413, opened 2025-01-16
fix: remove trailing slash from internal URL

B4nan:
This fixes the lychee 404 errors.

https://github.com/apify/apify-docs/actions/runs/12799427833


B4nan:
@TC-MO now this one please 🙏 😄

PR #1415, opened 2025-01-17
docs: add full schema link

TC-MO:
Add full schema link to API docs introductory paragraph

### DIFF
--- a/apify-api/openapi/openapi.yaml
+++ b/apify-api/openapi/openapi.yaml
@@ -12,10 +12,7 @@ info:
     around [RESTful](https://en.wikipedia.org/wiki/Representational_state_transfer)
     HTTP endpoints. 
     
-You can download the complete OpenAPI schema of Apify API in the [YAML](http://docs.apify.com/api/openapi.yaml) or [JSON](http://docs.apify.com/api/openapi.json) formats. The source code is available on [GitHub](https://github.com/apify/apify-docs/tree/master/apify-api/openapi).
-    
-    - [OpenAPI YAML](http://docs.apify.com/api/openapi.yaml)
-    - [OpenAPI JSON](http://docs.apify.com/api/openapi.json)
+    You can download the complete OpenAPI schema of Apify API in the [YAML](http://docs.apify.com/api/openapi.yaml) or [JSON](http://docs.apify.com/api/openapi.json) formats. The source code is also available on [GitHub](https://github.com/apify/apify-docs/tree/master/apify-api/openapi).
 
     All requests and responses (including errors) are encoded in
     [JSON](http://www.json.org/) format with UTF-8 encoding,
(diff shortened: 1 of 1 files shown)


apify-api/openapi/openapi.yaml
```diff
@@ -10,7 +10,9 @@ info:
     The Apify API (version 2) provides programmatic access to the [Apify
     platform](https://docs.apify.com). The API is organized
     around [RESTful](https://en.wikipedia.org/wiki/Representational_state_transfer)
-    HTTP endpoints.
+    HTTP endpoints. 
+    
+    You can check the complete OpenAPI schema [here](https://docs.apify.com/api/apify-api.yaml).
```
jancurn, line 15: Wouldn't it be better to link https://github.com/apify/apify-docs/tree/master/apify-api/openapi instead of this strange page? People can find more in that directory. Also, please let's add JSON version

B4nan: It's the very same content, compiled into one file. I believe this is what anyone would use if they want to work with our schema.

I will add the JSON version tomorrow, again, it will be a single file. We can surely link to the source files if that makes some sense (but I am not sold on that really, who does that, what would be the use case?).

Thoughts @netmilk?

jancurn: Oh yeah, but still I think it's good to link the GitHub source code from which the actual API docs are generated. We do the same for all our docs pages. And who knows, maybe somebody will send a PR...

TC-MO: That makes more sense to me than to have a button & separate link within prose

B4nan: Btw I also changed the URL (as agreed with Adam) to http://docs.apify.com/api/openapi.yaml and http://docs.apify.com/api/openapi.json in that PR, so if we want to keep it in the description, we'll have to adjust that.

TC-MO: I've adjusted the description, it's now shown in a bullet list like this 

image

but with the changed label on the button, it seems redundant.

Any thoughts on this @jancurn, @netmilk  ?

apify-api/openapi/openapi.yaml
```diff
@@ -10,7 +10,12 @@ info:
     The Apify API (version 2) provides programmatic access to the [Apify
     platform](https://docs.apify.com). The API is organized
     around [RESTful](https://en.wikipedia.org/wiki/Representational_state_transfer)
-    HTTP endpoints.
+    HTTP endpoints. 
+    
+    You can check the complete OpenAPI schema in the following formats:
```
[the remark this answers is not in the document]
TC-MO: Sure thing

jancurn (approved):
Great cheers

B4nan (approved):
I am still not sold on the need for this, but it wont hurt either 🙃 

Note that the links open the content instead of triggering download (I don't think we can force download it via markdown links, we do that in the dropdown I added via #1419) .

PR #1419, opened 2025-01-20
feat: allow downloading openapi specs in JSON

B4nan:


netmilk (approved):
Good to go, thank you!

PR #1421, opened 2025-01-21
fix: treat component props as immutable to make react happy again

B4nan:


barjin (approved):
Yep, this should do it 👍🏽 Cheers!

PR #1422, opened 2025-01-21
feat(api): added openapi endpoints

danpoletaev:
Recently, we introduced two new endpoints that return the OpenAPI specification:

- `/acts/<actorId>/builds/<buildId>/openapi-specification`
- `/actor-builds/<buildId>/openapi-specification`

**This PR updates the documentation to include these endpoints.**

@TC-MO Let me, please, know if you think we should add any additional information 🙏

### DIFF
--- a/apify-api/openapi/components/schemas/actor-builds/GetOpenApiResponse.yaml
+++ b/apify-api/openapi/components/schemas/actor-builds/GetOpenApiResponse.yaml
@@ -1,0 +1,348 @@
+title: GetOpenApiResponse
+type: object
+properties:
+  openapi:
+    type: string
+    example: 3.0.1
+  info:
+    type: object
+    properties:
+      title: 
+        type: string
+        example: Your Magic Actor
+      version: 
+        type: string
+        example: "1.0"
+      x-build-id: 
+        type: string
+        example: "ID of build"
+  servers:
+    type: array
+    items:
+      type: object
+      properties:
+        url: 
+          type: string
+          example: https://api.apify.com/v2
+  paths:
+    type: object
+    properties:
+      /acts/<username>~<actor>/run-sync-get-dataset-items:
+        type: object
+        properties:
+          post:
+            type: object
+            properties:
+              operationId:
+                type: string
+                example: run-sync-get-dataset-items
+              x-openai-isConsequential:
+                type: boolean
+                example: false
+              summary:
+                type: string
+                example: Executes an Actor, waits for its completion, and returns Actor's dataset items in response.
+              tags:
+                type: array
+                items:
+                  type: string
+                example: ["Run Actor"]
+              requestBody:
+                type: object
+                properties:
+                  required:
+                    type: boolean
+                    example: true
+                  content:
+                    type: object
+                    properties:
+                      application/json:
+                        type: object
+                        properties:
+                          schema:
+                            type: object
+                            properties:
+                              $ref:
+                                type: string
+                                example: '#/components/schemas/inputSchema'
+              parameters:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    name:
+                      type: string
+                      example: token
+                    in:
+                      type: string
+                      example: query
+                    required:
+                      type: boolean
+                      example: true
+                    schema:
+                      type: object
+                      properties:
+                        type:
+                          type: string
+                          example: string
+                    description:
+                      type: string
+                      example: Enter your Apify token here
+              responses:
+                type: object
+                properties:
+                  '200':
+                    type: object
+                    properties:
+                      description:
+                        type: string
+                        example: OK
+      /acts/<username>~<actor>/runs:
+        type: object
+        properties:
+          post:
+            type: object
+            properties:
+              operationId:
+                type: string
+                example: runs
+              x-openai-isConsequential:
+                type: boolean
+                example: false
+              summary:
+                type: string
+                example: Executes an Actor and returns information about the initiated run in response.
+              tags:
+                type: array
+                items:
+                  type: string
+                example: ["Run Actor"]
+              requestBody:
+                type: object
+                properties:
+                  required:
+                    type: boolean
+                    example: true
+                  content:
+                    type: object
+                    properties:
+                      application/json:
+                        type: object
+                        properties:
+                          schema:
+                            type: object
+                            properties:
+                              $ref:
+                                type: string
+                                example: '#/components/schemas/inputSchema'
+              parameters:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    name:
+                      type: string
+                    in:
+                      type: string
+                      example: query
+                    required:
+                      type: boolean
+                    schema:
+                      type: object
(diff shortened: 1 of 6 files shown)


jirimoravcik (approved):
Didn't we also add an endpoint for default build? Is it going to be added in another PR?

apify-api/openapi/components/tags.yaml
```diff
@@ -251,6 +251,17 @@
...
+  x-displayName: Build OpenAPI Specification
+  x-parent-tag-name: Actors
+  x-trait: 'true'
+  description: |
+    To obtain the OpenAPI schema for actor builds, two similar endpoints are available:
+
+    - [First Endpoint](/api/v2/act-openapi-specification-get): Requires both `actorId` and `buildId`. You can pass `default` instead of buildId to retrieve the OpenAPI schema for the default actor build.
+    - [Second Endpoint](/api/v2/actor-build-openapi-specification-get): Requires only `buildId`.
+  x-legacy-doc-urls:
+  - '#/reference/actors/openapi-specification'
```
B4nan, line 264: if those endpoints were not there before, i guess it doesn't make sense to have the apiary legacy URLs for them either. or can this URL be used somewhere already?

apify-api/openapi/components/tags.yaml
```diff
@@ -251,6 +251,17 @@
     ```
 
     In order to save new items to the dataset, send HTTP POST request with JSON payload to the same URL.
+- name: Actors/Build OpenAPI specification
+  x-displayName: Build OpenAPI Specification
```
B4nan, line 255: maybe its just me but it feels weird to call an endpoint "build specs", that's an implementation detail of the endpoint that it needs to build it, right? i'd be in favor of "Get OpenAPI specification". i can understand where this is coming from (specs can differ for each build), but it feels rather confusing to me to call it this way.

also, why should `Specification` be capitalized here?

danpoletaev:
> Didn't we also add an endpoint for default build? Is it going to be added in another PR?

I'll add it in another PR

apify-api/openapi/components/tags.yaml
danpoletaev: Those endpoints are new, so I removed `x-legacy-doc-urls`

apify-api/openapi/components/tags.yaml
danpoletaev: OpenAPI specification belongs to build - that's why I named it this way. But it makes sense to me that this is confusing, so I changed it to "Get OpenAPI Specification" 🙃

apify-api/openapi/components/tags.yaml
```diff
@@ -548,6 +557,15 @@
...
   description: Check out [Logs](#/reference/logs) for full reference.
+- name: Actor builds/Get OpenAPI specification
+  x-displayName: Get OpenAPI specification
+  x-parent-tag-name: Actor builds
+  x-trait: 'true'
+  description: |
+    To obtain the OpenAPI schema for actor builds, two similar endpoints are available:
+
+    - [First Endpoint](/api/v2/act-openapi-specification-get): Requires both `actorId` and `buildId`. You can pass `default` instead of buildId to retrieve the OpenAPI schema for the default actor build.
+    - [Second Endpoint](/api/v2/actor-build-openapi-specification-get): Requires only `buildId`.
```
TC-MO, line 568:
```suggestion
  description: |
    Get the OpenAPI specification for Actor builds. Two similar endpoints are available:

    - [First endpoint](/api/v2/act-openapi-specification-get): Requires both `actorId` and `buildId`. Use `default` as the `buildId` to get the OpenAPI schema for the default Actor build.
    - [Second endpoint](/api/v2/actor-build-openapi-specification-get): Requires only `buildId`.
```

apify-api/openapi/components/tags.yaml
```diff
@@ -251,6 +251,15 @@
...
     In order to save new items to the dataset, send HTTP POST request with JSON payload to the same URL.
+- name: Actors/Get OpenAPI specification
+  x-displayName: Get OpenAPI specification
+  x-parent-tag-name: Actors
+  x-trait: 'true'
+  description: |
+    To obtain the OpenAPI schema for actor builds, two similar endpoints are available:
+
+    - [First Endpoint](/api/v2/act-openapi-specification-get): Requires both `actorId` and `buildId`. You can pass `default` instead of buildId to retrieve the OpenAPI schema for the default actor build.
+    - [Second Endpoint](/api/v2/actor-build-openapi-specification-get): Requires only `buildId`.
```
TC-MO, line 262:
```suggestion
  description: |
    Get the OpenAPI specification for Actor builds. Two similar endpoints are available:

    - [First endpoint](`/api/v2/act-openapi-specification-get`): Requires both `actorId` and `buildId`. Use `default` as the `buildId` to get the OpenAPI schema for the default Actor build.
    - [Second endpoint](`/api/v2/actor-build-openapi-specification-get`): Requires only `buildId`.
```

TC-MO (changes requested):
I've rewritten the descriptions slightly to address few things


- Actor capitalization
- Consistency
- Small descriptions to give more context instead just giving the note about similar endpoints

apify-api/openapi/paths/actor-builds/actor-builds@{buildId}@openapi-specification.yaml
```diff
@@ -0,0 +1,31 @@
+get:
+  tags:
+    - Actor builds/Get OpenAPI specification
+  summary: Get OpenAPI specification
+  description: |
+
+    > **You can also use a [similar endpoint](/api/v2/act-openapi-specification-get)**
+    > from the `acts` namespace to retrieve the OpenAPI specification for a build.
+
+    This endpoint does not require the authentication token. Instead, calls are authenticated using a hard-to-guess ID of the build.
```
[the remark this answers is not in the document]
B4nan: nit: I don't think you need the first blank line now, only if there is a quote

apify-api/openapi/components/schemas/actor-builds/GetOpenApiResponse.yaml
```diff
@@ -0,0 +1,348 @@
...
+    example: 3.0.1
+  info:
+    type: object
+    properties:
+      title: 
+        type: string
+        example: Your Magic Actor
+      version: 
+        type: string
+        example: "1.0"
```
B4nan, line 15: the quoting is inconsistent here, lets use single quotes everywhere (there are more places using double quotes, not just in this file)

TC-MO (approved):

B4nan (approved):
just one nit

apify-api/openapi/components/schemas/actor-builds/GetOpenApiResponse.yaml
danpoletaev: Hm, fixed in the files I've changed, but it is harder to change everywhere else. I think it should be fixed in lint, but we'll need to set up yaml lint for that...

PR #1428, opened 2025-01-23
fix: do not double wrap GitHub user links in changelog

barjin:
The new `git-cliff`-based workflow in Apify SDK Python already generates the changelog with actual Markdown links for users. Together with the logic from the theme, this caused the userlinks to be "doubled" - see below:

obrazek


barjin:
There are still some more (arguably smaller) issues there, though:

- The double heading - see screenshot above (maybe we could remove the first heading from the loaded changelog here, if we see it's only **Changelog**?)
- At the very end, Docusaurus renders the `<!-- generated by git-cliff -->` comment, since we escape special characters in our sanitization script. Maybe that doesn't have to be there? cc @janbuchar 

obrazek

B4nan (approved):

PR #1430, opened 2025-01-24
fix: improve Vale configuration

honzajavorek:
- Should allow backtick Markdown code blocks with additional parameters, such as `title="..."`
- Should prevent Vale to spellcheck some inline code blocks
- Fixes some additional, even unrelated lint issues


B4nan (approved):
nice

PR #1433, opened 2025-01-27
docs: vale fixes

TC-MO:
Add various exemptions to accept.txt
fix initial issues within academy

### DIFF
--- a/.github/styles/config/vocabularies/Docs/accept.txt
+++ b/.github/styles/config/vocabularies/Docs/accept.txt
@@ -105,3 +105,6 @@ learnings
 subwoofer
 captcha
 captchas
+deduplicating
+
+jQuery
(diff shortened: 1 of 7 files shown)


.github/styles/config/vocabularies/Docs/accept.txt
```diff
@@ -93,3 +92,19 @@ preconfigured
...
 parallelization
 IMDb
+iPhone
+iPhones
+iPad
+iPads
+screenshotting
+Fakestore
+SKUs
+SKU
```
honzajavorek, line 102: `SKUs?` should be the same as separate `SKU` and `SKUs` (the same is valid for the other terms)

honzajavorek (approved):
The `accept.txt` can be optimized, but I don't think it should block merging this PR.

PR #1436, opened 2025-01-27
chore: update lockfile

B4nan:
resolves one security report (katex)


barjin (approved):
Seems fine to me, as long as we're not bumping anything in `package.json`, this should be fairly safe to do. Thanks!

PR #1439, opened 2025-01-28
chore: update eslint to v9

### LINKED ISSUE #1276: Update eslint to v9, use new flat Apify eslint config
We should update eslint to v9 along with the new flat Apify eslint config that supports it.

See [notion.so/apify/New-linting-and-formatting-configuration-across-Apify-0b62890892cd4b71b4fe2dee9291b4af?pvs=4#6152e086487441f6b974e6fbda9614be](https://www.notion.so/apify/New-linting-and-formatting-configuration-across-Apify-0b62890892cd4b71b4fe2dee9291b4af?pvs=4#6152e086487441f6b974e6fbda9614be) for a detailed guide.

B4nan:
Closes #1276


apify-docs-theme/src/theme/Layout/index.jsx
```diff
@@ -1,9 +1,9 @@
-import React from 'react';
-// cannot use any of the theme aliases here as it causes a circular dependency :( ideas welcome
+import { useLocation } from '@docusaurus/router';
 import Layout from '@docusaurus/theme-classic/lib/theme/Layout/index';
-import { usePluginData } from '@docusaurus/useGlobalData';
 import useBaseUrl from '@docusaurus/useBaseUrl';
-import { useLocation } from '@docusaurus/router';
+import { usePluginData } from '@docusaurus/useGlobalData';
+import React from 'react';
+// cannot use any of the theme aliases here as it causes a circular dependency :( ideas welcome
```
vladfrangu, line 6: this comment was moved from its place, either drop or move back if possible

eslint.config.mjs
```diff
@@ -0,0 +1,55 @@
...
+        ignores: ['**/dist', 'node_modules', '.docusaurus', 'build', 'sources/api'],
+    },
+    ...apify,
+    {
+        languageOptions: {
+            parserOptions: {
+                project: 'tsconfig.eslint.json',
+            },
+        },
+    },
```
vladfrangu, line 17: I'm not fully sure if this works, in my own project I had to do something like this to define tsconfig for the rules but maybe it works without it anymore
 https://github.com/vladfrangu/highlight/blob/7d1a50341d175192e2cf54219d730f8d58407944/eslint.config.js#L19-L27

B4nan: i believe it does, without it i get warnings like 

```
Error: Error while loading rule '@typescript-eslint/no-floating-promises': You have used a rule which requires type information, but don't have parserOptions set to generate type information for this file. See https://typescript-eslint.io/getting-started/typed-linting for enabling linting with type information.
Parser: typescript-eslint/parser
```

and before I've added the sources and examples folders it was complaining about that like 

```
/Users/adamek/htdocs/apify/apify-docs/examples/ts-parallel-scraping/orchestrator/src/main.ts
  0:0  error  Parsing error: "parserOptions.project" has been provided for @typescript-eslint/parser.
The file was not found in any of the provided project(s): examples/ts-parallel-scraping/orchestrator/src/main.ts

/Users/adamek/htdocs/apify/apify-docs/examples/ts-parallel-scraping/scraper/src/main.ts
  0:0  error  Parsing error: "parserOptions.project" has been provided for @typescript-eslint/parser.
The file was not found in any of the provided project(s): examples/ts-parallel-scraping/scraper/src/main.ts
```

so i guess this is enough (and the CI confirms that)

B4nan: i guess its because the apify eslint config is using that plugin

https://github.com/apify/apify-eslint-config/blob/master/index.js#L145

vladfrangu (approved):

apify-docs-theme/src/markdown.js
```diff
@@ -95,7 +75,7 @@ const prettifyPRLinks = () => (tree) => {
         const prLinkRegex = /https:\/\/github.com\/[^\s]+\/pull\/(\d+)/g;
         const match = prLinkRegex.exec(node.value);
 
-        if (!match) return;
+        if (!match) return 0;
```
barjin, line 78: I have a hunch this might be why the `apify-sdk-python` docs build is stalling, as [returning an index from the visitor](https://github.com/syntax-tree/unist-util-visit-parents?tab=readme-ov-file#index) means "continue by visiting my n-th sibling", which, with 0, might potentially cause an infinite loop.

Was this a quick eslint's `consistent-return` fix, or did you have a deeper reason for this? 😅

B4nan: > Was this a quick eslint's consistent-return fix, or did you have a deeper reason for this? 😅

Yeah, exactly, feel free to revert that.

Mixed return types are evil, especially in vanilla JS. This is just another proof :D

PR #1445, opened 2025-02-06
docs: reorganize api docs

TC-MO:


TC-MO:
@netmilk from my lcoal testing, it seems that endpoints themselves did not change URLs whatsoever with those changes, just the subgroups did and for that I've used nginx to redirect to the new groupings.

B4nan (approved):

PR #1447, opened 2025-02-06
docs: Update Gmail and Drive integration docs

valekjo:
With the migration away from appmixer, we need to update docs for gmail and drive integrations, because those work a bit differently now.

I've dropped the videos, because those are showing the old flow too (cc @TheoVasilis )


TC-MO (approved):
LGTM, just a few suggestion based on the screenshots & adjustments to formatting

PR #1449, opened 2025-02-07
docs: add redirect for python docs

TC-MO:


B4nan (approved):
lets give it a try, worst case we can revert it

PR #1451, opened 2025-02-10
docs: update privacy policy

lenkab151:


TC-MO (approved):
Fixed the issue with vale triggering on /legal directory, is should now completely ignore it while linting , we may even get rid of `<!-- vale off -->` now as it should be unnecessary

mnmkng (approved):

PR #1455, opened 2025-02-11
docs: Better explenation of automatic dependencies

martin-matess-mares:


tobice (approved):

TC-MO (approved):

TC-MO:
I guess the issue with UI package is still biting us when someone without proper setup tries to contribute to docs... 
I don't foresee this breaking any setup so I think this is safe to merge,

PR #1458, opened 2025-02-12
feat: Mention that webhooks are sent through select static IPs

fnesveda:
We have started sending webhooks from the Apify platform through a predefined list of static IP addresses. This enables users (typically enterprise customers) who have the webhook destinations behind a firewall to whitelist these IP addresses in their firewall, and allow the delivery of these webhooks.


seyhello (approved):

fnesveda:
@TC-MO I will merge this so that it's ready for all hands, but feel free to do any changes later 🙂

TC-MO:
If you are okay with the changes suggested I'll make them later today :)

fnesveda:
I don't see any suggested changes 🤔

TC-MO:
I did them as a comment in review not `request changes` about the admonition & etc

PR #1460, opened 2025-02-13
docs:change placement of the note & turn it into admonition

TC-MO:
turned new paragraph into admonition
changed placement
added details tag with IP list


fnesveda (approved):
Thanks! 👏

PR #1477, opened 2025-02-25
feat: analytics support

katzino:
Part of [#4455](https://github.com/apify/apify-web/issues/4455)


webrdaniel (approved):

docusaurus.config.js
```diff
@@ -46,6 +46,21 @@ module.exports = {
                 sizes: 'any',
             },
         },
+        // Segment analytics
+        process.env.SEGMENT_TOKEN && {
```
jbartadev, line 50: @katzino How is the `SEGMENT_TOKEN` env var used here? 🤔 I don't see it anywhere in the script 😄

PR #1480, opened 2025-02-26
docs: typo

MatousMarik:


sources/platform/actors/development/programming_interface/basic_commands.md
```diff
@@ -299,7 +299,7 @@ The SDK provides convenient methods for exiting Actors:
 
 1. Use `exit()` with custom messages to inform users about the Actor's achievements or issues.
 
-2. The `exit()` method emits `exit` event allowing components to perform cleanup or state persistence.
+2. The `fail()` method emits `exit` event allowing components to perform cleanup or state persistence.
```
B4nan, line 302: the `fail` method calls the `exit` method. both of them will issue `process.exit()` unless explicitly told not to (or better say, the `exit` method is the one doing this, `fail` just calls it), the difference is `fail` defaults to exit code 1.

https://github.com/apify/apify-sdk-js/blob/master/packages/apify/src/actor.ts#L292-L294

TC-MO:
why are we still getting 401's? I thought the package issue has been solved?

sources/platform/actors/development/programming_interface/basic_commands.md
TC-MO: So we could change it to :
```suggestion
1. Use `exit()/``fail()`` with custom messages to inform users about the Actor's achievements or issues.
2. The `exit()`/`fail()` method emits `exit` event allowing components to perform cleanup or state persistence.
```
?

Or am I misunderstanding it?

TC-MO: Righ then maybe something like this, with bullets instead of ordered list

```suggestion
- Use `exit()` with custom messages to inform users about the Actor's achievements or issues.
- Use `fail()` as a shortcut for `exit()` when indicating an error. It defaults to an exit code of `1` and emits the `exit` event, allowing components to perform cleanup or state persistence.
- The `exit()` method also emits the `exit` event, enabling cleanup or state persistence.
```

B4nan: yeah that's much better 👍

B4nan (approved):

PR #1509, opened 2025-03-26
fix: navigation structure for Emails

tomnosek:


honzajavorek (commented):
I don't have the developer account to verify, I'm sorry. Otherwise I'd rubber stamp it.

TC-MO (approved):
Verified, LGTM thank you!

PR #1510, opened 2025-03-26
docs: add missing actor-tasks endpoint

TC-MO:
add missing actor-tasks endpoint
add TODO's to be discussed regarding categorization & linking add missing endpoint to openapi.yaml

### DIFF
--- a/apify-api/openapi/paths/actor-tasks/actor-tasks@{actorTaskId}@runs@last.yaml
+++ b/apify-api/openapi/paths/actor-tasks/actor-tasks@{actorTaskId}@runs@last.yaml
@@ -1,0 +1,161 @@
+get:
+  tags:
+    - Actor tasks
+  summary: Get last run
+  description: |
+    This is not a single endpoint, but an entire group of endpoints that lets you to
+    retrieve and manage the last run of given actor task or any of its default storages.
+    All the endpoints require an authentication token.
+
+    The endpoints accept the same HTTP methods and query parameters as
+    the respective storage endpoints.
+    The base path represents the last actor task run object is:
+
+    TODO confirm correct endpoint here, current seems to be deprecated
+
+    `/v2/actor-tasks/{actorTaskId}/runs/last{?token,status}`
+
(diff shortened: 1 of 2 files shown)


TC-MO:
Changed from draft too early, my mistake still needs work

apify-api/openapi/paths/actor-tasks/actor-tasks@{actorTaskId}@runs@last.yaml
```diff
@@ -0,0 +1,157 @@
+get:
+  tags:
+    - Actor tasks
+  summary: Get last run
+  description: |
+    This is not a single endpoint, but an entire group of endpoints that lets you to
+    retrieve and manage the last run of given actor task or any of its default storages.
+    All the endpoints require an authentication token.
```
janbuchar, line 8: Not gonna lie, this is confusing. Can you spare the time to split this into four endpoints instead, each with their own documentation? 

Of course, if that's something you want to pursue later, we can merge this as is - it sure is better than nothing

TC-MO: Oh I fully agree. This is something that is next on TODO list, to split this and every other endpoint that is done this way into separate ones. This PR was meant to just reintroduce the missing endpoint so we have parity between old API Blueprint and current OpenAPI

TC-MO: If I'm not mistaken we have two like this already

[Get last run](https://docs.apify.com/api/v2/act-runs-last-get)
[Get run](https://docs.apify.com/api/v2/actor-run-get)

and this would be the third like tha.

janbuchar (approved):
LGTM - some doc is better than undocumented endpoints. I imagine you won't be able to use the "try it out" box in the docs that comfortably, which is a shame...

fnesveda (approved):

PR #1517, opened 2025-04-01
chore: use `@apify/ui-components` package from NPM

B4nan:
This PR adopts the new `@apify/ui-components` library which was previously available only on our private NPM registry. This means we should be finally open source again - and people from outside apify org should be able to run things locally (as well as PR checks should work again on PRs from forks).

I also did a full lockfile update to be sure we have no duplicates or dated dependencies after the update.


HonzaKirchner (approved):
🎉

PR #1518, opened 2025-04-01
chore: configuration for docs-preview

seyhello:
- nginx configuration for apify-docs-preview - everything pointed to *.preview.docs.apify.com domain will be proxied to the apify-docs-preview S3 bucket directly. It will be used for "staging/preview" documentation


fnesveda:
@seyhello I have zero context for this, could you give me some?

nginx.conf
```diff
@@ -441,3 +441,15 @@ server {
...
   rewrite ^/python/docs/?$ /python/docs/quick-start;
 }
+
+server {
+    listen 80;
+
+    server_name ~^(?<subdomain>[^.]+)\.preview\.docs\.apify\.com$;
+
+    location / {
+        rewrite ^/(.*)/$ https://s3.amazonaws.com/apify-docs-preview/$subdomain/$1/index.html permanent;
```
fnesveda, line 451: As discussed in person - let's try to use `proxy_pass` here so that it doesn't require enabling public read on the bucket.

PR #1519, opened 2025-04-01
docs: add new getting started to API docs

TC-MO:
add new getting started page to API docs
add exception for new page in .gitignore file
add basic workflow to getting started

### DIFF
--- a/sources/api/getting-started.mdx
+++ b/sources/api/getting-started.mdx
@@ -1,7 +1,6 @@
 ---
 title: Getting started with Apify API
 sidebar_label: Getting started
-displayed_sidebar: api
 slug: /getting-started
 ---
 
--- a/sources/api/getting-started.mdx
+++ b/sources/api/getting-started.mdx
@@ -54,7 +54,7 @@ Expected response codes:
 
 ## Basic workflow
 
-A most common workflow involving Apify API consists of the following steps:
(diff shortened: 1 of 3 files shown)


sources/api/getting-started.mdx
```diff
@@ -0,0 +1,183 @@
+---
+title: Getting started with Apify API
+sidebar_label: Getting started
+displayed_sidebar: api
+slug: /getting-started
```
B4nan, line 5: shouldn't this be `/api/getting-started`?

sources/api/getting-started.mdx
```diff
@@ -0,0 +1,183 @@
...
+Before you can start using the API, check if you have all the necessary prerequisites:
+
+* An Apify account with an API token.
+* A tool to make HTTP requests (cURL, Postman, or your preferred programming language).
+
+## Authentication
+
+You must authenticate all API requests presented on this page. You can authenticate using your API token:
+
+`Authorization: Bearer YOUR_API_TOKEN`
```
B4nan, line 37: shouldn't this be code block (I mean not an inline one)

TC-MO: It's still a work in progress/draft I'm looking at what looks best, thanks for suggestion!

PR #1543, opened 2025-04-17
docs: update publish.mdx

mhamas:
Adding section on logos.

### DIFF
--- a/sources/platform/actors/publishing/publish.mdx
+++ b/sources/platform/actors/publishing/publish.mdx
@@ -32,6 +32,7 @@ To verity that your Actor has been published successfully, go to the [Apify Stor
 ![Actor page](./images/actor-page.webp)
 
 ### Logo
+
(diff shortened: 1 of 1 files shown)


sources/platform/actors/publishing/publish.mdx
```diff
@@ -31,6 +31,9 @@ To verity that your Actor has been published successfully, go to the [Apify Stor
 
 ![Actor page](./images/actor-page.webp)
 
+### Logo
+We strongly recommend adding a unique image to your Actor that visually represents the service it provides. This helps users quickly understand its purpose. However, do not use official logos or branded images from the sites you're scraping, as this can lead to copyright or trademark issues.
```
TC-MO, line 35: This should fix markdownlint issues

```suggestion
### Logo

We strongly recommend adding a unique image to your Actor that visually represents the service it provides. This helps users quickly understand its purpose. However, do not use official logos or branded images from the sites you're scraping, as this can lead to copyright or trademark issues.
```

TC-MO (approved):

PR #1547, opened 2025-04-23
fix: missing fields on dataset endpoint response

stepskop:
Adds `stats` and `schema` fields to the response definition of `GET /datasets/:id`

Closes [#20565](https://github.com/apify/apify-core/issues/20565)

### DIFF
--- a/apify-api/openapi/components/schemas/datasets/Dataset.yaml
+++ b/apify-api/openapi/components/schemas/datasets/Dataset.yaml
@@ -50,6 +50,22 @@ properties:
   schema:
     type: object
     nullable: true
+    description: Defines the schema of items in your dataset, the full specification can be found in [Apify docs](/platform/actors/development/actor-definition/dataset-schema)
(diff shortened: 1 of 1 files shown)


apify-api/openapi/components/schemas/datasets/Dataset.yaml
```diff
@@ -47,6 +47,11 @@ properties:
       type: string
     description: ''
     nullable: true
+  schema:
+    type: object
+    nullable: true
```
janbuchar, line 52: Asking to describe the schema in OpenAPI would be too much, but a link to the output schema spec would be solid gold.

Also, could you add a minimal example here, too?

stepskop: Added mentioned. 
image

janbuchar (approved):

fnesveda (approved):

PR #1552, opened 2025-04-28
fix: remove the TypeScript course

honzajavorek:
The TypeScript course is an offtopic in the context of the Academy. It is not a web scraping course, and it is not related to the Apify platform per se. I see no reason why Apify should be teaching TypeScript. And I think the team has already enough other, more relevant docs to maintain.

When I joined, I wrote down an internal document called **Apify Academy white paper**, where I suggested deleting this content. I think it's more or less 1 year since I joined Apify, and I decided: Today is the day! Today I'm bold enough to actually propose the change in a PR. If this gets merged, I'll consider it as a gift to my work anniversary 😆 

In the document, @mnmkng wrote the following:

> I think it was created mostly to help users make the switch to TypeScript with Crawlee when we released full TS support. I don't even remember what's in there frankly and I'm pretty sure we can live without it.

Miley Cyrus


TC-MO (approved):
From my POV more than fine, it always felt out of place within Academy

mnmkng:
I'm wondering if perhaps @metalwarrior665 uses it to onboard new devs. If he does, I'd propose merging it into one long post and dropping it into tutorials, or moving it to Notion at least. If he doesn't, let's delete it.

honzajavorek:
OK, I'll wait for @metalwarrior665's input.

honzajavorek:
Do you think it's worth to redirect the URLs somewhere or should I just let them 404?

mnmkng:
> Do you think it's worth to redirect the URLs somewhere or should I just let them 404?

404s suck for SEO, so maybe let's create a single page document in tutorials with a similar title, redirect there, and explain that we recommend a list of good external typescript courses?

metalwarrior665:
For lazy solution, it could redirect to https://apify.com/templates/categories/typescript even though it is not all related to scraping

honzajavorek:
I like the lazy solution 😅

TC-MO:
I would second the lazy solution

honzajavorek:
It's enough to add a rewrite to nginx.conf, right? Does it auto-deploy after merge?

TC-MO:
From what I remember it should work like that

honzajavorek:
Hopefully I got the redirect right. Can someone please check? I asked ChatGPT and it told me it swears it's correct, but I'd appreciate human eyes as well 😅

mnmkng:
> Hopefully I got the redirect right. Can someone please check? I asked ChatGPT and it told me it swears it's correct, but I'd appreciate human eyes as well 😅

@B4nan 🙏

B4nan (approved):
the redirect looks good

PR #1568, opened 2025-05-06
style: consistency fixes

honzajavorek:
I made final reading of the whole course. I fixed a few things and I unified WE vs YOU with some exceptions, like admonitions, etc. The lessons whould be WE, the exercises should be YOU.


TC-MO (approved):
One nit, otherwise LGTM

honzajavorek:
Thank you!

PR #1579, opened 2025-05-13
Re-structure the JS course so that it's a flat list of numbered lessons

honzajavorek:
_work in progress_


B4nan:
Huh, do we really want this? I like the current form with nesting better to be honest. Are you sure you are not reading that "zen of python" too much in between the lines? Isn't it about code rather than docs?

honzajavorek:
The Zen of Python is there more as an easter egg. I should have provided a real explanation.

I'm of an opinion that courses should be a flat series of numbered lessons, where the only navigation is prev & next. It makes orientation straightforward and I immediately know where I am, what is ahead, what is my progress, etc.

This is a first step in revamp of the JS course, which I'm supposed to do now, as the Python one is done. I want to first flatten it like this, and then make changes which will gradually make the sectioning disappear, and fix up the lessons so that they're more standalone and single topic oriented, similar to the ones in the Python course.

If possible, I'd also lower their number, as their amount is probably also a factor contributing to the need of sectioning.

So yes, I agree the result of this PR is not exactly the most appalling, but it's a temporary step towards a more polished course.

honzajavorek:
> But we have that even with the current structure, next page of the last nested item points to the first item in the next "nested list".

I know, technically it works. What I'm after is the mental model and overview of the reader. Which I've been myself, when I started in Apify, and the nesting was both confusing and overwhelming, because looking at the ToC I had no idea where I am regarding my progress, how many lessons I'm in and how many lessons is left.

B4nan:
I guess it depends on personal preferences, to me the nesting adds to clarity, since it groups what belongs together, and unless you open all the categories as you did in that screenshot, you only see what's relevant to the current tutorial. But if we keep the titles are they are in your screenshot, it's not far away from that.

honzajavorek:
The goal, in long term, would be to get closer to something like this:

Screenshot 2025-05-13 at 19-36-59 Web scraping basics for Python devs Academy Apify Documentation

TC-MO:
I get the goal in the long term and generally I concur, but do we want to make the experience worse in the short term ? (also those are some nebulous terms, short term can be a day or two or a month or two ¯\_(ツ)_/¯ )

honzajavorek:
Is it so much worse? I consider the orientation better. The next step would be improvement of the titles and making the lessons more standalone. But yes, with my part-time commitment, this can take a month or two.

If this is a concern, I can stack my next changes on top of this PR, but then it's gonna be rebasing hell if meanwhile `master` changes significantly.

metalwarrior665:
I don't have a strong opinion on this. Do you plan to do something similar with the other courses? For me nested structure makes sense if:

1. There are more articles than X (could be 10 or 15)
2. The subsections also make sense in isolation so they don't necessarily need to follow each other.

In our case, both of these are kinda borderline but the course sections probably follow each other

honzajavorek:
I'm putting this on hold for now to explore alternative approaches.

Issue #947, opened 2024-04-26
Update Academy's approach to setting up a Node.js project


honzajavorek:
Current content around setting up a Node.js project is outdated. They should be revisited and simplified, or even removed:

- https://docs.apify.com/academy/web-scraping-for-beginners/data-extraction/computer-preparation
- https://docs.apify.com/academy/web-scraping-for-beginners/data-extraction/project-setup
- https://docs.apify.com/academy/web-scraping-for-beginners/best-practices

Notes:

- The course links https://javascript.plainenglish.io/how-to-use-modular-patterns-in-nodejs-982f0e5c8f6e and explains package setup for Node.js, which can be simplified as of 2024
- Maybe we could use an official Node.js _hello world_ tutorial or something? Installing Node.js changes over time and is different for each OS.

Issue #1550, opened 2025-04-28
Update the Web scraping basics for JavaScript devs course


honzajavorek:
Per discussion with @mnmkng, once https://github.com/apify/apify-docs/issues/1319 is done, I should go through the JavaScript web scraping course for beginners and do a 'quick' check with at least basic fixes, so that it is presentable. I'm opening this issue for planning and tracking the work.

### Known issues

- https://github.com/apify/apify-docs/issues/947
- https://github.com/apify/apify-docs/issues/948
- https://github.com/apify/apify-docs/issues/950

### Goals

- Rename the JavaScript course to "Web scraping basics for JavaScript devs"
- Clearly distinguish the course intro from the rest and align it with the Python course
- Re-structure the course so that it's a flat list of numbered lessons
- Re-read the whole course and fix what can be quickly fixed
- Remove some parts which are relevant only for people who are completely new to coding

### Non-goals

- There's no time to spend effort on adding exercises
- There's no time to fully align the lessons with the Python course and create the same polished 'story arc', as in going from simple scripts towards a framework and the platform

PR #1584, opened 2025-05-16
feat: use the Python course as a template for a new JavaScript course

honzajavorek:
This is a **proof of concept** exploring whether using the Python course as a template for a brand new JavaScript course would take less time than trying to improve the original JavaScript course. This work is sanctioned by both @mnmkng and @TC-MO. If viable, this should have the potential to implement https://github.com/apify/apify-docs/issues/1550 and, of course, deprecate https://github.com/apify/apify-docs/pull/1579

### Done

- Fixes https://github.com/apify/apify-docs/issues/947
- Fixes https://github.com/apify/apify-docs/issues/1648

### Challenges

- Rewrite the lessons to use JS technologies, which often significantly differ.
    - replace httpx with fetch https://nodejs.org/en/learn/getting-started/fetch & https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
    - replace bs4 with cheerio https://cheerio.js.org/ because https://crawlee.dev/js/api/cheerio-crawler/class/CheerioCrawler
    - use https://github.com/juanjoDiaz/json2csv, the original package is abandoned
- See if the original JS course contains stuff which isn't in the Python course, but we want to keep it in some form.
    - lesson about Headless browsers - we could add it to both Python and JS course (after crawlee lesson?)
- Make sure we're not duplicating images etc.
- Make sure all historical links lead somewhere.

### Todo

- [x] rewrite extracting data in the Python course so that it still mentions Decimal, but also converts to cents (it's in saving data, too!) - only after https://github.com/apify/apify-docs/pull/1658
- [x] don't forget any /TBD/ occurrences 
- [x] resolve the temporary symlink to the images directory
- [x] eventually rename javascript2 to javascript
- [x] DRY intro of the course? DRY partials?
- [x] fix sidebar position of the intro
- [x] add redirects
- [x] missing semicolon const listingURL = "https://warehouse-theme-metal.myshopify.com/collections/sales"
- [x] use :::tip admonition for hints in exercises
- [x] fix: use abstract Markdown numbering


honzajavorek:
Closing in favor of https://github.com/apify/apify-docs/pull/1907

PR #1614, opened 2025-06-03
docs: Changed MCP image

vystrcild:


vystrcild:
@TC-MO Yeah, I put it inside ticks `` instead of link. Doesn't matter much.

jirispilka (changes requested):
Hi, thank you, the image is nice.

But we should not have there Apify Actors MCP Server. It should be `Apify MCP Server` or just `MCP Sever` (since it is already under the Apify "umbrella")

Also, MCP Hosts does not make much sense IMO.
I would use `Host with MCP client`.

vystrcild:
Changed picture:
- I kept "Apify" in MCP Server because it's a standalone solution and based on feedback from Adam Frankl, we should put Apify everywhere we could
- "Host with MCP client" is unnecessary complicated - just using "Clients" as everyone in the world in theirs schemas

jirispilka (approved):
Looks good to me, thanks!

PR #1615, opened 2025-06-03
chore: bump Docusaurus + React, use `yarn`

barjin:
Bumps Docusaurus to 3.8.0 and React to version 19. Switches the package manager to `yarn` because of faulty version resolution with `npm`.


B4nan:
@vladfrangu could you take a look locally too to see if we can stay on NPM? I'll try to do the same later today or tomorrow, feels weird that the module resolution wouldn't work like this. Maybe we just need to use `overrides` or add some dependency explicitly somewhere.

PR #1624, opened 2025-06-09
fix: adding a view param into dataset items

drobnikj:


fnesveda (approved):

apify-api/openapi/paths/datasets/datasets@{datasetId}@items.yaml
```diff
@@ -350,6 +350,15 @@ get:
       schema:
         type: boolean
         example: false
+    - name: view
+      in: query
+      description: |
```
janbuchar, line 355: I'd mention that `overview` is more or less the default view and anything else is up to the Actor developer.

PR #1642, opened 2025-06-18
feat: Add configurable search engine indexing behavior [internal]

fnesveda:
If the `NO_INDEX` environment variable is used during build, the [`noIndex` Docusaurus config](https://docusaurus.io/docs/api/docusaurus-config#noIndex) will be used and the docs will be built so that they're not indexable by search engines.

### DIFF
--- a/apify-docs-theme/src/config.js
+++ b/apify-docs-theme/src/config.js
@@ -10,7 +10,7 @@ if (process.env.LOCALHOST) {
     absoluteUrl = process.env.APIFY_DOCS_ABSOLUTE_URL;
 }
 
-const noIndex = ['true', '1'].includes(process.env.NO_INDEX ?? '')
+const noIndex = ['true', '1'].includes(process.env.NO_INDEX ?? '');
 
 const themeConfig = {
     docs: {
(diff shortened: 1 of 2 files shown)


apify-docs-theme/src/config.js
```diff
@@ -10,6 +10,11 @@ if (process.env.LOCALHOST) {
     absoluteUrl = process.env.APIFY_DOCS_ABSOLUTE_URL;
 }
 
+let noIndex = false;
+if (process.env.NO_INDEX) {
+    noIndex = [true, 'true', 1, '1'].includes(process.env.NO_INDEX);
```
B4nan, line 15: nit: i dont think you can get anything else than a string out of `process.env` vars

image

also the `if` block seems unnecessary, `const noIndex = ['true', '1'].includes(process.env.NO_INDEX ?? '');` will do the same on a single line (which can be inlined to the place where its used)

B4nan (approved):

apify-docs-theme/src/config.js
fnesveda: Good point, will clean up

PR #1690, opened 2025-07-15
docs: replace all api docs links

TC-MO:
replace all old api-docs link to current working URLs
fix prose

### DIFF
--- a/sources/academy/platform/getting_started/apify_api.md
+++ b/sources/academy/platform/getting_started/apify_api.md
@@ -11,7 +11,7 @@ slug: /getting-started/apify-api
 
 ---
 
-[Apify's API](/api/v2) web-interface. The API is organized around RESTful HTTP endpoints.
+[Apify's API](/api/v2) is your ticket to the Apify platform without even needing to access the [Apify Console](https://console.apify.com?asrc=developers_portal) web-interface. The API is organized around RESTful HTTP endpoints.
 
 In this lesson, we'll be learning how to use the Apify API to call an Actor and view its results. We'll be using the Actor we created in the previous lesson, so if you haven't already gotten that one set up, go ahead do that before moving forward if you'd like to follow along.
 


sources/academy/platform/getting_started/apify_api.md
```diff
@@ -11,7 +11,7 @@ slug: /getting-started/apify-api
 
 ---
 
-[Apify's API](/api/v2#/reference) is your ticket to the Apify platform without even needing to access the [Apify Console](https://console.apify.com?asrc=developers_portal) web-interface. The API is organized around RESTful HTTP endpoints.
+[Apify's API](/api/v2) web-interface. The API is organized around RESTful HTTP endpoints.
```
tomnosek, line 14: It seems there was an editing mistake here - removing up to the wrong parenthesis.

sources/academy/platform/running_a_web_server.md
```diff
@@ -236,4 +236,4 @@ When we deploy and run this Actor on the Apify platform, then we can open the **
 
 With that, we're done! And our application works like a charm :)
 
-The complete code of this Actor is available [here](https://apify.com/apify/example-web-server). You can run it there or copy it to your account.
+The complete code of this Actor is available at it's [page](https://apify.com/apify/example-web-server). You can run it there or copy it to your account.
```
tomnosek, line 239: It should be "on its", not "at it's". But also, saying "at its page" with a basic link is lazy writing 😄 Why not to say "on its Apify Store page" and link it directly to https://apify.com/apify/example-web-server/source-code?

tomnosek (changes requested):
I wasn't checking the exactness of the links, but I presume a link checker would find insufficiencies.

TC-MO:
Yeah, link checker should have caught anything that is not working. This is just fix to not use legacy URLs that seem to have broke and are not working correctly when trying to open them, so I swapped them with current URLs and by checking on preview it seems to point to correct endpoints and opens up correctly now , no matter whether you try to open it on new tab or within the same tab

tomnosek (approved):

PR #1699, opened 2025-07-18
docs: add n8n documentation

jakcinmarina:


TC-MO (changes requested):
Some changes requested to adjust the style to fit our style guide as well as for clarity & morecontext

B4nan:
The PR is from a fork, so the CI doesn't have the secrets. We could turn the validation error into a warning instead, I think I changed it this way for local development already.

B4nan:
It would only help with the build step, not with previews, those fail because of missing GH token (same problem, but I don't see any workarounds there, we can't expose secrets to forks).

TC-MO:
Ah gotchu. then we shall do without previews then.

sources/platform/integrations/workflows-and-notifications/n8n.md
```diff
@@ -0,0 +1,191 @@
...
+
+![Apify Install Node](../images/n8n-install-node-self-hosted.png)
+
+## Install the Apify Node (n8n Cloud)
+
+For n8n Cloud users, installation is even simpler and doesn't require manual package entry.  Just search and add the node from the canvas.
+
+1. Go to the **Canvas** and open the **nodes panel**
+2. Search for **Apify** in the community node registry
+3. Click **Install node** to add the Apify node to your instance
```
TC-MO, line 43:
```suggestion
1. Go to the **Canvas** and open the **nodes panel**
1. Search for **Apify** in the community node registry
1. Click **Install node** to add the Apify node to your instance
```

TC-MO (approved):
One last change, otherwise LGTM

PR #1729, opened 2025-07-25
docs: change security description within API docs

TC-MO:
merge apiKey & htttpBearer to one per
make sure all endpoints refer to correct security
adjust descriptions
add admonitions

### DIFF
--- a/apify-api/openapi/openapi.yaml
+++ b/apify-api/openapi/openapi.yaml
@@ -14,25 +14,22 @@ info:
     [JSON](http://www.json.org/) format with UTF-8 encoding,
     with a few exceptions that are explicitly described in the reference.
 
-    To access the API using [Node.js](https://nodejs.org/en/), we recommend the
-    [`apify-client`](https://docs.apify.com/api/client/js) [NPM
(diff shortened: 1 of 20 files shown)


apify-api/openapi/openapi.yaml
```diff
@@ -627,20 +624,8 @@ components:
...
-    httpBearerQueueId:
-      $ref: components/securitySchemes/httpBearerQueueId.yaml
     apiKey:
       $ref: components/securitySchemes/apiKey.yaml
-    apiKeyActorBuilds:
-      $ref: components/securitySchemes/apiKeyActorBuilds.yaml
-    apiKeyStoreId:
-      $ref: components/securitySchemes/apiKeyStoreId.yaml
-    apiKeyQueueId:
-      $ref: components/securitySchemes/apiKeyQueueId.yaml
```
fnesveda, line 643: I think these files are not referenced anywhere anymore, so you can delete them completely.

fnesveda (approved):
Looks good, just one note 🙂

apify-api/openapi/openapi.yaml
TC-MO: Yeah, good point, I left them on the off chance that we would like to go back to this way, but I don't think that will be the case. Will delete them then, thanks!

PR #1734, opened 2025-07-28
docs: Add quick start prompt to the docs

patrikbraborec:
The PR adds a button to copy a prompt for quick start. 

**Important**: 

- Only quick demo to open discussion. It would need probably little bit improve the prompt, add analytics to measure click rate, etc.

### DIFF
--- a/sources/platform/actors/development/quick_start/start_locally.md
+++ b/sources/platform/actors/development/quick_start/start_locally.md
@@ -45,9 +45,7 @@ The CLI will:
 
 Now, you can navigate to your new Actor directory:
 
-```bash
-cd your-actor-name
-```
+- tbd
 
 ### Step 2: Run your Actor
 
--- a/sources/platform/actors/development/quick_start/start_locally.md
+++ b/sources/platform/actors/development/quick_start/start_locally.md
@@ -9,6 +9,10 @@ slug: /actors/development/quick-start/locally
 
 ---
 
+import QuickStartPromptButton from "@site/src/components/QuickStartPromptButton";
+
+<QuickStartPromptButton/>
+
 ## What you'll learn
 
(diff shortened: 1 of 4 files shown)


patrikbraborec:
To be honest, I got the inspiration in the Clerk's doc as I quite liked it: https://clerk.com/docs/quickstarts/nextjs. 

If I understand your suggestion, you would show the whole prompt (for example with "accordion" - https://ui.shadcn.com/docs/components/accordion, or other UI element), let users read the prompt, and let them decide if they want to copy it, or not?

patrikbraborec:
Yeah, I think that is a good point. I will redesign the component.

patrikbraborec:
You can check it @TC-MO. Do you know, if it is possible to add analytics (i.e. to measure click events on the `copy prompt` button)?

sources/platform/actors/development/quick_start/start_locally.md
```diff
@@ -45,9 +49,7 @@ The CLI will:
 
 Now, you can navigate to your new Actor directory:
 
-```bash
-cd your-actor-name
-```
+- tbd
```
TC-MO, line 52: Is this supposed to be like this? 

image

TC-MO (commented):
Looks nice ,and thank you for adding the ability to see the prompt ;) 🚀

TC-MO:
As for the analytics, I think this is a request for @jbartadev

TC-MO (approved):
LGTM

patrikbraborec:
Thanks for the approve, @TC-MO. The next steps:

- I want to add event to measure clicks on copy prompts. The reason for that is that it can give us insights about usage of such a "feature".
- I want to also test the prompt in other environments, I only tested it briefly with Cursor. 

Therefore, I will wait with merge. :) Thanks!

PR #1736, opened 2025-07-29
feat: Enable llms for api & sdk docs

### LINKED ISSUE #1711: Have 100% of content covered by `/llms-full.txt`
We currently have content in another 5 different repositories for our docs (JS client, Python client, JS SDK, Python SDK, CLI), and we need to include that content in the `/llms-full.txt` for file to be useful.

HonzaTuron:
Closes #1711

- Enable merging of llms across all docusaurus plugins
- Add markdown link to head
- Setup route categories


B4nan (approved):
looks good, but we might run this periodically, or trigger it when the child repositories like SDKs change, now the file would be regenerated only when this repository changes

PR #1752, opened 2025-08-04
docs: Fix titles based on discussion

patrikbraborec:
This is a follow up for the PR: https://github.com/apify/apify-docs/pull/1751. Some titles need to improve as I did not wait for @TC-MO - sorry!

### DIFF
--- a/sources/platform/actors/development/index.md
+++ b/sources/platform/actors/development/index.md
@@ -1,6 +1,7 @@
 ---
 title: Actor development
 desc: Read about the technical part of building Apify Actors. Learn to define Actor inputs, build new versions, persist Actor state, and choose base Docker images.
+sidebar_label: Development
 sidebar_position: 7.4
 slug: /actors/development
 ---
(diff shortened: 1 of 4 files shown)


sources/platform/actors/development/index.md
```diff
@@ -5,8 +5,6 @@ sidebar_position: 7.4
 slug: /actors/development
```
TC-MO, line 5: Can we add `sidebar_label: Development` to frontmatter? I don't think placement matter much, but for consistency I would prefer under description

TC-MO (changes requested):
One change requested, IIRC we spoke in the previous thread about keeping sidebar label unchanged and changing just the page title

sources/platform/actors/development/index.md
TC-MO: And based on [this discussion](https://github.com/apify/apify-docs/pull/1751#discussion_r2250849010) can we get `Account settings` instead of `Console settings`

PR #1753, opened 2025-08-04
docs: change Console settings to Account settings

TC-MO:


sources/platform/console/settings.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Console settings
+title: Account settings
```
patrikbraborec, line 2: nit: in the text below, we talk about `Apify Console settings`, do you think that is ok, or should we align it with the title?

TC-MO: In the end went with different approach

patrikbraborec (approved):

sources/platform/console/settings.md
patrikbraborec: Good 👍

patrikbraborec:
LGTM

PR #1791, opened 2025-08-12
docs: add cursor rules

TC-MO:


TC-MO:
I guess file organization might be on the chopping block? Not sure how usable it will be, especially if we are thinking about IA rewrites soon.

PR #1792, opened 2025-08-12
feat: add separate `Help` tab for `help.apify.com` results in Inkeep

barjin:
As per request of @TC-MO , Help articles are now not intermingled with Docs anymore (and are only accessible through the `All` tab).


TC-MO (approved):
LGTM, thanks!

PR #1798, opened 2025-08-14
fix: rename eventCount to count

mhamas:
The correct field in the body of the request is called count.


TC-MO (approved):
LGTM ( I could have sweared that I already fixed that, thanks!)

janbuchar (approved):

fnesveda (approved):

patrikbraborec (approved):

PR #1801, opened 2025-08-15
fix: Always clear cache when on MCP page to help with redirect issue

jirispilka:
Always clear cache when on MCP page to help with redirect issues.
Previously, mcp.apify.com had a 301 redirect to this docs page, which was wrong.
Many users have this redirect cached in their browsers, so they can't access the new MCP UI.
This helps with their cached redirect, allowing them to access mcp.apify.com

@B4nan is there is a better place to put the code than in the src/theme/layout? 

Kudos to @JJetmar 

FYI @MQ37 

close https://github.com/apify/apify-mcp-server/issues/129
close https://github.com/apify/ai-team/issues/33

EDIT: Remove this fix after some time: https://github.com/apify/apify-docs/issues/1802


TC-MO (approved):
LGTM (tho not that great at React :D)

jirispilka:
@B4nan thanks. I've moved to `custom.js`, make sense. 

If you can, please give a quick look and we can release it. 
I've tested it at local, it works

jirispilka:
Let's merge this and see: `#yolo`

PR #1803, opened 2025-08-15
docs: Add copy for LLMs and view as markdown buttons

patrikbraborec:
A draft for "LLM buttons" in our docs.

It allows you to:

- Copy for LLM - button that copies the whole page as markdown to your clipboard
- View as Markdown - button that opens a new tab with the current page as a markdown

I call it a draft because it needs to be tested more properly as I edited layout files, and it would be also good to add to ALL docs pages (so far it works only on the academy and platform pages).


patrikbraborec:
Sorry @mnmkng, Github added you automatically as a reviewer, but I discussed this thing mainly with @TC-MO. Anyway, you are right that it should have proper description, and I added it. >]

fnesveda (changes requested):
Also, one more thing - I noticed that now the markdown files contain some weird characters that they didn't have before, e.g. here:
https://pr-1803.preview.docs.apify.com/platform/actors.md

Screenshot 2025-08-20 at 16 40 34

fnesveda (approved):

PR #1817, opened 2025-08-20
docs: upgrades for privilege-less Docker images

vladfrangu:
Related to https://github.com/apify/apify-actor-docker/pull/188

https://pr-1817.preview.docs.apify.com/platform/actors/development/docker-user-changes


vladfrangu:
I don't know exactly where to add it in considering its half changelog half "hey btw changes to docker images happened"

vladfrangu:
Do them as you wish, we can discuss tomorrow! But keep in mind I wanted to make it a separate page, and I'm not sure if cramming it into the existing page is smart 😅

sources/platform/actors/development/actor_definition/docker.md
```diff
@@ -112,3 +112,109 @@ This means the system expects the source code to be in `main.js` by default. If
...
 
 :::
+
+## Updating older Dockerfiles
+
+Certain Apify base Docker images now use a non-root user to enhance security. This change requires updates to existing Actor `Dockerfile`s that use the `apify/actor-node`, `apify/actor-python`, `apify/actor-python-playwright`, or `apify/actor-python-selenium` images. This section provides guidance on resolving common issues that may arise during this migration.
+
+:::danger Action required
+
+The base Docker images display a deprecation warning. This warning will be removed in future versions, so you should update your Dockerfiles to ensure forward compatibility.
```
vladfrangu, line 122: Its not a future version, I'll be dropping the warning in Q1 2026!

sources/platform/actors/development/actor_definition/docker.md
```diff
@@ -112,3 +112,109 @@ This means the system expects the source code to be in `main.js` by default. If
 You can check out various optimization tips for Dockerfile in our [Performance](../performance.md) documentation.
 
 :::
+
+## Updating older Dockerfiles
+
+Certain Apify base Docker images now use a non-root user to enhance security. This change requires updates to existing Actor `Dockerfile`s that use the `apify/actor-node`, `apify/actor-python`, `apify/actor-python-playwright`, or `apify/actor-python-selenium` images. This section provides guidance on resolving common issues that may arise during this migration.
```
vladfrangu, line 118:
```suggestion
All Apify base Docker images now use a non-root user to enhance security. This change requires updates to existing Actor `Dockerfile`s that use the `apify/actor-node`, `apify/actor-python`, `apify/actor-python-playwright`, or `apify/actor-python-selenium` images. This section provides guidance on resolving common issues that may arise during this migration.
```

sources/platform/actors/development/actor_definition/docker.md
```diff
@@ -112,3 +112,109 @@ This means the system expects the source code to be in `main.js` by default. If
...
+
+#### `uv` package manager fails to install dependencies
+
+If you are using the `uv` package manager, you might have the following line in your `Dockerfile`:
+
+```dockerfile
+ENV UV_PROJECT_ENVIRONMENT="/usr/local"
+```
+
+With the move to a non-root user, this variable will cause `uv` to throw a permission error. You can safely remove this line or adjust it to point to the `/home/myuser` directory.
```
vladfrangu, line 152:
```suggestion
With the move to a non-root user, this variable will cause `uv` to throw a permission error. You can safely remove this line, or, if you need it set to a custom path, adjust it to point to a location in the `/home/myuser` directory.
```

vladfrangu (commented):
@TC-MO not sure how I feel about having this as crammed into the main dockerfile document but 🤷 see what @patrikbraborec and @B4nan think about it.

But these small changes I'd want applied, thoughts?

patrikbraborec:
I think it is better to have with existing docker docs. I checked it, and it looks good imho. If it gets complicated, we can always separate it, but I would definitely keep it in `Actor definition` section.

sources/platform/actors/development/actor_definition/docker.md
TC-MO: I tried to avoid anchoring in time, but I think it might be unavoidable, we can add specific 
```suggestion
The base Docker images display a deprecation warning. This warning will be removed in Q1 2026, so you should update your Dockerfiles to ensure forward compatibility.
```

And then make a note in calendar to remove the whole admonition after we drop the warning. Would that work for you?

B4nan:
Same here, I think it's fine on a single page.

Would be good to mention a date of the change too, so people can understand what "older dockerfile" means in this context.

sources/platform/actors/development/actor_definition/docker.md
vladfrangu: Sounds good

vladfrangu:
> I'll add specific time (Q1 2026) to the admonition, tho if we can settle on specific date @vladfrangu that would be preferable, then we can setup reminder to remove it past that date

I want to do the actual release Monday, sooo... I guess thats the date?

TC-MO:
Oh and I meant the day we drop the warning :D For the date of change we can reprhase the warning to state:

:::danger Action required

As of <here the date> the base Docker images display a deprecation warning. This warning will be removed in Q1 2026 (or specific date), so you should update your Dockerfiles to ensure forward compatibility.

:::

TC-MO (approved):

PR #1839, opened 2025-08-25
chore: skip fragment link checks with lychee

barjin:
Recent `lychee` updates (`0.19.0` and above) improve Lychee and align it more with the HTML spec. This, however, also means that Lychee now finds 2000+ "broken" links due to incorrect setup in this repo. 

Since fragment links inside of docs.apify.com are checked internally with Docusaurus anyway, this PR disables the fragment link checker in Lychee.


barjin:
Example run here:

https://github.com/apify/apify-docs/actions/runs/17211763113/job/48825140318

imo all the reported errors are relevant.

TC-MO (approved):
LGTM thank you, I'll handle all the relevant errors

PR #1887, opened 2025-09-04
feat: use the tip admonition for exercise hints

honzajavorek:
7gbw0x-1966945061


TC-MO (commented):
I'm not sure if `Hint` is the best title. I know it would be a lot of work to title each admonition separately but it doesn't look great to me

image

Thoughts?

## EDIT ##

I can just foresee them skipping over it easily 🤔

PR #1894, opened 2025-09-05
feat: General Resource Access

raethlo:
This PR adds documentation for General Resource Access, new account and resource level setting that enables to fine tune resource access when using resource ids or links.


raethlo:
Going over the preview myself:

- I think this screenshot might be a bit too large 👀 https://pr-1894.preview.docs.apify.com/platform/collaboration/general-resource-access#exception-automatically-sharing-runs-via-actor-issues
- formatting `unrestricted` vs. **unrestricted**
- maybe the special case of named storages and **Anyone with id or name can read** could go to the summary instead of where it is
- The how it works section seems thin after re-reading, do you think we should expand it?

PR #1903, opened 2025-09-08
chore: ignore docs quality checks for legal documents

honzajavorek:
Extracted from https://github.com/apify/apify-docs/pull/1890/ Context is explained by @TC-MO in this comment: https://github.com/apify/apify-docs/pull/1890#issuecomment-3258243768

> @mnmkng context for you, in `/legal` I moved `<!-- vale off --> comment before frontmantter to not have the new rule targeting H1's affect anything. Also I've added ` I don't think you were using it and it was throwing a lot of false positives when working locally on docs repo. But if you are indeed using it, we can revert this change easily.
> 
>  * The goal here is it create new rule for Vale targeting H1's to prevent issues with the LLM buttons 
> * It also led to issues with the custom rules I've created previously so this PR should fix that
> * I also took this opporunity to update the packages for `write-good` & `msoft style guide` that we use in docs


TC-MO (approved):
LGTM, but let's wait for @mnmkng approval as well

B4nan:
This likely broken all the legal pages, I dont think you can put anything above the frontmatter.

The [legal page](https://docs.apify.com/legal) now looks like this:

image

PR #1916, opened 2025-09-10
fix: mix of small improvments

JMatej:


TC-MO:
I can take a look later today and fix vale issues, from what I saw it is just screaming about H1's

TC-MO:
Just FYI @JMatej in this case it is actually better to remove H1's instead of turning them into H1's. 

Docusaurus will create H1 automatically based on `title` property from frontmatter so we will have now in those files H1 generated from title and immediately below H2 that is exactly the same.

PR #1925, opened 2025-09-15
fix: bubble integration doc

daveomri:
Fixed missing code block closure in Bubble integration documentation

This PR fixes a formatting issue in the Bubble integration documentation where a JSON code block was not properly closed. The missing closing code block marker (```) has been added after line 109, ensuring the proper rendering of the documentation and preventing subsequent content from being incorrectly formatted as code.

This minor adjustment makes the rest of the doc more readable and follows the Markdown format.


TC-MO (approved):
Thanks, good catch, shame we can't have previews on these PR's enabled (hopefully for now)

PR #1931, opened 2025-09-17
docs: Add templates link to the academy

patrikbraborec:


TC-MO:
I'm finishing up work on removing this section and deploying section completely from Academy, not sure if it's worth making any changes ot it. Around ~85% of content is already covered in Platform docs and what is not will be added there from here, while this sections are revoved completely

PR #1935, opened 2025-09-18
fix: nginx configuration; add nginx linter

lukasmrtvy:


lukasmrtvy:
Fixed.

The last deployment ( https://github.com/apify/apify-docs-private/actions/runs/17827556817/job/50684203179 ) failed because of the wrong configuration introduced in https://github.com/apify/apify-docs/pull/1933, thus the validation workflow :)

PR #1957, opened 2025-09-30
feat(platform): IA rework

TC-MO:
- Create 10 numbered directories for new IA
- Add _category_.yml for sidebar ordering
- Add minimal placeholder index files (to be replaced)

Part of Platform IA reorganization

### DIFF
--- a/sources/academy/glossary/tools/apify_cli.md
+++ b/sources/academy/glossary/tools/apify_cli.md
@@ -15,7 +15,7 @@ The [Apify CLI](/cli) helps you create, develop, build and run Apify Actors, and
 
 ## Installing {#installing}
 
-To install the Apfiy CLI, you'll first need npm, which comes preinstalled with Node.js. If you haven't yet installed Node, learn how to do that [here](../../webscraping/scraping_basics_javascript/data_extraction/computer_preparation.md). Additionally, make sure you've got an Apify account, as you will need to log in to the CLI to gain access to its full potential.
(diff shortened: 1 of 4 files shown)


TC-MO:
I wanted to create a preview so people could click through the proposed changes for IA but then I got stuck in a constant loop of:

1. Fix links
2. Wake up Vale with new issues

ad nausem.

I might try to split this PR into some more manageable chunks instead of this monstrosity, but I'll leave this as draft for nwo as reference. Apart from that best I can do is I guess add screenshot of how the new IA looks, and then maybe you can infer something from the changed files (tho it is overwhelming since I was moving all the files around)

TC-MO:
The idea behind this IA is to try and address few issues at the same time. 

1. Make `Getting started`/`Quick start` much more prominent to guide users there if they want an absolute TL:DR on creating their first Actor.
2. Split docs for:
	* Actor users - thus Console docs & whole Using Actors section where we explain concepts like
		* Store
		* Inputs & outputs
		* Runs and builds
		* Usage and resources
		* etc.
	* Actor builders (develoeprs)
		* Docs where we explain more technical aspects like
			*  What does Actors consist of
			* `actor.json`
			* Input. Dataset & KV store schema (and in the future Output schema)
			* Deployment
			* etc.
		* Publishing & monetization

The rest of docs stays as is current.

So this would be what would greet users. `Getting started` is deliberately by default uncollapsed so users can see everything. (might change the name to `Quick start`)

image

Console section
image

Using Actors
image

Core concepts (not completely sold on this name, or this division for that matter)
image

Building Actors section
image
image

Publishing & monetization
image

Integrations - as currently

Collaboration - as currently

PR #1965, opened 2025-10-02
feat: move restartOnError to run options

DaveHanns:
Documentation adjustments according to changes based on https://app.zenhub.com/workspaces/platform-team-5f6454160d9f82000fa6733f/issues/gh/apify/apify-core/

Core: https://github.com/apify/apify-core/pull/23381
Worker: https://github.com/apify/apify-worker/pull/1515
Shared: https://github.com/apify/apify-shared-js/pull/546
JS Client: https://github.com/apify/apify-client-js/pull/760
Py Client: https://github.com/apify/apify-client-python/pull/508


apify-api/openapi/components/schemas/actors/CreateActorRequest.yaml
```diff
@@ -46,6 +42,5 @@ properties:
     description: ''
     nullable: true
   defaultRunOptions:
-    oneOf:
```
DaveHanns, line 49: Question: @B4nan, why do we declare the nullable fields in some cases like this ?
It results in a bit weird notation in docs, see screenshot.

image

janbuchar: What is the actual type accepted by the API? Is it "`DefaultRunOptions` or `null`"?

I'm asking so that I can check out if there's a better way to express this. But you're right that this looks suspicious.

janbuchar: Cool! Thanks for investigating :slightly_smiling_face:

PR #1973, opened 2025-10-04
docs(mcp): add note regarding the dynamic tool discovery logic based on client capabilities

MQ37:


TC-MO:
We definitely cannot have two admonitions next to each other, it visually way to distracting and just not a good practice. But I think I may have a solution (tho a little more involved)

What are your thoughts about doing a slight reshuffle of the order of headings. I would propose the following

### Available tools 

Table

`get-actor-output` admonitions

#### Dynamic tool discovery (as it to me feels like a extension of all available tools)

All of it's content

Dynamic tool discovery admonition


Thoughts @jirispilka @MQ37 ?

jirispilka:
@TC-MO thanks for the suggestion. I decide to fix it right away based on it. Please check it out. Thanks

TC-MO (approved):
LGTM

PR #1989, opened 2025-10-08
fix: Remove Claude integration

birosrichard:


package-lock.json
katzino, line 1: Is this updated on purpose?

katzino (approved):

PR #1995, opened 2025-10-10
docs: reorganize monetization section

TC-MO:
Goal is to de emphasize rental and emphasize PPE more


patrikbraborec:
Agree, lets merge it.

PR #2027, opened 2025-10-17
chore: lint Python files

honzajavorek:
Just a PoC related to https://github.com/apify/apify-docs/issues/1243 Issues:

- Many, many files have Python lint issues. For now I have the GitHub Action work just on top of changed files, but `npm run lint:code:py` can give you an idea. Actually, it cannot, because it stops on the first error: https://github.com/adamtheturtle/doccmd/issues/569
- In tutorials, code blocks are incomplete and follow up on each other. The `doccmd` tool supports [grouping them](https://adamtheturtle.github.io/doccmd/group-code-blocks.html), but that requires explicit marks in the Markdown file. Doable, but a lot of manual work to get the linter passing. OTOH, sometimes giving the reader more context, like repeating imports, etc., is actually desirable for better understanding. See [apify_client.md](https://github.com/apify/apify-docs/pull/2027/commits/263f807d223f55794884ba9e87c3c9f5c9c50a71) as a showcase.
- The syntax with `title="filename.py"` isn't supported as of now: https://github.com/simplistix/sybil/issues/155
- I'm unsure how it's gonna work with `.mdx`. For now I think it ignores it unless we explicitly tell it to look after those files. I don't think there's much Python code in `.mdx` files, though… (my unverified assumption)
- For now I didn't include a `fix` option. The Python linter, ruff, does have a `--fix` option, but I'm unsure we want anyone to run it without thinking. Most of the _unused import_ errors shouldn't be auto-fixed as _let's remove the import_, usually it's the case that it's a tutorial with several code blocks following up, and the correct solution is to [group them](https://adamtheturtle.github.io/doccmd/group-code-blocks.html).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -186,7 +186,7 @@ jobs:
                 run: |
                     IFS=',' read -ra FILE_ARRAY <<< "$ALL_CHANGED_FILES"
                     for file in "${FILE_ARRAY[@]}"; do
-                      uv run --with doccmd --with ruff doccmd --language=py --language=python --command=\"ruff check --quiet\" "$file"
+                      uv run --with doccmd --with ruff doccmd --language=py --language=python --command="ruff check --quiet" "$file"
                     done
 
             # TODO: once we fix existing issues, switch to checking the whole sources/ directory


.github/workflows/test.yaml
```diff
@@ -7,10 +7,10 @@ on:
 
 jobs:
     build:
-        name: Docs build
+        name: Docs Build
```
B4nan: why did you rename all the jobs? :D i dont think we want title case (or whatever is this style called) in here, lets not touch unrelated parts

honzajavorek: Sorry, OCD. It was random, some names were title cased and some not. I felt like most are title cased, so I unified it, but I have no other opinions about it and I'm fine making the effort to restrain myself to touch those parts 😅

adamtheturtle:
Very cool to see this @honzajavorek ! I'm happy to help you get onboarded to `doccmd`. For my own projects, it has been a lot of work but it pays dividends regularly.

> Many, many files have Python lint issues. For now I have the GitHub Action work just on top of changed files, but npm run lint:code:py can give you an idea. Actually, it cannot, because it stops on the first error: https://github.com/adamtheturtle/doccmd/issues/569

You can update `doccmd` and use `--continue-on-error`.

In my own projects, I instead [used ignores for each linter](https://adamtheturtle.github.io/doccmd/file-names-and-linter-ignores.html), and then just slowly removed ignores.

> In tutorials, code blocks are incomplete and follow up on each other. The doccmd tool supports [grouping them](https://adamtheturtle.github.io/doccmd/group-code-blocks.html), but that requires explicit marks in the Markdown file. Doable, but a lot of manual work to get the linter passing. OTOH, sometimes giving the reader more context, like repeating imports, etc., is actually desirable for better understanding. See [apify_client.md](https://github.com/apify/apify-docs/pull/2027/commits/263f807d223f55794884ba9e87c3c9f5c9c50a71) as a showcase.

I'm open to ideas on how to make this better in `doccmd`.

> I'm unsure how it's gonna work with .mdx. For now I think it ignores it unless we explicitly tell it to look after those files.

That's right. You can use [`--markdown-extension`](https://adamtheturtle.github.io/doccmd/commands.html#cmdoption-doccmd-markdown-extension) to tell `doccmd` which extensions to use to look for Markdown files.

adamtheturtle:
@honzajavorek I believe that every one of the issues you mentioned has been resolved in `doccmd`, as of [doccmd 2025.12.5](https://pypi.org/project/doccmd/2025.12.5/).

I'm happy to be of help if you want to move this PR forward.

I also plan to look at https://github.com/adamtheturtle/doccmd/issues/574.

adamtheturtle:
See https://github.com/apify/apify-docs/pull/2128 which includes some language changes and doccmd groups so that many of these files can be linted.

Personally I have found the biggest value of `doccmd` to be `mypy` which can show that the documented APIs are incorrect.

PR #2037, opened 2025-10-22
docs: Dify integration docs

mortyyy1:
This is almost ready part of Dify documentation only without OAuth part. It's still in process...


sources/platform/integrations/workflows-and-notifications/dify.md
```diff
@@ -0,0 +1,183 @@
...
+
+![Chatflow Payload Template](../images/dify-define-input-variable.png)
+
+![Chatflow Payload Template](../images/dify-workflow-with-input-variable.png)
+
+We can also use that variable inside a JSON object we send and see the results.
+
+![Chatflow Payload Template](../images/dify-input-variable-webhook.png)
+
+![Chatflow Payload Template](../images/dify-input-variable-webhook-succeded.png)
```
TC-MO:
```suggestion
### Access Apify Data in Dify

When an event in Apify triggers your Dify endpoint, Apify sends a detailed [JSON response object](https://docs.apify.com/api/v2/act-runs-post#responses) containing information about the event. You can use this data to power your workflow.

Dify cannot directly use nested syntax like `resource.id` in its variable system. Instead, access nested properties using a flattened format (`resource__id`), where each level of nesting is represented by a double underscore (`__`).

To work with a nested property in your workflow, define it as an input variable in your workflow's Start node by selecting **+**.

![Chatflow Payload Template](../images/dify-workflow-input-variable.png)

![Chatflow Payload Template](../images/dify-define-input-variable.png)

![Chatflow Payload Template](../images/dify-workflow-with-input-variable.png)

You can also use that variable inside a JSON object you send and view the results.

![Chatflow Payload Template](../images/dify-input-variable-webhook.png)

![Chatflow Payload Template](../images/dify-input-variable-webhook-succeded.png)

```

TC-MO (changes requested):
Rephrased a lot of lists to prose. We should not rely as heavily on unorderd or ordered lists to explain workflows to users.
Fixed admonition to use docusaurus syntax
Added title to code example

Apart from that few questions about the example ID's used & question regarding the `Troubleshooting` section

PR #2038, opened 2025-10-23
feat: Add Apify MCP server to the docs front page

jirispilka:
know I probably should’ve asked first, but since it was a small change, I decided to open a PR right away and see how it goes. ʸᵒˡᵒ  😊 

Changes:
- add MPC to the front docs page
- add link to Open Source -> Apify MCP server

A bit of context: https://apify.slack.com/archives/C09AUCEABDM/p1759954446027269

FYI @lukas-bekr

### DIFF
--- a/apify-docs-theme/src/config.js
+++ b/apify-docs-theme/src/config.js
@@ -123,10 +123,6 @@ const themeConfig = {
                         href: 'https://github.com/apify',
                     },
                     {
-                       label: 'Apify MCP Server',
-                       href: 'https://github.com/apify/apify-mcp-server',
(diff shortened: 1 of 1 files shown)


TC-MO:
Good initiative, I agree with adding it to the front page, not sure about link in open-source category 🤔

B4nan (approved):
codewise looks good, i'll leave it up to you guys if it should be on both places, no strong opinions.

jirispilka:
Thanks guys, we discussed this with @TC-MO over slack, conclusion was to remove the open source link! Done

PR #2039, opened 2025-10-23
docs: Update resource input field docs

mfori:
Mainly:
- say that the value can be also name not just ID
- add `textfield` editor and validation properties for single value
- split properties tables for single/multiple values for better clarity


mfori:
> but is it in Console already?

Not yet, I'll wait with merge until it's released of course

PR #2050, opened 2025-10-30
fix: Remove output schema to dataset schema rewrites

B4nan:


nginx.conf
```diff
@@ -533,10 +533,6 @@ server {
     rewrite ^/platform/actors/development/testing-and-maintenance$ /platform/actors/development/deployment/automated-tests redirect;
     rewrite ^/platform/actors/development/source-code$             /platform/actors/development/deployment/source-types redirect;
 
-    # Rename output schema to dataset schema
-    rewrite ^/platform/actors/development/actor-definition/output-schema$ /platform/actors/development/actor-definition/dataset-schema permanent;
-    rewrite ^academy/deploying-your-code/output-schema$ /academy/deploying-your-code/dataset-schema permanent;
```
valekjo: Question: Was this even correct? It's missing the leading `/` (and so do many other academy links)

PR #2167, opened 2026-01-09
fix: add missing docs for attachment param for KV getRecord

mvolfik:
> [!NOTE]
> Updates OpenAPI docs for Key-value stores `getRecord`.
> 
> - Adds optional boolean `attachment` query parameter; when `true`/`1`, responses include `Content-Disposition: attachment` so browsers download HTML records instead of displaying them
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d7695485320b03955b49c07798324ece5f680420. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


mvolfik:
hmm should I also add a note that says we don't add the cookie snippet (see apify/apify-core#24758) when the flag is enabled? I.e. with this flag, responses are guaranteed to be unmodified?

fnesveda (approved):

Issue #2182, opened 2026-01-15
Ensure OpenAPI specs can serve as a reliable source of truth for typed client generation


vdusek:
Fix and consolidate the existing OpenAPI specifications to reflect the actual API behavior, ensuring they can serve as a source of truth for typed API client generation.

PR #2187, opened 2026-01-19
ci(openapi): Add spell-check CI step and fix existing typos

vdusek:
- Configure `typos` spell checker to check the repository.
- Add `typos.toml` config to ignore false positives.
- Fix all current typos.
- Relates: #2182

### DIFF
--- a/typos.toml
+++ b/typos.toml
@@ -1,26 +1,40 @@
 # Configuration for typos spell checker
-# These are example IDs and should not be flagged as typos
+# https://github.com/crate-ci/typos
 
+[default]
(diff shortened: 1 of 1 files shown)


janbuchar (approved):

TC-MO:
I'm fairly certain that we used this already before but retired it for one reason or another (not 100% sure atm). At the same time maybe the tool is better now or we could have better rules in place. I see no harm in setting this up for whole repo and trying it out. Worst case scenario we can revert to having it only work on `apify-api`

PR #2249, opened 2026-02-13
docs: add missing title property to webhook API schemas

### LINKED ISSUE #2035: Missing `title` property in "Create webhook" request body
The **Create webhook** API request body documentation does not include the `title` property, even though the Apify API supports it.

### Example
The following request works correctly:

```json
POST /v2/webhooks
{
  "eventTypes": ["ACTOR.RUN.SUCCEEDED"],
  "requestUrl": "https://example.com/webhook-handler",
  "title": "Actor run succeeded webhook",
  ....
}
```

But title is not listed in the docs.
It is probably also missing in other wehbook related endpoints.

marcel-rbro:
## Summary
- Adds the missing `title` property to `WebhookCreate`, `WebhookUpdate`, and `Webhook` OpenAPI schemas
- The Apify API supports a `title` field on webhooks, but it was not documented in the OpenAPI spec

Closes #2035

## Test plan
- [ ] Verify the OpenAPI spec still validates correctly
- [ ] Check that the webhook create/update/get endpoints show the `title` property in rendered API docs

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apify-api/openapi/components/schemas/webhooks/Webhook.yaml
+++ b/apify-api/openapi/components/schemas/webhooks/Webhook.yaml
@@ -54,6 +54,9 @@ properties:
   headersTemplate:
     type: [string, "null"]
     examples: ['{\n "Authorization": "Bearer ..."}']
+  title:
+    type: [string, "null"]
+    examples: [Actor run succeeded webhook]
   description:
     type: [string, "null"]
     examples: [this is webhook description]
--- a/apify-api/openapi/components/schemas/webhooks/WebhookCreate.yaml
+++ b/apify-api/openapi/components/schemas/webhooks/WebhookCreate.yaml
@@ -35,6 +35,9 @@ properties:
   headersTemplate:
     type: [string, "null"]
     examples: ['{\n "Authorization": "Bearer ..."}']
+  title:
+    type: [string, "null"]
+    examples: [Actor run succeeded webhook]
   description:
     type: [string, "null"]
     examples: [this is webhook description]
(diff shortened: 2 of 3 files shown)


marcel-rbro (commented):
## Review: LGTM

The change correctly adds the missing `title` property to all three relevant webhook OpenAPI schemas.

### What looks good

1. **Pattern consistency**: The `title` property definition (`type: [string, "null"]` with `examples`) exactly matches the pattern used by the adjacent `description` property in all three files. Good adherence to existing conventions.

2. **Cross-schema consistency**: The definition is identical across `Webhook.yaml`, `WebhookCreate.yaml`, and `WebhookUpdate.yaml`, which is correct since they should all support the same `title` field.

3. **Logical placement**: `title` is placed immediately before `description` in each schema, which is a natural grouping.

4. **Example value**: `Actor run succeeded webhook` is appropriate - it matches the `ACTOR.RUN.SUCCEEDED` event type used in other examples in the same files, making the schema self-consistent.

5. **All CI checks pass**: Lint, spec validation, bundled spec build, docs build, and spell check all green.

### Considered but not an issue

I checked whether `WebhookShort.yaml` (used by the list webhooks endpoint via `ListOfWebhooks.yaml`) also needs `title`. It does not - `WebhookShort` already omits `description`, `payloadTemplate`, and `headersTemplate`, so omitting `title` is consistent with the "short" representation pattern.

No changes requested. This is a clean, minimal fix that addresses issue #2035.

PR #2263, opened 2026-02-19
docs: add PPE event charging examples and restructure best practices

TC-MO:
Add three generic `Actor.charge()` examples (JS + Python) to the PPE docs:
- Charge per result with spending limit check
- Charge for multiple event types
- Charge for multiple items at once using count parameter

Move "Respect user spending limits" from standalone section into best practices and reorder subsections for logical flow: setup, spending limits, charging patterns, then pricing guidance. Clean up best practices intro prose.

### DIFF
--- a/sources/platform/actors/publishing/monetize/pay_per_event.mdx
+++ b/sources/platform/actors/publishing/monetize/pay_per_event.mdx
@@ -203,114 +203,6 @@ If you're not using the Apify SDKs (JS/Python), you need to handle idempotency (
 
 The following examples show common charging patterns using the Apify SDK. Use the [JS](/sdk/js/reference/class/Actor#charge) or [Python](/sdk/python/reference/class/Actor#charge) SDK to handle idempotency, API errors, and spending limits automatically.
 
-### Respect spending limits
-
-Check `eventChargeLimitReached` after each charge and stop the Actor when the user's budget is exhausted. For Actors with multiple event types, use `chargeableWithinLimit` to check whether any event can still be charged before stopping.
-
-<Tabs groupId="main">
-<TabItem value="JavaScript" label="JavaScript">
-
-```js
-import { Actor } from 'apify';
-
-await Actor.init();
-
-const input = await Actor.getInput();
-const { items } = input;
-
-for (const item of items) {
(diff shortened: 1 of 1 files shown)


sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -57,178 +57,242 @@ An Actor's negative net profit does not affect the positive profit of another Ac
...
 ```
 
 </TabItem>
 </Tabs>
 
-:::note Crawlee integration and spending limits
+### Charge for multiple event types
 
-When using [Crawlee](https://crawlee.dev/), use `crawler.autoscaledPool.abort()` instead of `Actor.exit()` to gracefully finish the crawler and allow the rest of your code to process normally.
+Charge multiple event types in a single operation. Each event type must be defined in your Actor's pricing configuration.
```
marcel-rbro: Maybe explain this more thoroughly? The text says "charge multiple event types in a single operation", but the example shows two `charge` calls. Maybe rewrite to something like: "Charge for multiple different event types in one Actor run"?

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -57,178 +57,242 @@ An Actor's negative net profit does not affect the positive profit of another Ac
...
+const data = { url: 'https://example.com', title: 'Example' };
+const chargeResult = await Actor.charge({ eventName: 'result' });
 
 if (chargeResult.eventChargeLimitReached) {
-  await Actor.exit();
+    await Actor.exit();
 }
 
-// Rest of the Actor logic
+await Actor.pushData(data);
```
Patai5: Just from a coding standards perspective, it is recommended to keep variables as close as possible to where they are first used. So we could initialize the data right before pushing it 👀 


```suggestion
const chargeResult = await Actor.charge({ eventName: 'result' });

if (chargeResult.eventChargeLimitReached) {
    await Actor.exit();
}

const data = { url: 'https://example.com', title: 'Example' };
await Actor.pushData(data);
```

(this applies to the other examples as well)

PR #2306, opened 2026-03-03
docs(openapi): Use more autogenerated response examples

Pijukatel:
## Summary

- Delete some explicit examples and use auto-generated examples from the schema.
- Keep explicit examples where it helps:
  - lists, where an autogenerated example would contain just one item
  - specific non-default values. For example, aborted runs will have some specific run status, so returning an autogenerated example run could be confusing 

## Motivation
- Remove massive code duplication
- Preparation for adding undocumented redispatch endpoints, which would otherwise have to repeat the examples again, making the duplication even worse.
- Manual examples are missing newly added optional fields

**Example:**
image

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 2fe84fefe36f1c74507681f5996abf4399d05061. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


TC-MO (approved):
LGTM

PR #2400, opened 2026-04-07
docs: openclaw plugin docs

protoss70:
Adding newly released OpenClaw plugin documentation. 

Here is the original repo: https://github.com/apify/apify-openclaw-plugin

### DIFF
--- a/sources/platform/integrations/ai/openclaw.md
+++ b/sources/platform/integrations/ai/openclaw.md
@@ -95,11 +95,10 @@ The `apify` tool must be listed in `tools.alsoAllow` in your config, or you can 
 
 ### Verify the setup
 
-Check your configuration and test the connection:
+Check your configuration:
 
 ```bash
 openclaw apify status
-openclaw apify test
 ```
 
 ## How the Apify tool works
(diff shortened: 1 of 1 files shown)


sources/platform/integrations/ai/openclaw.md
```diff
@@ -0,0 +1,178 @@
...
+
+The tool uses an asynchronous workflow. Your agent starts an Actor run and gets a `runId` immediately, then collects results later when the run completes. This lets agents do other work while Actors are running.
+
+```text
+discover (search) -> discover (schema) -> start -> collect
+```
+
+### Batch multiple targets
+
+Most Actors accept arrays in their input (for example, `startUrls`, `queries`, `usernames`). Always batch multiple targets into a single run - one run with 5 URLs is cheaper and faster than 5 separate runs.
```
drobnikj: Note: The batch advice is mentioned twice (here and line 130). Consider consolidating into one section or removing duplication.

drobnikj (approved):
My a few bits, but pre-approving and leaving more to the doc team to check.

marcel-rbro (approved):
LGTM 👍

PR #2407, opened 2026-04-09
feat: residential proxy state targeting

daniil-poletaev:
This PR adds information about US state targeting to residential proxy page.

More context about the feature here: https://github.com/apify/apify-proxy/issues/1350

### DIFF
--- a/sources/platform/proxy/residential_proxy.md
+++ b/sources/platform/proxy/residential_proxy.md
@@ -109,11 +109,13 @@ async def main():
 </Tabs>
 
 ### How to set a state
+
 :::note US states only
 State-level targeting is currently only supported for the United States.
 :::
 
-To use state targeting, specify the `country` parameter in the [username](./usage.md#username-parameters) using the [ISO 3166-2:US](https://en.wikipedia.org/wiki/ISO_3166-2:US) format: `country-US_XX`, where `XX` is the two-letter state abbreviation. For example, to target California when using the proxy URL directly, set the username to `groups-RESIDENTIAL,country-US-CA` 
+To use state targeting, specify the `country` parameter in the [username](./usage.md#username-parameters) using the [ISO 3166-2:US](https://en.wikipedia.org/wiki/ISO_3166-2:US) format: `country-US_XX`, where `XX` is the two-letter state abbreviation. For example, to target California when using the proxy URL directly, set the username to `groups-RESIDENTIAL,country-US-CA`
+
 
 In the [Apify SDK](/sdk) you set the state in your proxy configuration using the `countryCode`/`country_code` parameter:
 
--- a/sources/platform/proxy/residential_proxy.md
+++ b/sources/platform/proxy/residential_proxy.md
@@ -69,13 +69,7 @@ async def main():
 
 ### How to set a proxy country
 
-When using [standard libraries and languages](./datacenter_proxy.md), specify the `country` parameter in the [username](./usage.md#username-parameters) as `country-COUNTRY-CODE`.
-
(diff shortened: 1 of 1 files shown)


sources/platform/proxy/residential_proxy.md
```diff
@@ -114,6 +114,57 @@ async def main():
...
+
+:::note US states only
+State-level targeting is currently only supported for the United States.
+:::
+
+To use state targeting, specify the `country` parameter in the [username](./usage.md#username-parameters) using the [ISO 3166-2:US](https://en.wikipedia.org/wiki/ISO_3166-2:US) format: `country-US_XX`, where `XX` is the two-letter state abbreviation.
+
+For example, your `username` parameter when targeting California will look like this:
+
+```python
```
bliuchak: Why python? 🤔

bliuchak (approved):
LGTM

sources/platform/proxy/residential_proxy.md
daniil-poletaev: That's a good point, but that's how we had it before 😄 

```
### How to set a proxy country

When using [standard libraries and languages](./datacenter_proxy.md), specify the `country` parameter in the [username](./usage.md#username-parameters) as `country-COUNTRY-CODE`.

For example, your `username` parameter when using [Python 3](https://docs.python.org/3/) will look like this:

```python
username = "groups-RESIDENTIAL,country-JP"
```
```

Comment (TC-MO) [reply sources/platform/proxy/residential_proxy.md]: Could we also get equivalent code in JS and use tabs component?

Comment (daniil-poletaev) [reply sources/platform/proxy/residential_proxy.md]: Yeah, but just to be clear we use ```python for that 👇 

Maybe I should use it without language (just ```)?

Screenshot 2026-04-13 at 12 56 18

Comment (TC-MO) [reply sources/platform/proxy/residential_proxy.md]: Without code declaration I think docusaurus would throw an error

Comment (daniil-poletaev) [reply sources/platform/proxy/residential_proxy.md]: Yeah, but it doesn't make sense to have there two tabs as the code is identical 🤔

Comment (TC-MO) [reply sources/platform/proxy/residential_proxy.md]: Yeah good point, I'll take a closer look at this section maybe a slight rewrite could help this issue

File: sources/platform/proxy/residential_proxy.md [hunk 2]
```diff
@@ -114,6 +114,57 @@ async def main():
...
+username = "groups-RESIDENTIAL,country-US_CA"
+```
```
TC-MO: I think inline code will be better here than using codeblock with python declaration just for generic username I'll update as well the country targeting section for consistency

```suggestion
### How to set a state

:::note US states only
State-level targeting is currently only supported for the United States.
:::

To use state targeting, specify the `country` parameter in the [username](./usage.md#username-parameters) using the [ISO 3166-2:US](https://en.wikipedia.org/wiki/ISO_3166-2:US) format: `country-US_XX`, where `XX` is the two-letter state abbreviation. For example, to target California when using the proxy URL directly, set the username to `groups-RESIDENTIAL,country-US-CA` 
```

sources/platform/proxy/residential_proxy.md
TC-MO, line 73: Fix for consistency with new section

```suggestion
### How to set a proxy country

When using [standard libraries and languages](./datacenter_proxy.md), specify the `country` parameter in the [username](./usage.md#username-parameters) as `country-COUNTRY-CODE`. For example, to target Japan, set the username to `groups-RESIDENTIAL,country-JP`.

```

TC-MO (approved):
Minor issues, once done safe to merge

PR #2412, opened 2026-04-09
docs(openapi): add 402 response to Actor execution endpoints

tobice:
Updates API schema to match changes from apify/apify-core#26917: unauthenticated Actor execution requests now return 402 with X402 payment info instead of 401.

I have also updated the `PamentRequired` example to be more realistic.


tobice:
The `POST <...>/runs` execution endpoint has been already updated by @Pijukatel :smile: :runner: https://github.com/apify/apify-docs/pull/2390

mvolfik (approved):

apify-api/openapi/components/responses/PaymentRequired.yaml
```diff
@@ -1,9 +1,9 @@
-description: Payment required - the user has exceeded their usage limit or does not have enough credits.
+description: Payment required - the user has exceeded their usage limit, does not have enough credits, or the request lacks authentication and payment credentials.
 content:
   application/json:
     schema:
       $ref: ../schemas/common/ErrorResponse.yaml
```
janbuchar, line 5: You should use `allOf` here to make sure that `type` is set to the one true allowed value

janbuchar (changes requested):

apify-api/openapi/components/responses/PaymentRequired.yaml
tobice: It's following the same pattern as all other responses 🤔 Should it be changed everywhere? If so, then I'd do it in a separate PR.

janbuchar: Well, in https://github.com/apify/apify-docs/blob/3b28c7aeebbb3f77c51a6a89f7fc017e7a689eac/apify-api/openapi/components/schemas/common/errors/ActorErrors.yaml, it is correct already... I thought that PR changed every error response. @vdusek can you share you opinion please?

stepskop (approved):

apify-api/openapi/components/responses/PaymentRequired.yaml
tobice: @janbuchar @vdusek Can we resolve this? :smile:

vdusek: Sorry! A lot of notifications after free Friday, and I got lost.

I would not do it in this PR. Let's resolve this in a dedicated PR that updates all generic error responses to use the `allOf` + `const` pattern (which will constrain each response's type field to only the relevant error type). I'll do it myself.

tobice: Thanks! In that case, @janbuchar can I ask for your approve? :pray:

janbuchar (approved):

PR #2415, opened 2026-04-09
fix(openapi): allow multiple values in request queue requests list

mvolfik:
add docs for feature added in https://github.com/apify/apify-core/issues/12115


mvolfik:
hm, for some reason the comma is escaped in the `curl`, PHP `curl`, javascript `xhr` and Java unirest examples:

image

It's fine for the rest though:

image

Is this blocking? Do you guys know why this might be?

fnesveda (approved):

Pijukatel (approved):

Pijukatel:
> Is this blocking? Do you guys know why this might be?

I tested with our API, and it seems to handle both cases identically, so I would not worry about that.

Issue #2286, opened 2026-02-26
Fix OpenAPI specification issues reported by the validator


Pijukatel:
OpenAPI validator middleware was created to validate requests and responses using the OpenAPI specification: https://github.com/apify/apify-core/issues/22169

Already, when running the integration tests, many validation errors exposed problems with the API. All those errors have to be fixed before the validator goes to production. Validator should be running in production only if there are no more errors detected in integration tests.

Full log from first round of integrations tests from validator draft https://github.com/apify/apify-core/pull/26052:
https://apify-pr-test-env-logs.s3.us-east-1.amazonaws.com/apify/apify-core/26052/api-8086980ca33efb927256133b4498375bf3fa955b.html

Such log should contain no `OpenAPI validation error`

vdusek:
We did our best, let's consider it as done and close it.

PR #2422, opened 2026-04-13
docs(openapi): Align 404 to use generic error and specific example

Pijukatel:
- Add missing, but possible errors to the `ErrorResponse.yaml`
- Simplify 404s to refer to a generic error and give generic missing resource text

**Motivation:** 
- Simplify code generation by reducing the number of error objects.
- Simplify upcoming documentation of redispatch endpoints (This will allow complete reuse of the response section. Using granular 404s would not allow it due to, for example, `last run` endpoints always having `actor/run not found` errors even when referring to the default dataset, while getting the dataset directly through id will never return such 404 variant )

Follow up for: https://github.com/apify/apify-docs/pull/2398

Partially implements: https://github.com/apify/apify-docs/issues/2286


fnesveda (approved):

vdusek (changes requested):
please see what it generated in https://github.com/apify/apify-client-python/pull/719/ - `ErrorType`, `Type1`

vdusek (approved):
LGTM

PR #2403, opened 2026-04-08
docs(openapi): Autofix OpenAPI spec validation errors

Pijukatel:
## Summary
Autogenerated OpenAPI fixes suggestions based on validation errors generated from running API integration tests with OpenAPI validator turned on.

**Error log:** https://apify-pr-test-env-logs.s3.us-east-1.amazonaws.com/apify/apify-core/26280/api-e8976db3b5b2476d2f01af1a84d4268fc61e839f.log

**apify-core version:** https://github.com/apify/apify-core/tree/e8976db3b5b2476d2f01af1a84d4268fc61e839f

**Stop reason:** No new fixes possible. All remaining validation errors are deliberate negative tests, known false positives (nullable date-time fields), validator limitations (oneOf for KV store records), missing endpoints (out of scope), or global middleware artifacts (method query parameter override). 16 unique error types fixed out of 20 target.

#### 1. Fix pricingInfo oneOf validation errors
- **Files:** `components/schemas/actor-pricing-info/FreeActorPricingInfo.yaml`, `PayPerEventActorPricingInfo.yaml`, `PricePerDatasetItemActorPricingInfo.yaml`, `FlatPricePerMonthActorPricingInfo.yaml`
- **Error:** `must have required property 'pricePerUnitUsd'` / `must have required property 'pricingPerEvent'` / `must have required property 'trialMinutes'` / `must have required property 'unitName'` at GET/PUT `/v2/acts/{actorId}`
- **Root cause:** The pricingInfo oneOf schemas had mismatched required fields. Each pricing model type had required fields that don't exist for that model (e.g., FREE model doesn't have pricePerUnitUsd). Inlined the pricingModel enum values to each schema so the validator can discriminate correctly.
- **Reference:** https://github.com/apify/apify-core/tree/e8976db3b5b2476d2f01af1a84d4268fc61e839f/src/packages/actor/src/pricing/pricing.both.ts#L22

#### 2. Fix maxItems type in RunOptions
- **Files:** `components/schemas/actor-runs/RunOptions.yaml:26`
- **Error:** `must be integer` at response path `/response/data/options/maxItems`
- **Root cause:** maxItems was typed as `integer` but the API returns `null` when not set. Changed to `type: [integer, "null"]`.
- **Reference:** https://github.com/apify/apify-core/tree/e8976db3b5b2476d2f01af1a84d4268fc61e839f/src/packages/schemas/src/actor_tasks.ts#L53

#### 3. Fix version DELETE 404 error type
[...]

PR #2423, opened 2026-04-13
docs(openapi): Remove uri format where not enforced by API

Pijukatel:
- Remove uri format where not enforced by API:
  - `RequestBase.loadedUrl`

- Remove unnecessary properties from UserData
 
- `Format uri` left for following:                                                    
  - Webhook.yaml, WebhookUpdate.yaml, WebhookShort.yaml → requestUrl: regex + runtime validation (WEBHOOK_REQUEST_URL_REGEXP + new URL())
  - Profile.yaml → websiteUrl: HTTP_URL_REGEX validation                                                                                     
  - Profile.yaml → pictureUrl: runtime S3 URL check     
  - All output-only fields (consoleUrl, keysPublicUrl, itemsPublicUrl, containerUrl, recordPublicUrl, standbyUrl,                            
  StoreListActor.pictureUrl/userPictureUrl/url): generated by the API itself, always valid URIs   

**Follow-up for:** https://github.com/apify/apify-docs/pull/2403#discussion_r3071462194

**Partially implements:** https://github.com/apify/apify-docs/issues/2286

### DIFF
--- a/apify-api/openapi/components/schemas/request-queues/RequestBase.yaml
+++ b/apify-api/openapi/components/schemas/request-queues/RequestBase.yaml
@@ -11,7 +11,6 @@ properties:
     $ref: ./SharedProperties.yaml#/RetryCount
   loadedUrl:
     type: [string, "null"]
-    format: uri
     description: The final URL that was loaded, after redirects (if any).
     examples: [https://apify.com/jobs]
   payload:
--- a/apify-api/openapi/components/schemas/request-queues/RequestUserData.yaml
+++ b/apify-api/openapi/components/schemas/request-queues/RequestUserData.yaml
@@ -2,12 +2,6 @@ title: RequestUserData
 description: Custom user data attached to the request. Can contain arbitrary fields.
 type: object
 additionalProperties: true
-properties:
-  label:
-    type: [string, "null"]
-    description: Optional label for categorizing the request.
-    examples: [DETAIL]
-  image:
-    type: [string, "null"]
-    description: Optional image URL associated with the request.
-    examples: [https://picserver1.eu]
+example:
+  label: DETAIL
+  customField: custom-value
--- a/apify-api/openapi/components/schemas/request-queues/RequestUserData.yaml
+++ b/apify-api/openapi/components/schemas/request-queues/RequestUserData.yaml
@@ -9,6 +9,5 @@ properties:
     examples: [DETAIL]
   image:
     type: [string, "null"]
-    format: uri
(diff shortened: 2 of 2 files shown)


vdusek (approved):
lgtm, thanks

apify-api/openapi/components/schemas/request-queues/RequestBase.yaml
```diff
@@ -11,7 +11,6 @@ properties:
     $ref: ./SharedProperties.yaml#/RetryCount
   loadedUrl:
     type: [string, "null"]
-    format: uri
     description: The final URL that was loaded, after redirects (if any).
     examples: [https://apify.com/jobs]
```
fnesveda, line 15: Huh, I didn't even know this is a thing. I don't think anyone actually uses it, but I guess we can't remove it now since it's a part of the API.

apify-api/openapi/components/schemas/request-queues/RequestUserData.yaml
```diff
@@ -9,6 +9,5 @@ properties:
     examples: [DETAIL]
   image:
     type: [string, "null"]
-    format: uri
     description: Optional image URL associated with the request.
     examples: [https://picserver1.eu]
```
fnesveda: These aren't specified anywhere in the API. `request.userData` should be a blackbox object (at least as far as the API is concerned).

I would just remove these.

fnesveda (changes requested):
Where are these properties even coming from? We don

apify-api/openapi/components/schemas/request-queues/RequestUserData.yaml
Pijukatel: I do not know the history behind these, but it has been part of the spec for years:

 https://github.com/apify/apify-docs/blob/5442ba282b65c52ae732a7bd964d1654b303b628/docs/api_v2/api_v2_reference.apib#L3434

I could not find anything in apify-core related to this, so maybe we can just remove both `label` and `image` and keep it as an object with extra properties allowed. We can put `label` in the example since Crawlee is using it.

fnesveda: Agreed.

fnesveda (approved):

apify-api/openapi/components/schemas/request-queues/RequestBase.yaml
Pijukatel: I will just remove this non existent contraint from the spec.

PR #2426, opened 2026-04-14
docs: Update monetization screenshots and procedure

szaganek:
- Turn UI descriptions into a procedure.
- Move admonition under a bullet it refers to.
- Add a new type of screenshot.
- Delete unnecessary screenshots.


sources/platform/actors/publishing/monetize/index.mdx
```diff
@@ -40,21 +40,17 @@ The following table compares the two main pricing models available for monetizin
...
-![ppe monetization wizard](../images/ppe-wizard.png)
-</TabItem>
-<TabItem value="Rental" label="Rental">
-![rental monetization wizard](../images/rental-wizard.png)
-</TabItem>
-</Tabs>
+To monetize your Actor:
+
+1. Log in to [Apify Console](https://console.apify.com).
+1. In the left-side panel, go to **Development** > **My Actors**.
```
TC-MO, line 46: nit: should we use directional language as such I always thought it's not a good practice due to accessibility

sources/platform/actors/publishing/monetize/index.mdx
```diff
@@ -40,21 +40,17 @@ The following table compares the two main pricing models available for monetizin
...
+1. From the table, select the Actor you want to monetize.
+1. Go to the **Publication** tab.
+1. Complete each section with required information.
+    - In **Monetization** section, complete your billing and payment details, and choose your [pricing model](#pricing-models).
+1. Click **Publish on Store**.
```
TC-MO: Perhaps `select` instead of `click` to account for different devices?

TC-MO (approved):
Two nits, otherwise LGTM

sources/platform/actors/publishing/monetize/index.mdx
szaganek: I generally agree, but I also feel that we need some kind of term for this element, since it's a starting point for all paths. `Main menu` doesn't really cut it for me 🤔 And `left-side` seemed generic enough.

sources/platform/actors/publishing/monetize/index.mdx
szaganek: Agree, could we maybe add it to the style guide as well?

PR #2431, opened 2026-04-14
docs(openapi): Add undocumented dataset endpoints

Pijukatel:
### Description
- Add undocumented, but public API endpoints related to datasets:
  - default Actor run datasets
  - last Actor run default dataset
- Abstract reusable parts of nearly duplicate endpoints and use references.
- Create a new group of endpoints, `Convenience endpoints`, to avoid cluttering the documentation.

(More undocumented endpoints will be added in follow-up PRs)

### Issues
Partially implements: https://github.com/apify/apify-docs/issues/2286


Pijukatel:
In docs:
image

fnesveda (approved):
Uff, I thought we documented these when we implemented this feature. Good job!

Pijukatel:
> Uff, I thought we documented these when we implemented this feature. Good job!

They were documented in plain text in the description of [act-runs-last-get](https://docs.apify.com/api/v2/act-runs-last-get), but that is not OpenAPI compliant, and it is not possible to process it with tools that consume the specification. 

Once I add all those endpoints in follow-up PRs I will also update that description with links.

PR #2445, opened 2026-04-20
feat: Add icon to external links

szaganek:

PR #2456, opened 2026-04-22
fix: Reference image from docs-theme for external links

szaganek:
Fixes: https://github.com/apify/apify-docs/pull/2445#issuecomment-4294527326


barjin (approved):
Looks alright to me, thank you for the quick fix 👍

PR #2353, opened 2026-03-23
docs: update workato docs

daveomri:
## Summary

- Remove OAuth 2.0 authentication flow, keeping only API token authentication as a single streamlined set of steps.
- Add a lead paragraph and horizontal rule at the top of the page for consistency with other integration pages.
- Update connector details to reflect current behavior: crawler type list (removed JSDOM), memory range info, validation rules for URL fields and dataset item limits, Actor name format (`owner/name` with slash and tilde).
- Remove promotional/filler language and fix formatting to match Apify docs style guide: remove italics from action/trigger descriptions, use bold with dashes for definition lists, add periods to full-sentence list items, fix heading hierarchy, and shorten front matter description to meet the 140-160 character SEO limit.
- Fix markdownlint errors (indented sub-list, missing blank lines).
- Update the Discord link text to "developer community on Discord" and replace the removed OAuth troubleshooting tip with a validation errors tip.
- Delete unused OAuth-related screenshots.

PR #2461, opened 2026-04-22
docs: updated the workato docs, removed old sections

daveomri:
## Summary

Follow-up fixes for the Workato integration docs (#2353) to address inaccuracies found when testing against the live Workato UI.

- **Remove redundant connector release step**
   - Workato prompts for release immediately after installation, so the manual "Release latest version" instructions (and screenshot) are unnecessary.
- **Fix project creation step** 
   - Simplified the step and corrected the option name (the old "Build a workflow app" option no longer exists in the Workato UI).
- **Fix build tag default description** 
   - Changed "defaults to `latest`" to "defaults to the Actor's default build tag" in both Run Actor and Run task sections, since the actual default comes from the Actor's configuration in Apify Console.
- **Update Scrape URL screenshot** 
   - Replaced with a current screenshot that no longer shows deprecated crawlers.


marcel-rbro (approved):
LGTM 👍

PR #2465, opened 2026-04-24
feat: implement segment link tracking, opt-in solution

marekh19:
Implements a global, opt-in Segment link click tracking solution.

Resolves #2409.

---

To verify:

1. Navigate to https://pr-2465.preview.docs.apify.com/legal
2. Click the "Apify Trust Center: https://trust.apify.com/" link
3. Check for a POST request to https://analytics.apify.com/v1/t and validate that it sends all the required attributes and responds with a success code.

### DIFF
--- a/src/plugins/docusaurus-plugin-segment/linkTracking.js
+++ b/src/plugins/docusaurus-plugin-segment/linkTracking.js
@@ -1,0 +1,59 @@
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+
+// Opt-in link trackers. Currently limited to Trust Center to avoid Segment noise - see https://github.com/apify/apify-docs/issues/2409.
+const LINK_TRACKERS = [
+    {
+        test: (url) => url.hostname === 'trust.apify.com',
+        element: 'trust-center-link',
+    },
+    // Examples for future opt-in:
+    // { test: (url) => !/(^|\.)apify\.com$/.test(url.hostname) } // all outbound
+    // { test: () => true }                                       // all links
+];
+
+let installed = false;
+
+function handleLinkClick(event) {
+    // auxclick fires for all non-primary buttons; only handle middle click (button === 1)
+    if (event.type === 'auxclick' && event.button !== 1) return;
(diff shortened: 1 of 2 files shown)


src/plugins/docusaurus-plugin-segment/linkTracking.js
```diff
@@ -0,0 +1,59 @@
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+
+// Opt-in link trackers. Currently limited to Trust Center to avoid Segment noise - see https://github.com/apify/apify-docs/issues/2409.
+const LINK_TRACKERS = [
+    {
+        test: (url) => url.hostname === 'trust.apify.com',
+        element: 'trust-center-link',
```
webrdaniel, line 17: nit: please add a comment explaining also the `element` field

webrdaniel (approved):
Nice 👏

src/plugins/docusaurus-plugin-segment/linkTracking.js
```diff
@@ -0,0 +1,59 @@
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+
+// Opt-in link trackers. Currently limited to Trust Center to avoid Segment noise - see https://github.com/apify/apify-docs/issues/2409.
+const LINK_TRACKERS = [
```
katzino, line 14: Question: Haven't you considered to instead mark elements to track with dataset attribute? Like `data-tracking-id` or similar? I don't mind this approach but it might be more clear what is actually being tracked.

katzino (approved):

birosrichard (approved):

src/plugins/docusaurus-plugin-segment/linkTracking.js
marekh19: The link we're currently tracking is a raw Markdown autolink, which doesn't support HTML attributes - converting it to JSX just to attach a `data-tracking-id` didn't feel worth it. That said, I added support for it anyway: any `<a>` with `data-tracking-id` will fire the event with that as the element label, taking precedence over URL matchers.

PR #2470, opened 2026-04-28
docs: reduce llms.txt size from ~196K to ~84K characters

marcel-rbro:
## Summary

- Reduce `llms.txt` from ~196K to ~84K characters (57% reduction, under 100K target)
- Replace live fetching of external repo `llms.txt` with a curated static file (`scripts/llms-external-curated.txt`) that keeps overview, concept, and guide pages but drops ~400 individual class/interface/enum reference entries
- Exclude individual API endpoint pages from the Docusaurus llms-txt plugin via `excludeRoutes`, keeping only Introduction category pages and Getting started
- Exclude legacy Academy content (old JS scraping course, node-js/python tutorial collections, exercise solutions, marketing playbook deep pages) and outdated legal documents
- `llms-full.txt` remains unchanged - it still fetches all content from external repos

## Test plan

- [ ] `npm run build` succeeds
- [ ] `wc -c build/llms.txt` is under 100K characters
- [ ] `build/llms.txt` contains API Introduction pages but no individual endpoint pages
- [ ] `build/llms.txt` contains no legacy Academy content or reference class entries
- [ ] `build/llms-full.txt` size is unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2480, opened 2026-04-29
docs: omit .md pages from llms.txt without removing them completely

webrdaniel:
Follow-up to #2470. Listing pages in the llms-txt plugin's `excludeRoutes` also drops their `/<route>.md` counterparts from the build, so URLs like `https://docs.apify.com/sdk.md` started returning 404 (raised in https://github.com/apify/apify-docs/pull/2470#discussion_r3161627392).

This PR moves the exclusion from build time to post-build:

- **`docusaurus.config.js`**: revert `excludeRoutes` back to just `/` and `/search`; add a NOTE so future contributors don't re-introduce the regression.
- **`scripts/joinLlmsFiles.mjs`**: add `LLMS_INDEX_EXCLUDE_PATTERNS` and a `filterLlmsIndex()` postbuild step that strips matching `- [Title](url)` entries (and now-empty `## Section` headings) from the generated `build/llms.txt`. The `.md` files stay on disk and continue to serve. Also fixes a pre-existing fire-and-forget race between `joinFiles()` and `sanitizeFile()`.
- **`package.json`**: add `@docusaurus/utils` as a direct dependency (used for `createMatcher`).
- **`.github/workflows/test.yaml`**: add regression tests asserting that `/sdk.md`, `/open-source.md`, `/api/v2/actor-builds-get.md`, `/api/v2/dataset-get.md`, and `/academy/tutorials.md` still serve `text/markdown`. Also adds `assert_final_content_type` so child-repo homepages (`/sdk/js`, `/sdk/python`, `/api/client/{js,python}`, `/cli`) are checked through their nginx redirects for both HTML and `Accept: text/markdown` responses.

Net effect: same `llms.txt` index as #2470 produced, but the per-page `.md` files are restored.

## Test plan

- [ ] CI passes (the new `.md`-counterpart and child-repo redirect assertions exercise the regression)
- [ ] `npm run build` succeeds locally
- [ ] `build/llms.txt` size remains under the 100K limit enforced by `npm run test:llms-size`
- [ ] Manually verify a few previously-broken URLs once deployed: `https://docs.apify.com/sdk.md`, `https://docs.apify.com/open-source.md`, `https://docs.apify.com/api/v2/actor-builds-get.md`


jancurn:
Cheers. Pls can we add some tests for these special pages, to ensure the .md version work, and also that the HTML version contain the `<link rel="alternate" type="text/markdown" href="https://docs.apify.com/xxx.md">` tag ?

TC-MO:
The builds will fail until we deploy it or do we need to make some changes to test assertions in `nginx.conf`?

webrdaniel:
The tests should now run correctly against the staging

jancurn:
Great thank you

marcel-rbro (approved):
Can we nudge the review? Is @B4nan the one to review the technical side?

.github/workflows/test.yaml
```diff
@@ -49,17 +49,18 @@ jobs:
...
                     error_log ${PWD_PATH}/logs/error.log;
                     pid ${PWD_PATH}/logs/nginx.pid;
                     events {}
                     http {
                         access_log ${PWD_PATH}/logs/access.log;
-                        include ${PWD_PATH}/nginx.conf;
+                        include ${PWD_PATH}/nginx-test.conf;
                     }
                     EOF
-                    sed -i 's|https://apify.github.io/apify-docs|http://localhost:3000|g' default.conf
```
B4nan, line 55: are those changes actually necessary? the CI checks were working fine before, i was expecting you just add a few more test cases here

B4nan:
> The tests should now run correctly against the staging

Also what do you mean by that? We dont have any staging env for the docs.

.github/workflows/test.yaml
webrdaniel: Without the change, the tests fail on:
` Expected 'text/markdown' in 'Content-Type' for http://localhost:8080/sdk.md`
Looks like nginx is proxying to the deployed production site instead of the locally-built one.

B4nan: I think that was on purpose, it wasn't possible to test something locally. But I might be wrong, it's been quite some time since I was setting this up.

webrdaniel: Ok, so keeping it as it was

webrdaniel:
> > The tests should now run correctly against the staging
> 
> Also what do you mean by that? We dont have any staging env for the docs.

This referred to the fix mentioned above that is now reverted

B4nan (approved):
I don't wanna block this, feel free to put the changes back and merge. 

Btw I was also confused by you renaming the nginx config to `nginx-test.conf`, I don't think it's necessary.

PR #2491, opened 2026-05-04
docs: Delete unused images in integrations

szaganek:


marcel-rbro (approved):
LGTM, thanks 👍

marcel-rbro:
Also, consideration for the future, we should prolly add a check for unused assets.

szaganek:
> Also, consideration for the future, we should prolly add a check for unused assets.

Already considered, @TC-MO wants to work on it 👍

PR #2493, opened 2026-05-04
chore: migrate from eslint to oxlint

B4nan:
## Summary

Replaces ESLint with [oxlint](https://oxc.rs/docs/guide/usage/linter), pulling in the shared `@apify/oxlint-config` preset and `oxlint-tsgolint` for type-aware rules. Mirrors the migration already shipped in apify-storage-local-js, apify-shared-js, apify-core, apify-client-js, apify-sdk-js, and actor-scraper.

### What changed

- Drop `eslint`, `typescript-eslint`, `eslint-plugin-react`, `eslint-plugin-yml`, `globals`, `@apify/eslint-config`
- Add `oxlint@1.62.0`, `oxlint-tsgolint@0.22.0`, `@apify/oxlint-config@^0.2.5`
- Replace `eslint.config.mjs` with `oxlint.config.ts`; drop unused `tsconfig.eslint.json`
- `lint:code` / `lint:code:fix` now run `oxlint --type-aware`
- Drop the YAML-specific lint script (`openapi:lint:yaml`) and its CI step — Redocly + Spectral already validate the OpenAPI spec; oxlint doesn't lint YAML, and we don't want eslint hanging around just for style nits on YAML files
- Bump `@apify/tsconfig` to `^0.1.2` and add `module: Node16` / `moduleResolution: Node16` so oxlint-tsgolint can parse the project
- Disable `typescript/no-floating-promises` and `typescript/promise-function-async` repo-wide — the React event handlers in `apify-docs-theme/` intentionally fire-and-forget
- Ignore `examples/` in oxlint (each example has its own tsconfig with its own bundler settings)

### Lint output

\`\`\`
$ pnpm lint:code
Found 0 warnings and 0 errors.
Finished in 312ms on 132 files with 113 rules using 14 threads.
\`\`\`

## Test plan

- [x] \`pnpm lint:code\` — clean
- [x] \`pnpm openapi:lint\` — clean (Redocly + Spectral; YAML-style step dropped)

PR #2494, opened 2026-05-04
chore: migrate to oxfmt

B4nan:
## Summary

Adds [oxfmt](https://oxc.rs/docs/guide/usage/formatter) as the formatter, matching the toolchain in apify-storage-local-js, apify-shared-js, apify-core, apify-web, apify-client-js, apify-sdk-js, actor-scraper, and crawlee v4. The repo had no formatter before — this is a fresh adoption.

Stacked on top of #2493 (eslint→oxlint migration). Will be retargeted to \`master\` once the parent merges.

### Commits

- \`chore: add oxfmt tooling\` — adds \`oxfmt@0.46.0\` devDep, \`.oxfmtrc.json\` (4-space tabs, single quotes, 120 col), and \`format\` / \`format:check\` scripts
- \`chore: apply oxfmt formatting\` — pure \`oxfmt --write\` reformat across 219 files (mostly \`.css\`, \`.html\`, \`.js\`/\`.jsx\`, \`.ts\`/\`.tsx\`, \`.mjs\`)
- \`ci: add format-check step\` — gates the lint job on \`pnpm format:check\`

## Test plan

- [x] \`pnpm lint:code\` — clean
- [x] \`pnpm format:check\` — clean

## Follow-up

Once this lands, add the reformat commit SHA to \`.git-blame-ignore-revs\` so blame skips it.


.oxfmtrc.json
```diff
@@ -0,0 +1,24 @@
+{
+    "tabWidth": 4,
+    "singleQuote": true,
+    "trailingComma": "all",
+    "printWidth": 120,
```
fnesveda, line 2: Are we dead set on 120? We used to have 160 in ESLint.

.oxfmtrc.json
```diff
@@ -0,0 +1,24 @@
+{
+    "tabWidth": 4,
```
fnesveda: Can't `oxfmt` inherit these from `.editorconfig`, so that the config is not duplicated?

fnesveda (changes requested):
Just a couple of questions, nothing major

.oxfmtrc.json
B4nan: eslint was just enforcing the max line length, while formatters are building code from AST with this limit. if you use 160 for a formatter, it will use it as long as things fit in the limit, resulting in a lot of long lines. i think 120 is the sweet spot.

.oxfmtrc.json
B4nan: yeah, looks like it respects editorconfig, so we can omit `tabWidth` and `singleQuote` options

.oxfmtrc.json
fnesveda: OK 👍

fnesveda (approved):

janbuchar (approved):
LGTM

PR #2498, opened 2026-05-05
fix: US state targeting SDK configuration

daniil-poletaev:
This PR fixes examples of US state targeting SDK proxy configuration.


bliuchak (approved):
Gud!

sources/platform/proxy/residential_proxy.md
TC-MO:
```suggestion
In the [Apify SDK](/sdk) you set the state in your proxy configuration using the `countryCode`/`country_code` and `subdivisionCode`/`subdivision_code` parameter:
```

TC-MO (approved):
One nit, we should update the prose as well right? Otherwise LGTM

PR #209, opened 2021-03-09
Small fixes in Schedules API docs

mtrunkat:

PR #235, opened 2021-05-04
Integrate the Public Actor guide in actors docs

m-murasovs:
Integrating the PA guide from Notion into existing docs: https://www.notion.so/apify/Public-actor-Guide-7c2ef51d4c9846afa33567dc516cabfa

The preview is available here: https://docs.apify.com/actors?version=public-actor-guide

UPDATES:
- The "Actors" landing page now features some basics on building public actors and the Inspiration section: https://docs.apify.com/actors?version=public-actor-guide
- The Promotion and SEO content was added to the existing `Publishing` and `SEO` pages and SEO page renamed "SEO and promotion": https://docs.apify.com/actors/publishing/seo-and-promotion?version=public-actor-guide
- The "Development" page now contains the "Before you begin" and "Essential materials" sections that lead to public wiki pages
- Added the "Testing and maintenance" article: https://docs.apify.com/actors/development/testing-and-maintenance?version=public-actor-guide

closes #223

Issue #219, opened 2021-03-30
Docs & API ref: Add information about passing token using HTTP header


drobnikj:
Keep information about passing the token as a query param, but makes them less visible than info about passing token as HTTP header.
Passing token as HTTP header is more secure.

mtrunkat:
Header:

`Authorization: Bearer {token}`

Issue #217, opened 2021-03-24
Write an article about Server Migrations 


m-murasovs:
Maybe similar to State Persistence?
https://docs.apify.com/actors/development/state-persistence#how-often-do-migrations-occur

mtrunkat:
Closing as we have this: https://docs.apify.com/actors/development/state-persistence#code-examples

PR #2499, opened 2026-05-06
chore(deps): bump axios and assorted transitive deps in pnpm-lock.yaml

### LINKED ISSUE #215: Update the Updates and Versioning of SDK article
https://help.apify.com/en/articles/3184510-updates-and-versioning-of-apify-sdk

From Zuzka: 
Not sure how often we update these but I just ran across an article where we state that Apify 1.0 is not released yet -

### LINKED ISSUE #216: No search results view restyle

### LINKED ISSUE #220: Edit and add the Paid actors docs to Help/Docs
Content: https://www.notion.so/apify/Paid-actors-user-docs-draft-7f636a1930554c1f8c4bf3eefbb6bb4e

### LINKED ISSUE #223: Mov Public actor guide to docs
Conversation here: https://apifier.slack.com/archives/CQ96RHG2U/p1618840727002600

### LINKED ISSUE #236: Update Apify Client in Storage docs
The code examples use the old Apify Client, so this needs to be updated

https://apifier.slack.com/archives/C0L33UM7Z/p1620404424149400

B4nan:
## Summary

Resolves all open Dependabot alerts in `pnpm-lock.yaml` via lockfile-only updates plus an extended `pnpm.overrides` block.

### Direct + workspace-range bumps
- **axios** → 1.16.0 (closes #223–#235); apify-docs-theme range widened ^1.7.9 → ^1.16.0

### Overrides reorganized
- Migrated the existing root `overrides` to `pnpm.overrides` (pnpm honors it more uniformly across the dep graph) and extended with the remaining vulnerable transitives:
  - **protobufjs** `^7.5.5` (closes #216 — *critical*)
  - **dompurify** `^3.4.0` (closes #215, #217–#219)
  - **fast-xml-parser** `^5.7.0` (closes #220)
  - **uuid@>=11.0.0** `^11.1.1` (closes #236; leaves the unrelated 8.3.2 entry alone)
  - **js-yaml@>=4.0.0** `^4.1.1` (closes #208)
  - **lodash** `^4.17.23` — replaces prior `^4.18.0` with a wider range that still satisfies the patched-version requirement (closes #209, #213, #214)
  - **postcss** `^8.5.10` — already present, now actually applied uniformly via pnpm.overrides (closes #222)
  - **yaml@1** `^1.10.3` — already present (closes #211)

### Resolved versions in lockfile after install
protobufjs 7.5.6 · dompurify 3.4.2 · fast-xml-parser 5.7.3 · uuid 11.1.1 · js-yaml 4.1.1 · lodash 4.17.23 + 4.18.1 · postcss 8.5.14 · yaml 1.10.3 · axios 1.16.0

No vulnerable versions remain.

### Out of scope
- **serialize-javascript** alerts (#210, #212) were dismissed as `not_used` (rspack pipeline doesn't pull the webpack/terser path that loads it).


pnpm-workspace.yaml
```diff
@@ -1,6 +1,22 @@
...
+  "@stoplight/spectral-ruleset-bundler>rollup": 2.80.0
+  postcss: ^8.5.10
+  yaml@1: ^1.10.3
+  lodash: ^4.17.23
+  protobufjs: ^7.5.5
+  dompurify: ^3.4.0
+  fast-xml-parser: ^5.7.0
+  uuid@>=11.0.0: ^11.1.1
+  js-yaml@>=4.0.0: ^4.1.1
+  ajv@>=8.0.0 <9.0.0: 8.18.0
```
barjin: If something requires a new major version of `uuid` or `js-yaml`, will this force it back on `uuid@11` and `js-yaml@4`? Shouldn't we use the pattern from `ajv` here, instead?

barjin (approved):
A quick question, but not blocking (it seems it has been like this before as well). Thanks!

pnpm-workspace.yaml
B4nan: Good catch — bounded both ranges to the current major in 57a27c03 (mirroring the `ajv` pattern), so a future uuid@12 / js-yaml@5 won't get silently pinned back.

PR #2511, opened 2026-05-07
docs: add GEMINI.md symlink to AGENTS.md

TC-MO:
Add GEMINI.md since Gemini also doesn't follow AGENTS.md. Symlink like with CLAUDE.md


marcel-rbro (approved):
LGTM 👍

PR #2513, opened 2026-05-07
docs: Delete screenshots in Organization account topic

szaganek:
Section on converting an account needs to be rewritten as well.


sources/platform/collaboration/organization_account/index.md
```diff
@@ -20,17 +20,24 @@ You can set up an organization in two ways.
...
+<!-- vale on -->
+You can create up to 5 organizations and be a member of as many organizations as you need.
 
 ## Convert an existing account
 
-> **When you convert an existing user account into an organization,**
+> **When you convert an existing user account into an organization:**
 >
 > * **You will no longer be able to sign in to the converted user account.**
 > * **An organization cannot be converted back to a personal account.**
```
TC-MO: can we remake this into admonition instead of blockquote?


```suggestion
> **When you convert an existing user account into an organization:**
>
> - **You will no longer be able to sign in to the converted user account.**
> - **An organization cannot be converted back to a personal account.**
```

sources/platform/collaboration/organization_account/index.md
```diff
@@ -20,17 +20,24 @@ You can set up an organization in two ways.
...
-
-You can create a new organization by clicking the **Create new organization** button under the **Organizations** tab in your [account](https://console.apify.com/settings/organizations). If you want the organization to have a separate email address (used for notifications), enter it here. Otherwise, leave the **email** field empty and the owner's email will be used for notifications.
-
-![Create a new organization](../images/organizations/create-new-org.png)
-
-**You can own up to 5 and be a member of as many organizations as you need.**
+## Create an organization
+
+To create an organization account:
+<!-- vale off -->
```
TC-MO, line 26: why vale off here?

sources/platform/collaboration/organization_account/index.md
```diff
@@ -20,17 +20,24 @@ You can set up an organization in two ways.
...
+1. Log in to [Apify Console](https://console.apify.com).
+1. In the left-side panel, go to **Settings**.
+1. Select the **Organizations** tab.
+1. In **My organizations**, select **Create a new organization**.
+1. Complete the following information:
+   * Organization name
+   * Organization username
+   * Organization email _(optional)_ – a dedicated email address for notifications. To use the owner's email address instead, leave this field empty.
```
TC-MO:
```suggestion
   - Organization name
   - Organization username
   - Organization email _(optional)_ – a dedicated email address for notifications. To use the owner's email address instead, leave this field empty.
```

TC-MO (changes requested):
Few changes requested and questions

sources/platform/collaboration/organization_account/index.md
szaganek: I got an error when I didn't use asterikses 😅 
https://github.com/apify/apify-docs/actions/runs/25497218928/job/74820073367

sources/platform/collaboration/organization_account/index.md
szaganek: So it doesn't flag en dash in explanation.

sources/platform/collaboration/organization_account/index.md
TC-MO: ah I see the issue. We prefer dashes but this doc has * and once one is used (~LOC20) mdlint tries to enforce the style of first used. If you could update the whole doc with dashes it should work

sources/platform/collaboration/organization_account/index.md
szaganek: Actually, we overuse admonitions. How about just regular text in the beginning.

TC-MO: EIther works for me, just no blockquotes :D

TC-MO (approved):
LGTM, if we could take this opportunity as well, there is another blockquote at LOC17, if we could get rid of it, that would be perfect otherwise ✅

PR #2522, opened 2026-05-12
docs(api): rename /v2/acts/ → /v2/actors/ in OpenAPI spec and sweep doc-site references

mtrunkat:
Mirrors apify/apify-core#27780 on the documentation side. The server PR keeps `/v2/acts/...` as an undocumented alias, but the canonical, documented path family is now `/v2/actors/...`. This PR updates the published OpenAPI spec and the docs-site code samples accordingly.

## Intentionally out of scope

- **`acts-get` page slug references** in `integrate_with_apify.md` and `bubble.md`. These URLs (e.g. `https://docs.apify.com/api/v2/acts-get`) are derived from the OpenAPI `operationId`s (`acts_get`, `act_runs_post`, etc.), which §5 of the audit covers separately. Until the operationIds are renamed, the `acts-get` URLs remain valid, so leaving them is correct for this PR.
- **`"actId":` example payload** in `webhooks/actions.md` (§12 fine-print). That's a payload field, which is §2 of the audit.
- **Renaming the path YAML files** themselves from `paths/actors/acts@*.yaml` to `actors@*.yaml`. The internal file names don't show up in the published spec; the `$ref` targets in `openapi.yaml` resolve them at build time. Worth a separate cleanup PR if desired.

## Related

- apify/apify-core#27780 — the server-side migration (canonical routes + legacy `/v2/acts/` alias).

https://claude.ai/code/session_015MYkSfTz4CsX3Y4LatpSGf

---
_Generated by [Claude Code](https://claude.ai/code/session_015MYkSfTz4CsX3Y4LatpSGf)_


jancurn:
We should mention somewhere in the API docs that the legacy `/acts` paths have been deprecated, but keep working for backward compatibility. Just to have this recorded somewhere, so that it doesn't surprise or scare our users

apify-api/openapi/openapi.yaml
```diff
@@ -111,6 +111,17 @@ info:
...
+    ## Legacy `/v2/acts/` URL prefix
+    <span id="/introduction/legacy-acts-prefix"></span>
+
+    Actor-scoped endpoints previously lived under `/v2/acts/`. That prefix is
+    **deprecated but continues to work** for backward compatibility - every
+    `/v2/acts/...` URL routes to the same handler as its `/v2/actors/...`
+    counterpart, returns the same response, and shares the same rate-limit bucket
+    and API usage counter. New integrations should use the canonical
+    `/v2/actors/` prefix, but existing clients (older SDK versions, persisted
+    webhook `requestUrl`s, saved cURL snippets) keep working without any changes.
```
janbuchar: I'd shorten this warning to like 1 sentence, this simply doesn't deserve that much explanation.

apify-api/openapi/openapi.yaml
```diff
@@ -480,79 +491,79 @@ servers:
   - url: "https://api.apify.com"
     variables: {}
 paths:
-  /v2/acts:
```
janbuchar, line 483: It's _not impossible_ to keep copies of the old versions with `deprecated: true`. We already have reusable operations for storage-handling operations (see `components/objects`) so there is precedent.

Of course, it might clutter the rendered docs quite a bit. But that's just speculation and it should be easy to try it.

jancurn: I'd remove it from OpenAPI, to reduce token use for agents, space in docs etc. it really brings no value

tobice (approved):
>  The server PR keeps /v2/acts/... as an undocumented alias, but the canonical, documented path family is now /v2/actors/....

I guess we'll have to update the middleware that ensures that the spec is up to date, right? Because the specification for the legacy paths will be missing there and the middleware needs to be able to detect it.

-- 

Anyway, this looks like basically `s/acts/actors/` for all paths 😄 Just like in the other PR, not sure if the PR description length is really justified here...

But there are plenty of reviewers here -> approve.

apify-api/openapi/openapi.yaml
szaganek: I'd suggest something less verbose, like below: 

Older Actor endpoints use the `/v2/acts/` URL prefix. This prefix is deprecated but backward compatible. In new integrations, use `/v2/actors/`. Existing URLs and clients don't require changes.

szaganek (approved):

PR #2528, opened 2026-05-13
docs: preserve tab labels and code language tags in agent-facing .md files

### LINKED ISSUE #1884: tab divided code blocks lost its language tag
https://docs.apify.com/platform/proxy.md

for some reason e.g. here we've lost ```js

### LINKED ISSUE #2529: Docusaurus <Tabs> flatten into unlabeled sequential code blocks in .md output
JS/Python code examples in <Tabs> components flatten into sequential code blocks with no indication that they're the same example in different languages

birosrichard:
closes: #1884
closes: #2529 
Adds two rehype plugins to the signalwire llms-txt pipeline that run before HTML→markdown conversion:

- rehypeExpandTabs: expands Docusaurus <Tabs> into sequential labeled sections so all tab variants (e.g. JS and Python) are visible with their labels, instead of being dropped or merged into unlabeled blocks
- rehypeFixCodeLanguage: copies the language-* class from Prism's `<pre>` to its child `<code>` so rehype-remark can detect the language and emit correct fenced code blocks (e.g. ```javascript instead of bare ```)

cc: @TC-MO 
Testable here: https://pr-2528.preview.docs.apify.com/platform/proxy.md

### DIFF
--- a/tools/utils/rehypeExpandTabs.js
+++ b/tools/utils/rehypeExpandTabs.js
@@ -13,9 +13,7 @@ function isTabsContainer(node) {
 }
 
 function buildReplacement(tabsContainer) {
-    const labels = selectAll('ul[role="tablist"] li[role="tab"]', tabsContainer).map((li) =>
-        toString(li).trim(),
-    );
+    const labels = selectAll('ul[role="tablist"] li[role="tab"]', tabsContainer).map((li) => toString(li).trim());
     const panels = selectAll('div[role="tabpanel"]', tabsContainer);
 
     const result = [];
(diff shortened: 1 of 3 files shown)


tools/utils/rehypeExpandTabs.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
+
+const { visit, SKIP } = require('unist-util-visit');
+const { selectAll } = require('hast-util-select');
+const { toString } = require('hast-util-to-string');
```
webrdaniel, line 5: Should these be added to package.json?

webrdaniel (approved):

PR #2534, opened 2026-05-14
chore: add images lint

### LINKED ISSUE #1549: Automate using webp for images to save repository size
Under https://github.com/apify/apify-docs/pull/1424, @vdusek suggested using webp would result in much smaller images. @B4nan agrees:

> The size of the repo is also important, if the difference is 5x let's just go with webp. We use them pretty much exclusively in the crawlee blog posts too for the same reason.
> – https://github.com/apify/apify-docs/pull/1424#issuecomment-2733898507

Since the images are already technically part of the repository, replacing them with optimized webp images would just add to the repo size right now. As this is not the first time we're discussing images under PRs, I think this topic deserves a tracking issue.

What is the preferred way to handling images? My take:

> I’m already putting new images through optimizt every time before commit, and it has an option to convert to webp. I just wonder if we could have it as a pre-commit so I don’t have to think about it. But changing extension breaks the path to the image, so I guess this must be a manual process… Unless I write myself a Python script to do the magic… 😅
> – https://github.com/apify/apify-docs/pull/1424#issuecomment-2734449062

Alternatively, the pre-commit would at least prevent me from committing other than webp images, so that I'm forced to do the manual work and not to forget about it.

The Docusaurus site should optimize the images automatically on deploy, so we're talking only about the repository size here – there should be no impact on viewers of the site, either way.

honzajavorek:
Fix https://github.com/apify/apify-docs/issues/1549


package.json
```diff
@@ -51,6 +52,7 @@
     "devDependencies": {
         "@apify/oxlint-config": "^0.2.5",
         "@apify/tsconfig": "^0.1.2",
+        "@funboxteam/optimizt": "^5.0.2",
```
B4nan: i am just wondering if we need this pretty fat dependency, isn't SVGO on its own (one of its deps) enough? that's what i've been using locally for image optimization

.github/workflows/test.yaml
```diff
@@ -275,3 +275,33 @@ jobs:
...
+
+            -   name: Fail on unoptimized images
+                if: steps.changed-files.outputs.any_changed == 'true'
+                env:
+                    UNOPTIMIZED_IMAGE_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
+                run: |
+                    echo "Unoptimized images detected! Please run the following commands:"
+                    echo ""
+                    while IFS= read -r file_path; do
+                      echo "npm run opt:images -- $file_path"
```
B4nan: we use pnpm nowadays, please test this locally, but afaik you shouldn't use that `--` thingy with pnpm anymore

```suggestion
                      echo "pnpm run opt:images $file_path"
```

package.json
honzajavorek: I thought SVGO optimizes SVGs? This takes bitmaps and converts them to webp (and minimizes them). If we don't want the dependency, we could keep the lint and just modify the suggestion in the message to this:

```
npm x -- @funboxteam/optimizt --webp path/to/your/images
```

(don't know pnpm alternative from the top of my head)

B4nan: Ok, maybe thats the only use case i used it for, ignore me then :D not a huge issue to include new dev dep here

B4nan:
Hmm thinking about it more, i would maybe skip this from the deps if its only supposed to be called locally. Can you run npx howfat … to see?

honzajavorek:
```
@funboxteam/optimizt@5.0.2 (727 deps, 20.9mb, 5511 files, ©MIT)
```

Includes:

```
svgo@2.6.1 (18 deps, 4.32mb, 503 files, ©MIT)
guetzli@5.0.0 (320 deps, 6.07mb, 2458 files, ©undefined)
sharp@0.30.7 (47 deps, 1.67mb, 424 files, NATIVE, ©Apache-2.0)
gifsicle@7.0.1 (366 deps, 5.92mb, 2836 files, ©MIT)
gif2webp-bin@5.0.0 (297 deps, 7.68mb, 2318 files, ©MIT)
...
```

Seems pretty fat. I tried to find a slim alternative just for conversion to WebP, but I didn't find anything nice (gave it 5mins) which would be mantained, packaged, and published to npm. Claude suggests that sharp is the go-to library and it has a sharp-cli package, so maybe that...

At the same time, one can easily use many C programs to do that, e.g. `magick input.png output.webp`, `cwebp input.png -o output.webp` or even `ffmpeg -i input.png output.webp`. But that's highly platform and installation dependent. The purpose of having it in dev deps should be that people don't have to think about it and just run it.

But that's without optimization. I think, for now, we can just leave it with `npx -- @funboxteam/optimizt`? I think the instructions in failed linter output will be straightforward. If someone adds a page and gets the message, can just copy-paste the command and run it. And people not adding images won't need it.

<small>When I think about it, I should also amend the message so that it warns people they'll need to update their markdown markup, too (from .png to .webp).</small>

B4nan:
Ok, I'd remove it from the project deps, and suggest `pnx` in that message so it installs globally instead (should be the same as `npx -y` basically). 

https://pnpm.io/cli/pnx

B4nan (approved):
the lockfile changes look a bit weird, are you using corepack or did you install some pnpm version manually? this usually happens when you use wrong version, corepack will resolve to the right one automatically.

otherwise looks good

honzajavorek:
- I found out the version of optimizt I was using was unmaintained and the new destination is `@343dev/optimizt`
- Using `pnx` now. I had to add `--config.onlyBuiltDependencies=*` otherwise for the first time it asks what to build (individual libraries for image formats) and user would have to think about it and select something or all of it.
- Because it was already too many cryptic arguments for a mere technical writer, I tried to keep it inside `opt:images` despite it's now `pnx`, and it works ✨

honzajavorek:
Hmm since I'm not adding or removing deps now, the lockfile should stay clean. I'll just checkout it with git. I think it's a leftover after a merge conflict I had with master.

Issue #950, opened 2024-04-26
Teach making requests in a consistent way


honzajavorek:
The courses now recommend using variety of tools to make HTTP requests. Sometimes it's more confusing, sometimes less.

- The [got](https://github.com/sindresorhus/got) library seems to be superseeded by [ky](https://github.com/sindresorhus/ky), at least in the README Sindre mentions it.
- Apify develops [got-scraping](https://github.com/apify/got-scraping), which seems to be married to the `got` library. Wouldn't it make sense to turn `got-scraping` into something more agnostic to a client library? Could it just prepare the request details, so that they can be attached to the request by any library?
- Some guides use [axios](https://github.com/axios/axios)
- Some guides mention [request](https://github.com/request/request/) and [request-promise](https://www.npmjs.com/package/request-promise), which are now both [deprecated](https://github.com/request/request/issues/3142)
- Meanwhile, Node.js has adopted [fetch to the stdlib](https://nodejs.org/docs/latest/api/globals.html#fetch)

Using `got-scraping` in the basic tutorial is probably unnecessary, any HTTP client can be used in the initial lessons. The value of `got-scraping` should emerge with more complicated use cases.

But wouldn't it make more sense to skip `got-scraping` and promote Crawlee right away at that point? Is `got-scraping` something Apify wants to spend marketing energy on, or is it an implementation detail?

As of now, `got-scraping` doesn't have good Python alternatives I'd know about. There are independent libraries one can use, such as [fake-user-agent](https://github.com/fake-useragent/fake-useragent), which have [integrations with scraping frameworks](https://github.com/alecxe/scrapy-fake-useragent).

Regarding request libraries, the scene is similarly shattered in Python, featuring [requests](https://github.com/psf/requests), [aiohttp](https://github.com/aio-libs/aiohttp), or [httpx](https://github.com/encode/httpx), each having their fans and use cases.

I'd like to kick off this as a discussion on what should be the preferred way for the Academy to teach making requests in 2024, using Node.js and Python.

honzajavorek:
@B4nan What HTTP client do you use in Python Crawlee under the hood - requests or httpx? Or is it built on top of aiohttp?

vdusek:
we use `httpx`

honzajavorek:
Cool, thanks! I thought about teaching making HTTP requests using httpx,
because it's very similar to widespread requests, but also has async and in
general, it has future 😅

PR #2535, opened 2026-05-14
chore: replace axios with fetch

honzajavorek:
When triaging issues and reading https://github.com/apify/apify-docs/issues/950, I searched "axios" in the docs repo to see in how many examples and tutorials it's actually used. And it's in a lot of them!

But I found it in dependencies, too. It is used on a single line in a changelog helper. In this PR I propose that we avoid the dependency and replace it with a simple native `fetch`.


B4nan (approved):
good idea

PR #2539, opened 2026-05-18
fix: skip alternate markdown link on /search page

TC-MO:
/search is in the llms-txt plugin's excludeRoutes, so no /search.md counterpart is generated. The Layout was still rendering <link rel="alternate" type="text/markdown" href="/search.md">, producing a daily lychee 404 report on docs.apify.com/search.md.

Skip the alternate link for /search, matching the existing handling for /404 and versioned pages.

### DIFF
--- a/apify-docs-theme/src/theme/Layout/index.jsx
+++ b/apify-docs-theme/src/theme/Layout/index.jsx
@@ -26,9 +26,12 @@ export default function LayoutWrapper(props) {
         pluginData.versions?.some((version) => !version.isLast && pathname.startsWith(version.path)),
     );
 
-    const shouldRenderAlternateLink = currentPath && currentPath !== '404' && !isVersionedPage;
+    const shouldRenderAlternateLink =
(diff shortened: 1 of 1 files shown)


apify-docs-theme/src/theme/Layout/index.jsx
```diff
@@ -26,9 +26,12 @@ export default function LayoutWrapper(props) {
         pluginData.versions?.some((version) => !version.isLast && pathname.startsWith(version.path)),
     );
 
-    const shouldRenderAlternateLink = currentPath && currentPath !== '404' && !isVersionedPage;
+    const shouldRenderAlternateLink =
+        currentPath && currentPath !== '404' && currentPath !== 'search' && !isVersionedPage;
```
barjin, line 30: nit: for the sake of extensibility, I'd rather have `const disableMarkdownPaths = new Set(['404', 'search', ...)` and `disableMarkdownPaths.has(currentPath)` condition here, but it's rather a personal preference.

barjin (approved):
lgtm, thank you @TC-MO !

PR #2542, opened 2026-05-18
feat: return 410 status code on missing JS assets

barjin:
As per the findings of the SEO team:

> The 404 JS files (e.g. docs.apify.com/assets/js/3b91d6c4.c7552f0b.js) are auto-generated during deployment. Every time the site is updated and published, these JS files are generated with new random names and the old ones are deleted. Googlebot had cached the old filenames and kept trying to fetch them after they were already gone, causing the 404s.
> Why we should solve this: Googlebot is wasting its crawl budget on docs.apify.com requesting JS files that no longer exist, causing an 8% 404 instead of crawling actual content pages.

Returning `410 Gone` instead of `404 Not Found` on regenerated JS assets should send a stronger signal that the asset path is no longer valid to the consumer.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -21,6 +21,18 @@ server {
     set $backend "https://apify.github.io/apify-docs";
     resolver 1.1.1.1 8.8.8.8 valid=30s ipv6=off;
 
+    # Stale Docusaurus chunks: when a deploy invalidates hashed JS bundles, signal
+    # search engine crawlers to drop cached references by returning 410 (Gone) instead of 404.
+    proxy_intercept_errors on;
+    error_page 404 = @maybe_gone_js_asset;
+
+    location @maybe_gone_js_asset {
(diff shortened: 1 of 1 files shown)


B4nan (approved):

barjin:
For the record, this doesn't seem to have helped (see, e.g., the response to https://docs.apify.com/assets/js/abc.js )

TC-MO:
I think nginx takes a while to start working properly (or at least it did for regular 301's I was doing so maybe here is the same case? I'll observe this for the next day/two and see how it behaves)

B4nan:
nginx is fast, but there is a CDN in front of it, which takes a bit. last time I was battling something here, I tried 7 different variants before surrending, only to see the other day that it actually worked.

barjin:
Okay, confirmed now. 

image

PR #2551, opened 2026-05-20
docs: correct regenerate-models workflow filename in AGENTS.md

vdusek:
## Summary

The Python API client model generation section referenced `gh workflow run regenerate_models.yaml`, but the actual workflow file in `apify-client-python` is named `manual_regenerate_models.yaml`. Caught while debugging dangling empty branches from that workflow.


TC-MO (approved):
LGTM

PR #2555, opened 2026-05-21
fix(openapi): align Actor pricing and version schemas with live API

vdusek:
The Pydantic models generated from this spec in `apify-client-python` reject real responses from `GET /v2/acts/{id}` and `GET /v2/acts/{id}/builds/default` (see [apify/apify-client-python#811](https://github.com/apify/apify-client-python/issues/811)). Three schema mismatches with the live API:

1. **`PricePerDatasetItemActorPricingInfo`** required `pricePerUnitUsd`, but Actors using tiered PPR pricing return `tieredPricing` instead.
2. **`ActorChargeEvent`** required `eventPriceUsd`, but tiered PPE events return `eventTieredPricingUsd`. The fields `isPrimaryEvent` and `isOneTimeEvent` were also missing.
3. **`ActorDefinition.version`** pattern `^[0-9]+\.[0-9]+$` rejected semver-style triplets like `0.0.1` that real actors return.

### DIFF
--- a/apify-api/openapi/components/schemas/actors/ActorDefinition.yaml
+++ b/apify-api/openapi/components/schemas/actors/ActorDefinition.yaml
@@ -11,8 +11,8 @@ properties:
     description: The name of the Actor.
   version:
     type: string
-    pattern: ^[0-9]+\.[0-9]+$
-    description: The version of the Actor, specified in the format [Number].[Number], e.g., 0.1, 1.0.
+    pattern: ^[0-9]+(\.[0-9]+)+$
+    description: The version of the Actor, typically a dot-separated sequence of numbers (e.g., `0.1`, `1.0`, or `0.0.1`).
   buildTag:
     type: string
     description: The tag name to be applied to a successful build of the Actor. Defaults to 'latest' if not specified.
--- a/apify-api/openapi/components/schemas/actor-pricing-info/ActorChargeEvent.yaml
+++ b/apify-api/openapi/components/schemas/actor-pricing-info/ActorChargeEvent.yaml
@@ -1,13 +1,30 @@
 title: ActorChargeEvent
+description: |
+  Definition of a single chargeable event for a pay-per-event Actor.
+  Each event is either flat-priced (`eventPriceUsd` is set) or
+  tier-priced (`eventTieredPricingUsd` is set); the two are mutually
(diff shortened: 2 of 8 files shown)


janbuchar (approved):

apify-api/openapi/components/schemas/actors/ActorDefinition.yaml
```diff
@@ -11,8 +11,8 @@ properties:
     description: The name of the Actor.
   version:
     type: string
-    pattern: ^[0-9]+\.[0-9]+$
-    description: The version of the Actor, specified in the format [Number].[Number], e.g., 0.1, 1.0.
+    pattern: ^[0-9]+(\.[0-9]+)+$
+    description: The version of the Actor, typically a dot-separated sequence of numbers (e.g., `0.1`, `1.0`, or `0.0.1`).
```
fnesveda, line 15: Actor versions can only ever be two dot-separated numbers, never three, and never anything else.

See https://github.com/apify/apify-core/blob/master/src/packages/actor/src/actors/actors.both.ts#L303-L310.

fnesveda (changes requested):
Looks good, except the Actor version number changes, those are not correct. The three-number strings are build numbers, never versions.

apify-api/openapi/components/schemas/actors/ActorDefinition.yaml
vdusek: Try to call this:

```sh
curl -s 'https://api.apify.com/v2/acts/apify~facebook-pages-scraper/builds/default' | jq -r '.data.actorDefinition.version'
0.0.1
```

Bigger part of the response:

```json
"actorDefinition": {
      "actorSpecification": 1,
      "name": "facebook-pages-scraper",
      "title": "Facebook Pages Scraper",
      "description": "Extract data from Facebook Pages.",
      "version": "0.0.1",
      "dockerfile": "../../../Dockerfile",
      "dockerContextDir": "../../..",
      "...": "..."
}
```

fnesveda: Ah sorry, this is the version from Actor definition, not version from Actor directly, my bad 🙈

fnesveda (approved):

PR #2566, opened 2026-05-25
feat: block scraping and rewrite canonical on preview builds

vladfrangu:
Adds a postBuild plugin that, on PR/preview deployments (detected via `APIFY_DOCS_ABSOLUTE_URL` hostname, same heuristic as the existing preview banner), overwrites `robots.txt` with a global `Disallow: /` and rewrites every page's canonical link to point at `https://docs.apify.com{path}`. Keeps previews out of search indexes and prevents them from competing with production URLs.

The second commit handles SWC's HTML minifier stripping attribute quotes and omitting `</head>` — without it the canonical rewrite would have shipped as dead code (caught by staff-review against the actual build output).


vladfrangu:
(staff-review was already ran)

B4nan (approved):

PR #2567, opened 2026-05-25
fix: Update and delete broken links in Actor tasks

szaganek:


fnesveda (approved):

barjin (commented):
Note that many other links in the API docs are still like this ⬇️ 

Image

I believe this was a way of dealing with legacy links after the OpenAPI plugin migration (see the client [redirect script](https://github.com/apify/apify-docs/blob/497090d6a2e49806c1ff2cff95c1f9ce939a23ce/apify-docs-theme/static/js/custom.js#L87)). 

Unfortunately, this process obfuscates the links for link checkers, so we don't get to see which links are actually broken (= we imo should eventually fix these links altogether).

PR #2578, opened 2026-05-27
feat: remove Airtable console integration

drobnikj:
We removed the Apify Console integration for Airtable - it was replaced by Actor. We will add a doc on how to use Actor in the following PR.


marcel-rbro (changes requested):
We should add a redirect (in `nginx.conf`), something like this:
```nginx
rewrite ^/platform/integrations/airtable/console$ /platform/integrations/airtable permanent;
```

marcel-rbro (approved):
LGTM 👍

protoss70 (approved):

PR #2584, opened 2026-05-28
fix(openapi): relax required fields on actor pricing info schemas

vdusek:
The platform-set `APIFY_ACTOR_PRICING_INFO` env var omits `apifyMarginPercentage`, `createdAt`, `startedAt` and `eventDescription`, so consumers deserializing it via the generated pydantic models hit validation errors and resort to injecting placeholder defaults. Marking these optional matches reality and is consistent with the sibling `CurrentPricingInfo` schema, which already has these fields optional.

Triggers a pydantic model regeneration in `apify-client-python`; once merged, the SDK can drop the placeholder-injection workaround in `_normalize_actor_pricing_info`.

### DIFF
--- a/apify-api/openapi/components/schemas/actor-pricing-info/ActorChargeEvent.yaml
+++ b/apify-api/openapi/components/schemas/actor-pricing-info/ActorChargeEvent.yaml
@@ -5,7 +5,6 @@ description: |
 type: object
 required:
   - eventTitle
-  - eventDescription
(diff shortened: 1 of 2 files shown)


vdusek:
Closing - these fields are written into Actor env vars by the worker, which strips them before serializing. The API itself always returns them, so the schema should stay as-is (and we should create custom models in the SDK instead).

PR #2596, opened 2026-05-28
feat: add navigation header to per-page .md files

### LINKED ISSUE #2557: Add context links to Markdown version of pages for AI agents
When you open a docs page in browser (e.g. https://docs.apify.com/platform/actors/running), you see:

- Left menu - parent pages and sub-pages
- Breadcrumbs
- Prev/next page button at the bottom

All these links give you a context where to find related content.

On the other hand, when you open the Markdown version (e.g. https://docs.apify.com/platform/actors/running.md), you only get the actual content, with no navigation context.

Let's get the navigation also to Markdown, e.g. as a structured skill-like header on top:

```
---
Page: [Running Actors](https://docs.apify.com/platform/actors/running.md)
Parents:
- [Apify documentation](https://docs.apify.com/llms.txt)
- [Apify Platform documentation](https://docs.apify.com/platform.md)
- [Actors](https://docs.apify.com/platform/actors.md)
Children: 
- [Actors in Store](https://docs.apify.com/platform/actors/running/actors-in-store.md)
  - ...
- [Input and output](https://docs.apify.com/platform/actors/running/input-and-output.md)
- ...
--- 
```

The children should contain the full hierarchical sub-menu of all content underneath the current page, so the AI agent can find any relevant sub-content in the next turn.

Make sure to skip this new heading in llms-full.txt, as that one already contains everything, so that we don’t confuse the agent.

marekh19:
Resolves #2557

Prepends a small frontmatter block to every per-page `.md` served by docs.apify.com so AI agents reading a single page can follow `parents` / `children` / `previous` / `next` links without having to fetch `llms.txt` first. Shape modeled on Vercel's docs (`vercel.com/docs/builds/build-features.md`): flat YAML-style frontmatter, no nested trees, links are inline `[Title](url.md)` so the agent can pick by title without a probing fetch.

Implemented as `scripts/addNavHeaders.mjs` in the `postbuild` chain. Runs after `@signalwire/docusaurus-plugin-llms-txt` writes both the per-page `.md` files and `llms-full.txt`, so editing only the per-page files leaves `llms-full.txt` untouched (the issue requires the header to stay out of it). Reads the real left-menu trees from `.docusaurus/docusaurus-plugin-content-docs/*/p/*.json`, so labels, ordering, and prev/next exactly match the live sidebar.

---

## A few design decisions worth recording:

**Header shape** — modeled on [Vercel's docs](https://vercel.com/docs/builds/build-features.md): flat YAML-ish frontmatter, links as inline `[Title](url.md)` so an agent can pick by title without a probing fetch.

**Children stay flat (direct only)**. A full recursive sub-tree explodes on exactly the pages an agent lands on first — `/api/v2` would carry 269 nested links, /platform/integrations 66, `/platform/actors` 57. That pollutes the header and bloats neighbouring content for no real gain: the agent can walk down one hop at a time via each page's own children.

**Parents go all the way up**. The opposite trade-off — a breadcrumb is a single chain bounded by site depth (~5 max), so there's no blow-up risk. Every page now chains llms.txt root → section landing → category ancestors, giving a reliable "zoom back out" path. This also fixes 83 top-level pages that previously had no parents at all.

Kept out of `llms-full.txt` for free by running in `postbuild`, after the llms-txt plugin has already written llms-full.txt from the same per-page content.

**previous/next** links mirror the site's prev/next buttons.

---

### Example 👀 

https://pr-2596.preview.docs.apify.com/platform/actors/running.md


webrdaniel (approved):

birosrichard (approved):

jancurn:
Looks great, thanks for looking into this. Could we add some simple integration test on a few pages to ensure this works?

TC-MO (approved):
LGTM

PR #2618, opened 2026-06-04
docs(x402): clarify the payment signature is reusable until balance runs out

MQ37:
## Context

The x402 page doesn't state that the signed `PAYMENT-SIGNATURE` can be reused across calls. A reader following the direct-API steps may assume each call needs a fresh signature and a new on-chain settlement.

## Solution

- Add a `:::note` in the direct-API section: the same `PAYMENT-SIGNATURE` stays valid while the prepaid balance lasts; resend it on every call with no re-signing or new settlement, and only sign again when the API returns `402` (balance exhausted) — the manual equivalent of what `mcpc` does automatically on the MCP path.
- Tighten the prepaid-balance bullet to say subsequent calls reuse the same signature.

### DIFF
--- a/sources/platform/integrations/ai/x402.md
+++ b/sources/platform/integrations/ai/x402.md
@@ -112,12 +112,18 @@ curl -s \
 
 You can replace `apify~instagram-post-scraper` with any supported Actor from [Apify Store](https://apify.com/store).
 
+:::note Reuse the same signature
+
+The signed `PAYMENT-SIGNATURE` stays valid while your prepaid balance lasts. Send it with every subsequent request - each call draws from the balance with no new signing and no new on-chain settlement. When the balance is exhausted, the API responds with `402` again; repeat step 2 to sign a fresh payment and continue.
+
+:::
+
 ## Pricing and refunds
 
 Both paths (MCP server and direct API) share the same pricing model:
 
 - Minimum transaction: $1 in USDC.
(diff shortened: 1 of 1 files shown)


sources/platform/integrations/ai/x402.md
```diff
@@ -112,12 +112,18 @@ curl -s \
 
 You can replace `apify~instagram-post-scraper` with any supported Actor from [Apify Store](https://apify.com/store).
 
+:::note Reuse the same signature
+
+The signed `PAYMENT-SIGNATURE` stays valid while your prepaid balance lasts. Send it with every subsequent request - each call draws from the balance with no new signing and no new on-chain settlement. When the balance is exhausted, the API responds with `402` again; repeat step 2 to sign a fresh payment and continue.
```
marcel-rbro: `repeat step 2 to sign a fresh payment and continue`

I'd specify what you need to do: "Sign the payment again to continue."

Also, consider moving the entire note to the How it works section instead of the example.

marcel-rbro (changes requested):

sources/platform/integrations/ai/x402.md
MQ37: Done in 6d9e0d97 — moved the note up into the direct-API **How it works** section and reworded the last sentence to "sign the payment again to continue."

marcel-rbro (approved):
🚀

PR #2636, opened 2026-06-11
chore: remove Vale from review-docs skill

TC-MO:
Vale cannot be shipped like mdlint, so invoking it per-file in this skill was dead weight. Vale still runs as a repo-level PR check and via the TW's local editor extension; the delegated standards review covers prose-style ground for the skill itself.


marcel-rbro (approved):
LGTM 👍

PR #2644, opened 2026-06-16
chore(deps): update pnpm to v11.7.0

B4nan:
Updates pnpm from v10.33.4 to v11.7.0, replacing Renovate's #2601.

Renovate only bumped `packageManager`, which `devEngines.packageManager` overrides, so CI stayed on v10. This bumps both version fields and handles two pnpm v11 behavior changes:

- pnpm v11 enables `strictDepBuilds` by default and fails the install on undeclared build scripts (`ERR_PNPM_IGNORED_BUILDS`) - the previously ignored build scripts are now declared via `allowBuilds` in `pnpm-workspace.yaml`.
- `devEngines.packageManager.onFail` is set to `error` to disallow other package managers; the lone `npm run` step in CI is switched to `pnpm` so it no longer trips `EBADDEVENGINES`. On pnpm v11 this also records the pnpm binary into `pnpm-lock.yaml` as a leading `packageManagerDependencies` document (intended behavior).


pnpm-workspace.yaml
```diff
@@ -31,3 +31,16 @@ preferWorkspacePackages: true
...
+# These packages ship prebuilt binaries or have non-essential lifecycle
+# scripts, so we keep them disabled as on pnpm v10.
+allowBuilds:
+  "@apify/ui-icons": false
+  "@parcel/watcher": false
+  "@swc/core": false
+  core-js: false
+  core-js-pure: false
+  postman-code-generators: false
+  protobufjs: false
```
vladfrangu, line 46: Pretty sure we do _not_ want to deny these but approve, no?

B4nan: surely not all of them, e.g. core-js just spams with a sponsor message. i would guess if something is needed, the CI would be red? the preview also looks good to me (but i did only a quick comparison)

B4nan: These all ran ignored throughout the pnpm 10 era (pnpm 10 only warned, pnpm 11 errors), so `false` keeps the behavior production has shipped on — `true` would newly enable build steps we've safely skipped for ages. Checked each:

- `core-js`/`core-js-pure`: postinstall is just the OpenCollective sponsor message.
- `@swc/core`, `@parcel/watcher`: native bindings come from prebuilt platform packages (`@swc/core-*`, `@parcel/watcher-*`); the build script only compiles from source as a fallback.
- `@apify/ui-icons`: its `create-barrel-file` output (`dist/index.js`, `index.css`, `index.d.ts`) already ships in the published package.
- `postman-code-generators`, `protobufjs`: postinstalls aren't needed for the generated docs — confirmed by green CI + the preview.

Nothing here is needed at build time; if it were, the Docs build would be red.

vladfrangu (approved):

PR #2646, opened 2026-06-16
feat: Rename legacy act-* API operationIds to actor-* for consistency

jancurn:
Renames all legacy `act_*`/`acts_*` operationIds so API reference URLs match the canonical `/v2/actors/` naming - e.g. `https://docs.apify.com/api/v2/acts-post` becomes `/api/v2/actors-post`, `/api/v2/act-get` becomes `/api/v2/actor-get`.

Because the account-wide `/v2/actor-runs` and `/v2/actor-builds` endpoints already own the `actor-runs-*`/`actor-builds-*` slugs, operations on an Actor's own `/runs` and `/builds` collections (including their deprecated per-item aliases) use the `actors-*` prefix instead, mirroring `/v2/actors/{actorId}/...` - e.g. `/api/v2/act-runs-post` becomes `/api/v2/actors-runs-post`. Deeper scoped sub-resources (`runs/last`, `run-sync`, versions, webhooks) keep the `actor-*` member prefix.

No `act-*` pages remain: every legacy slug 301-redirects via nginx (one explicit rule for the nine runs/builds slugs plus two catch-alls). Documentation links, code sample filenames, and the LLMs index exclusions are updated to match.

https://claude.ai/code/session_01Q4HrVBuRKYG46ztwLXQkHz


jancurn:
Sorry guys for this "drop". It's a lot of changes but should be fairly safe I think.

nginx.conf
```diff
@@ -790,6 +790,15 @@ server {
...
 
+    # Actor API reference operation pages: legacy `act-*`/`acts-*` slugs -> `actor-*`/`actors-*`.
+    # Each page slug is kebab-case(operationId); the `act_*`/`acts_*` operationIds were renamed to
+    # `actor_*`/`actors_*` to match the canonical `/v2/actors/` naming. The deprecated actor-scoped
+    # build/run duplicates (act-build-get, act-builds-get, act-build-abort-post, act-run-get,
+    # act-runs-get, act-run-abort-post, act-run-metamorph-post) resolve to their canonical
+    # standalone actor-build*/actor-run* pages via the same act- -> actor- rewrite.
+    rewrite ^/api/v2/acts-(.*)$ /api/v2/actors-$1 permanent;
+    rewrite ^/api/v2/act-(.*)$ /api/v2/actor-$1 permanent;
+
```
TC-MO: Just off the top of my head I think this is problematic, weird UX, and partly a bug.

The genuinely deprecated endpoints (act-build-get, act-run-get, abort, metamorph) are already marked `deprecated: true` in the spec, with an admonition that tells you they're deprecated and links to the canonical one. A silent 301 strips that context. You click the endpoint and inexplicably land on a different page with no explanation.

But `act-builds-get` and `act-runs-get` aren't deprecated at all. They're the Actor-scoped list endpoints (`GET /v2/acts/{actorId}/builds`, `GET /v2/acts/{actorId}/runs`). The broad rule at line 800 catches them and redirects to `actor-builds-get` / `actor-runs-get`, which are the account-wide "Get user builds/runs list" endpoints. 

Different scope, no equivalent in the new namespace. So that's a straight regression: those pages serve correct content on prod today and would 301 away after this.

TC-MO (approved):
From my POV it looks good now

sources/platform/integrations/workflows-and-notifications/bubble.md
```diff
@@ -248,22 +248,22 @@ The Apify plugin provides two main types of operations:
...
+- [List User Actors](https://docs.apify.com/api/v2/actors-get)
 - [List User Tasks](https://docs.apify.com/api/v2/actor-tasks-get)
-- [Scrape Single URL As File As Data](https://docs.apify.com/api/v2/act-run-sync-get-dataset-items-post)
-- [Scrape Single URL As Json As Data](https://docs.apify.com/api/v2/act-run-sync-get-dataset-items-post)
-- [List Actor Runs](https://docs.apify.com/api/v2/act-runs-get)
-- [List Specific Actor Runs](https://docs.apify.com/api/v2/act-runs-get)
+- [Scrape Single URL As File As Data](https://docs.apify.com/api/v2/actor-run-sync-get-dataset-items-post)
+- [Scrape Single URL As Json As Data](https://docs.apify.com/api/v2/actor-run-sync-get-dataset-items-post)
+- [List Actor Runs](https://docs.apify.com/api/v2/actor-runs-get)
+- [List Specific Actor Runs](https://docs.apify.com/api/v2/actor-runs-get)
```
marcel-rbro: These links now point to a different page, not sure if that would be a problem, don't know enough about this integration.

https://docs.apify.com/api/v2/act-runs-get (runs of a specific actor)
vs.
https://docs.apify.com/api/v2/actor-runs-get (all runs by a user)

janbuchar:
> Sorry guys for this "drop". It's a lot of changes but should be fairly safe I think.

It's great that we're getting rid of the `act_` terminology, that's been long overdue. One concern though:

image

The `operationId` property determines the URL in the rendered API docs, and it's not cool to break links to tens of endpoints. I'm pretty sure that the `x-legacy-doc-urls` property on the Operation object should make it fairly easy to add redirects for compatibility.

**EDIT**: Now I noticed this in the PR description, sorry, my bad.

> - Added nginx rewrite rules to redirect legacy `act-*` and `acts-*` URL slugs to their `actor-*` and `actors-*` equivalents with 301 permanent redirects

Still, I'm not sure we want to have two parallel ways to handle redirects from deprecated URLs. I think we should either unify this right away or make a follow up issue.

janbuchar (approved):
I voiced my concern [earlier](https://github.com/apify/apify-docs/pull/2646#issuecomment-4768256212), but it's not a reason to block this PR — it's fine from an openapi quality standpoint.

TC-MO:
Wasn't the `x-legacy-doc-urls` just used for transition from API Blueprint, while in general redirects were always handled through nginx?

janbuchar:
> Wasn't the `x-legacy-doc-urls` just used for transition from API Blueprint, while in general redirects were always handled through nginx?

That sounds plausible. Maybe because `nginx` can't see the URL fragment?

fnesveda (approved):

jancurn:
Thanks for the reviews, but pls let's not merge this yet, i still need to resolve the conflict between `/actors/{x}/runs` and  `/actor-runs` in IDs

jancurn:
I've redone this with Fable, fixed the duplicates, and this should be good to go. Would be great if somebody else could check this, but in worst case, it only affects docs :)

PR #2658, opened 2026-06-19
refactor: convert proxy type page blockquotes to admonitions

TC-MO:
Convert informal `> ...` blockquotes on the datacenter, residential, and Google SERP proxy pages into titled admonitions (:::note, :::tip, :::caution).

### DIFF
--- a/sources/platform/proxy/datacenter_proxy.md
+++ b/sources/platform/proxy/datacenter_proxy.md
@@ -311,7 +311,7 @@ You can find your proxy password on the [Proxy page](https://console.apify.com/p
 
 :::note Username field
 
-The `username` field is _not_ your Apify username. Instead, you specify proxy settings (for example, `groups-BUYPROXIES94952`, `session-123`). Use `auto` for default settings.
+Use the `username` field to specify your proxy settings (for example, `groups-BUYPROXIES94952`, `session-123`), or set it to `auto` for default settings. It is not your Apify username.
 
 :::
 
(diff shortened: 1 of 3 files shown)


sources/platform/proxy/datacenter_proxy.md
```diff
@@ -309,9 +309,11 @@ await Actor.exit();
 
 You can find your proxy password on the [Proxy page](https://console.apify.com/proxy) of Apify Console.
 
-> The `username` field is **not** your Apify username.<br/>
-> Instead, you specify proxy settings (e.g. `groups-BUYPROXIES94952`, `session-123`).<br/>
-> Use `auto` for default settings.
+:::note Username field
+
+The `username` field is _not_ your Apify username. Instead, you specify proxy settings (for example, `groups-BUYPROXIES94952`, `session-123`). Use `auto` for default settings.
```
szaganek: I'm confused. Does it want to say that you use `username` field to specify proxy settings?

TC-MO: Basically `username` despite its unfortunate name has nothing to do with actual username... (see [here](https://docs.apify.com/platform/proxy/usage#external-connection))


Perhaps rephrasing to lead with what it is actually instead of what isn't will be more clear

```suggestion
Use the `username` field to specify your proxy settings (for example, `groups-BUYPROXIES94952`, `session-123`), or set it to `auto` for default settings. It is not your Apify username.
```

And same rewrite on SERP page.

szaganek: Yup, the new version explains it in much more clear way

TC-MO: Will rewrite both of those in that way then 💪

szaganek (approved):

PR #2659, opened 2026-06-19
fix: Stop truncating headings in TOC

szaganek:
Truncating headings in the table of contents makes it basically useless. We don't even provide a tooltip to see the whole thing (and that would be annoying in its own way). 

Before ([doc](https://docs.apify.com/platform/actors/publishing/monetize/pay-per-event)):
Screenshot 2026-06-19 at 10 11 57

Now:
Screenshot 2026-06-19 at 10 12 20

On mobile:
Screenshot 2026-06-19 at 10 12 48

### DIFF
--- a/apify-docs-theme/src/theme/custom.css
+++ b/apify-docs-theme/src/theme/custom.css
@@ -679,7 +679,6 @@ article .card h2 {
     display: inline-block !important;
     width: 100%;
     overflow: hidden;
-    white-space: nowrap;
 }
 
 .tsd-flag {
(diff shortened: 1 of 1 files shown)


barjin (approved):
We could also make some of the headings shorter 😅 But the changes look good to me overall.

Thanks, @szaganek !

szaganek:
> We could also make some of the headings shorter 

Definitely 😅 Tech writers will take it from here to make sure we don't cram our TOCs with fluff.

PR #2680, opened 2026-06-23
docs(mcp): mention CLI install command

MQ37:
## What
Adds the `apify mcp install <client>` setup path to the MCP server page's Client configuration section.

## Why
The command from apify/apify-cli#1145 gives users a direct CLI-based way to configure supported local MCP clients.

## Testing
- `/home/mq/.local/share/mise/installs/pnpm/latest/pnpm --dir /home/mq/workdir/apify/apify-docs exec markdownlint sources/platform/integrations/ai/mcp.md`
- `git -C /home/mq/workdir/apify/apify-docs diff --check`

### DIFF
--- a/sources/platform/integrations/ai/mcp.md
+++ b/sources/platform/integrations/ai/mcp.md
@@ -104,24 +104,6 @@ _MCP server configuration for other clients_: Use the [UI configuration tool](ht
 
 Here's how to add the Apify MCP server to popular text editors and AI assistants:
 
-:::tip Configure with Apify CLI
-
-Run `apify mcp install <client>` to add the Apify MCP server to a supported local client. Replace `<client>` with `claude-code`, `cursor`, `vscode`, `vscode-insiders`, `codex`, `kiro`, or `antigravity`.
-
-For clients that need a token, the command uses the API token from `apify login`. Pass `--token <APIFY_TOKEN>` to use a specific token.
-
-```bash
-apify mcp install cursor
-```
-
-Use `--tools` to expose only selected tools or Actors:
-
-```bash
-apify mcp install vscode --tools search-actors,apify/rag-web-browser
-```
-
-:::
-
 <Tabs>
 <TabItem value="cursor" label="Cursor">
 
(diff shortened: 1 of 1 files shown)


sources/platform/integrations/ai/mcp.md
```diff
@@ -104,6 +104,24 @@ _MCP server configuration for other clients_: Use the [UI configuration tool](ht
 
 Here's how to add the Apify MCP server to popular text editors and AI assistants:
 
+:::tip Configure with Apify CLI
+
+Run `apify mcp install <client>` to add the Apify MCP server to a supported local client. Replace `<client>` with `claude-code`, `cursor`, `vscode`, `vscode-insiders`, `codex`, `kiro`, or `antigravity`.
+
+For clients that need a token, the command uses the API token from `apify login`. Pass `--token <APIFY_TOKEN>` to use a specific token.
```
jirispilka: "For clients that need a token" - what are you trying to say by this?

jirispilka (commented):
Thank you.

Suggestion: 

It would be nice to tell users what's get created. 

This command will create a json, user level config, saved here ... if you want to edit it, you can do it manually or run the `apify mcp install` with different parameters. 

Can we add something in that spirit?

sources/platform/integrations/ai/mcp.md
MQ37: Some clients do support OAuth and some don't so in that case the user can use the `--token` and they can also pass a different token other than from the `apify login` if needed.

sources/platform/integrations/ai/mcp.md
```diff
@@ -104,6 +104,24 @@ _MCP server configuration for other clients_: Use the [UI configuration tool](ht
 
 Here's how to add the Apify MCP server to popular text editors and AI assistants:
 
+:::tip Configure with Apify CLI
```
marcel-rbro: Wouldn't it be better to just a new tab to the tab component? There are way too many tips and notes on this page already.

sources/platform/integrations/ai/mcp.md
marcel-rbro: Would be nice to also explain this in the docs?

sources/platform/integrations/ai/mcp.md
```diff
@@ -104,6 +104,24 @@ _MCP server configuration for other clients_: Use the [UI configuration tool](ht
...
+
+Run `apify mcp install <client>` to add the Apify MCP server to a supported local client. Replace `<client>` with `claude-code`, `cursor`, `vscode`, `vscode-insiders`, `codex`, `kiro`, or `antigravity`.
```
marcel-rbro: Is this the complete list of available clients? Would be nice to state that clearly:

"Available clients are:"

sources/platform/integrations/ai/mcp.md
marcel-rbro: Alternativaly, this could just be a new sub-heading

marcel-rbro (changes requested):

sources/platform/integrations/ai/mcp.md
MQ37: I reworded the sentence

sources/platform/integrations/ai/mcp.md
MQ37: I like the tab component idea, let's go with that

sources/platform/integrations/ai/mcp.md
MQ37: Yes, I will explicitly use "Available clients are:"

MQ37:
Thank you for the review @jirispilka and @marcel-rbro.

I refactored the whole section and it is now using tab component as @marcel-rbro suggested.

> This command will create a json, user level config, saved here ... if you want to edit it, you can do it manually or run the apify mcp install with different parameters.

@jirispilka I added generic explanation what it does without mentioning specific paths as I think the user that wants to manually edit files should be able to use the internet to look up the path for their specific client they are using - also adding specific paths would be more burden to maintain that could also get outdated.

marcel-rbro (approved):
Looks good now 👍

PR #2697, opened 2026-06-26
docs(openapi): Document `origin` query parameter for last run endpoints

Pijukatel:
The last run and last task run endpoints accept an `origin` query parameter to filter the last run by how it was started (e.g. `origin=API`), but it was missing from the OpenAPI spec. This adds it to those endpoints and all their convenience variants (default dataset/key-value store/request queue/log, plus abort/metamorph/reboot), reusing the existing `RunOrigin` schema for the enum.

Surfaced by apify/apify-client-js#952; API source: [actors/last_run.ts](https://github.com/apify/apify-core/blob/v0.1508.0/src/api/src/routes/actors/last_run.ts#L21) and [actor_tasks/last_run.ts](https://github.com/apify/apify-core/blob/v0.1508.0/src/api/src/routes/actor_tasks/last_run.ts#L21).

### DIFF
--- a/apify-api/openapi/components/parameters/runAndBuildParameters.yaml
+++ b/apify-api/openapi/components/parameters/runAndBuildParameters.yaml
@@ -98,6 +98,15 @@ statusLastRun:
     type: string
     example: SUCCEEDED
 
+originLastRun:
+  name: origin
+  in: query
+  description: Filter for the run origin, i.e. the means by which the run was started.
+  style: form
+  explode: true
+  schema:
+    $ref: ../schemas/actor-runs/RunOrigin.yaml
+
 timeout: &timeout
   name: timeout
   in: query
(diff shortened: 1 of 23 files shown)


apify-api/openapi/components/parameters/runAndBuildParameters.yaml
```diff
@@ -98,6 +98,15 @@ statusLastRun:
     type: string
     example: SUCCEEDED
 
+originLastRun:
```
janbuchar: Suggestion — maybe the `origin*` parameter definitions could be moved to a separate file such as `lastRunParameters`.

janbuchar (approved):
Fascinating, LGTM

apify-api/openapi/components/parameters/runAndBuildParameters.yaml
Pijukatel: Done

PR #2712, opened 2026-07-01
feat: unblocker proxy docs

daniil-poletaev:
Public facing documentation for Unblocker proxy.


bliuchak (approved):

sources/platform/proxy/unblocker_proxy.md
```diff
@@ -0,0 +1,118 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+Unblocker proxy is designed for scraping websites protected by anti-bot and anti-captcha systems. Instead of building and maintaining your own detection and bypass logic, you can route requests through Unblocker proxy, which automatically handles many common protection mechanisms.
+
+This makes it a good choice for accessing websites with advanced bot protection while keeping your scraping code simple.
+
+**Pricing is based on units.** Each request through Unblocker proxy is currently billed at 10 units.
```
TC-MO: What are these units? Without an explicit definition readers will assume compute units, since that's the closest existing term on the platform. At $ 0.2CU that would mean 2$ per request, is that the intention?

Two other things in this line:
- Drop currently. Hedging dates the sentence silently, and nothing downstream tells us when it stopped being true.
- Bold is reserved for UI elements 

I can propose rewrite once we establish what units are we referring to

sources/platform/proxy/unblocker_proxy.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Connect to Unblocker proxy
+
+Connecting to Unblocker proxy works the same way as [datacenter proxy](./datacenter_proxy.md), with one difference: the `groups` [username parameter](./index.md#username-parameters) should always specify `UNBLOCKER`.
```
TC-MO: Which is it :D Is is the same or is there a difference ^^?  Moreover isn't it actually 2 differences?

1. groups should always specify UNBLOCKER
2. Country is set automatically

so maybe something like:


```suggestion
Connecting to Unblocker proxy works the same way as [datacenter proxy](./datacenter_proxy.md), with two differences:

- The `groups` [username parameter](./index.md#username-parameters) must always specify `UNBLOCKER`.
- Unblocker proxy selects the country automatically, so you rarely need to set `country`.
```

Furthermore question what about session persistence? Does Unblocker support that? If it does we should have similar session persistence section as other pages if not we need to call it out explicitly.

sources/platform/proxy/index.md
```diff
@@ -190,6 +196,7 @@ The table below describes the available parameters.
             <br/>- <code>groups-[group name]</code> or <code>auto</code> when using datacenter proxies.
             <br/>- <code>groups-RESIDENTIAL</code> when using residential proxies.
             <br/>- <code>groups-GOOGLE_SERP</code> when using Google SERP proxies.
+            <br/>- <code>groups-UNBLOCKER</code> when using <a href="./unblocker_proxy">Unblocker proxy</a>.
```
TC-MO:
```suggestion
            <br/>- <code>groups-UNBLOCKER</code> when using [Unblocker proxy](./unblocker_proxy.md).
```

sources/platform/proxy/unblocker_proxy.md
```diff
@@ -0,0 +1,118 @@
...
+
+When using [standard libraries and languages](./datacenter_proxy.md), specify the `groups` parameter in the [username](./index.md#username-parameters) as `groups-UNBLOCKER`.
+
+For example, your **proxy URL** when using the [got-scraping](https://www.npmjs.com/package/got-scraping) JavaScript library will look like this:
+
+```js
+const proxyUrl = 'http://groups-UNBLOCKER:<YOUR_PROXY_PASSWORD>@proxy.apify.com:8000';
+```
+
+In the [Apify SDK](/sdk) you set the **groups** in your proxy configuration:
```
TC-MO: bold is reserved for UI elements.

```suggestion
In the [Apify SDK](/sdk) you set `groups` in your proxy configuration:
```

sources/platform/proxy/unblocker_proxy.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Connect to Unblocker proxy
+
+Connecting to Unblocker proxy works the same way as [datacenter proxy](./datacenter_proxy.md), with one difference: the `groups` [username parameter](./index.md#username-parameters) should always specify `UNBLOCKER`.
+
+### How to set a proxy group
+
+When using [standard libraries and languages](./datacenter_proxy.md), specify the `groups` parameter in the [username](./index.md#username-parameters) as `groups-UNBLOCKER`.
+
+For example, your **proxy URL** when using the [got-scraping](https://www.npmjs.com/package/got-scraping) JavaScript library will look like this:
```
TC-MO: We are talking here about a variable from the library here right?

```suggestion
For example, your `proxyURL` when using the [got-scraping](https://www.npmjs.com/package/got-scraping) JavaScript library will look like this:
```

TC-MO: bold is reserved for UI elements

```suggestion
For example, your proxy URL when using the [got-scraping](https://www.npmjs.com/package/got-scraping) JavaScript library will look like this:
```

TC-MO (changes requested):

TC-MO:
BTW this PR was opened before we made significant changes to Information Architecture, it would be good to reconcile it with `master` so we don't accidently introduce wrong URLs etc.

sources/platform/proxy/unblocker_proxy.md
bliuchak: No, we meant here **Unblocker Units**. It's a separate thing than CU.

TC-MO: Then we need to explain it better since they are not explained anywhere else.

sources/platform/proxy/unblocker_proxy.md
```diff
@@ -0,0 +1,123 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+Unblocker proxy is designed for scraping websites protected by anti-bot and anti-captcha systems. Instead of building and maintaining your own detection and bypass logic, you can route requests through Unblocker proxy, which automatically handles many common protection mechanisms.
+
+This makes it a good choice for accessing websites with advanced bot protection while keeping your scraping code simple.
+
+Unblocker proxy uses its own pricing unit, separate from [compute units](/actors/running/usage-and-resources#what-is-a-compute-unit). Pricing is per 1,000 units, and the rate depends on your subscription plan. Each request is billed 10 units.
```
TC-MO: Original is pretty good, here is a slightly extended version that names the units explicitly. 

3 things I need to confirm if we would like to go with that version

1. Is it gonna be Unblocker units, or unblocker units in UI?
2. I assume the consumption will be shown in proxy dashboard as with other proxies?
3. Is it every successful requests that consumes the 10 units? Or every request?

```suggestion
Pricing is based on Unblocker units, which are separate from [compute units](/actors/running/usage-and-resources#what-is-a-compute-unit). Each request costs 10 units. Units are priced per 1,000 at a rate that depends on your plan, and your consumption is displayed on your [proxy usage dashboard](https://console.apify.com/proxy/usage) in Apify Console.
```

TC-MO (approved):
General approve, slightly extended version provided for the pricing excerpt but needs confirmation first, also not sure why PR toolkit is failing ¯\_(ツ)_/¯

sources/platform/proxy/unblocker_proxy.md
daniil-poletaev: 1. I think Unblocker units in UI
2. Yes, it will be shown in proxy dashboard as amount of units used
3. Every successfull request is charged 10 units

TC-MO: then we can do it like this

```suggestion
Pricing is based on Unblocker units, which are separate from [compute units](/actors/running/usage-and-resources#what-is-a-compute-unit). Each successful request costs 10 units. Units are priced per 1,000 at a rate that depends on your plan, and your consumption is displayed on your [proxy usage dashboard](https://console.apify.com/proxy/usage) in Apify Console.
```

bliuchak (approved):

PR #2720, opened 2026-07-02
docs(x402): drop allowsAgenticUsers store filter

MQ37:
## What
Remove all mentions of the `?allowsAgenticUsers=true` store query parameter from the x402 integration page. Store discovery examples now use plain store queries, and the paragraph describing the curated agentic-Actor filter is dropped.

## Why
We switched to the `agi.apify.com` approach for agentic payments. Filtering the store to only agent-approved Actors is no longer needed, so the query parameter is obsolete. The Actor eligibility requirements (pay-per-event, limited permissions, no Standby) remain documented.

## Testing
Docs-only change. Verified no `allowsAgenticUsers` references remain in the page via grep.

### DIFF
--- a/sources/platform/integrations/ai/x402.md
+++ b/sources/platform/integrations/ai/x402.md
@@ -98,11 +98,11 @@ When using the API, replace the `/` in the Actor name with `~` (for example, `ap
 :::
 
 ```bash
-# Discover Actors approved for agentic payments
-curl -s "https://api.apify.com/v2/store?allowsAgenticUsers=true"
+# Discover Actors in the store
+curl -s "https://api.apify.com/v2/store"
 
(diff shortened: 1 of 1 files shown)


marcel-rbro (approved):
👍

stepskop (approved):
👍 ...maybe a bit out of scope, but ideally we should migrate the Skyfire/KYAPay method to `agi` as well and remove this even from there.

MQ37:
> 👍 ...maybe a bit out of scope, but ideally we should migrate the Skyfire/KYAPay method to `agi` as well and remove this even from there.

First we need to launch the `batch-settlement` and then we can also migrate the Skyfire, but I agree that ideall we should move all to agi.apify.com.

Issue #2684, opened 2026-06-24
`Actors` section


szaganek:
Complete missing descriptions, improve endpoint descriptions, review links.

PR #2731, opened 2026-07-03
docs(openapi): Add missing descriptions to Update Actor endpoint

szaganek:
- Adds missing descriptions to `Create Actor` request body.
- Cleans up endpoint description.
- Also fixes the `Unknown type error` in the preview.

Part of #2684.

### DIFF
--- a/apify-api/openapi/components/schemas/actors/CreateOrUpdateVersionRequest.yaml
+++ b/apify-api/openapi/components/schemas/actors/CreateOrUpdateVersionRequest.yaml
@@ -5,29 +5,37 @@ properties:
     type: [string, "null"]
     pattern: ^([0-9]|[1-9][0-9])\.([0-9]|[1-9][0-9])$
     examples: ["0.0"]
+    description: The version number of the Actor. Must be a dot-separated sequence of numbers.
   sourceType:
     anyOf:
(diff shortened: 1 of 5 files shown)


apify-api/openapi/components/schemas/actors/CreateOrUpdateVersionRequest.yaml
```diff
@@ -5,29 +5,37 @@ properties:
     type: [string, "null"]
     pattern: ^([0-9]|[1-9][0-9])\.([0-9]|[1-9][0-9])$
     examples: ["0.0"]
+    description: The version number of the Actor. Must be a dot-separated sequence of numbers.
```
mixalturek: Do I understand it correctly? Is it this?

Image

```suggestion
    description: The version number of the Actor. Must be a dot-separated sequence of two numbers which are the major and minor parts of major.minor.patch semantic versioning.
```

mixalturek (approved):
Thanks a lot!

apify-api/openapi/components/schemas/actors/CreateOrUpdateVersionRequest.yaml
szaganek: Good catch, I clarified it in one more place.

PR #2713, opened 2026-07-01
feat: revert " flip to the v3 information architecture (#2702)"

TC-MO:
This reverts commit e4463fba33a43f188c597c4e5e148e4e45fb1aa7.

PR #2702, opened 2026-06-29
feat: flip to the v3 information architecture

TC-MO:
Switches the docs from the `/platform/*` URL structure to the v3 information architecture: topic-first sections at the site root, `routeBasePath` dropped to `/`, and 301 redirects covering old URLs.

### Review by commit

The flip is categorized by concern; commits 1-2 are mechanical (bulk moves and link rewrites) and can be skimmed, the judgment lives in 3-6. Commits 7-10 are refinements from reviewing the live preview.

1. **restructure into the v3 IA** - the moves: console and collaboration fold into Account, schedules and monitoring move under Actors > running; adds the Account hub and the Get Started section. Check the destinations make sense.
2. **drop the /platform prefix from internal links** - 157 files, move-aware (relocated sections point at their new paths). Bulk rewrite; the OpenAPI source and React pages are included alongside content.
3. **flip routeBasePath + v3 navbar** - the actual routing switch, the nav items, Open source moving from nav dropdown to footer, and per-section sidebars. The core change to scrutinize.
4. **scope theme doc-content and search to the v3 roots** - `DocItemContent` allowed paths and the `SearchBar` instance check.
5. **remove the platform landing page** - the site root is the new entry point.
6. **redirect /platform/\* URLs** - nginx catch-all plus structural pre-rules for the relocated sections.
7. **label v3 section overviews** - the Actors, Storage, Proxy, and Integrations hub pages now read "Overview" and sort first.
8. **footer dedup** - drop the duplicate Open source link now that there is a dedicated Open source column.
9. **split Get Started into Agent onboarding and a Build group** - move "Apify for AI agents" out of Integrations into Get Started, and group the four build pages under an expanded Build category.
10. **remove the duplicate quick start tree** - it duplicated the new Get Started build pages; its unique Technology stack intro moved into the Actor development overview.

### Verification

- Build green at the tip, 0 broken links. The broken-anchor warnings are pre-existing and non-blocking (`onBrokenAnchors: 'warn'`).

- **nginx double-hop cleanup** - stripping `/platform/` from the ~135 existing rule targets is a separate post-flip PR; those URLs still resolve via the catch-all in the meantime, just with an extra hop.

PR #2745, opened 2026-07-07
feat: ia v3 flip

TC-MO:
Re-applies the IA v3 flip (`/platform/*` → root) after #2702 shipped, and was reverted (#2713). The root cause was nginx deploy propagation, not this config and has since been fixed.


barjin (approved):
Approving, since this is the same as https://github.com/apify/apify-docs/pull/2702.

A sidenote - I'd appreciate a more verbose PR title :) (although it doesn't really matter here in the docs)

PR #2758, opened 2026-07-10
fix: only show subnavbar in mobile menu when path matches its regex

### LINKED ISSUE #2757: The Academy subnavbar is rendered everywhere in responsive mode
Unlike the desktop mode, the mobile view renders the Academy subnavbar on non-academy pages, leading to confusion.

Image

We should only render the `Academy / Courses / Tutorials` part when in the `Academy` category.

barjin:
Mobile menu ignored subNavbar.pathRegex, unlike desktop, so it showed on every page.

Fixes #2757


TC-MO (approved):
LGTM

PR #2759, opened 2026-07-10
feat(docs): change code-block component background [internal]

birosrichard:
* changed the code block background color

Testable here: https://pr-2759.preview.docs.apify.com/actors/running


webrdaniel (approved):
✨

PR #2761, opened 2026-07-10
feat: split navbar into separate API, SDK, and CLI items

TC-MO:
More context in [this](https://apify.slack.com/archives/CQ96RHG2U/p1783680924152209) slack thread

Replace the single "APIs, SDKs & CLI" dropdown with three right-side items: API and SDK as hover dropdowns, CLI as a plain link. Each label links to its landing page (/api, /sdk, /cli) while API/SDK still reveal their sub-links on hover, and the API dropdown groups the experimental clients under a labeled subsection.

Override `DropdownNavbarItem/Desktop` so a dropdown label navigates on href (stock only navigates on to), since the shared navbar uses absolute hrefs for cross-repo links.


barjin (commented):
The design doesn't seem to work correctly in responsive mode (missing dropdown labels). Also, the `Experimental clients` label styling seems a little out of place.

Image

TC-MO:
Thanks @barjin, after all we are moving with simpler approach (but keeping this open for a while)

jancurn:
Here's another way how to do it: https://github.com/apify/apify-docs/pull/2819

TC-MO:
Superseded by #2819

PR #2762, opened 2026-07-12
docs: align list of permissions page with Console permissions modal

### LINKED ISSUE #2696: Review and update the list of available permissions
From a [Slack message](https://apify.slack.com/archives/C010Q0FBYG3/p1782317848437559).

- https://docs.apify.com/platform/collaboration/list-of-permissions
- https://docs.apify.com/platform/collaboration/organization-account/setup

Docs contain outdated screenshots (that can be most probably deleted completely) + miss info on recently added permissions.

TC-MO:
Closes https://github.com/apify/apify-docs/issues/2696


szaganek (approved):
Hm, just wondering here, because this repetitive `To learn about X, see its documentation` feels a lil heavy. Wouldn't it look better if the X was linked in the intro sentence above the table, without all this extra wording 🤔 Just an idea.

tobice (changes requested):
Hot take: Is there any value in keeping this documentation at all? This all is already available through the UI and the page brings very little added value.

One reason could be AI, which cannot access Console (yet), so at least it'd be aware of what is possible and could inform the user. But I'm not sure it's worth maintaining a page for, plus if we expose this functionality to agents, it'll be via a new, well documented API.

TC-MO:
1. Console just exposes the UI with no explanation. Sure you can infer it from context. But at the same time Console sends out users to specifically this page if you want to learn more about permissions
2. AI source of truth in 95% cases is docs.
3. If and when such functionality would be exposed, I'm happy to revisit this topic. But before that I think the page is warranted.

tobice:
> Console just exposes the UI with no explanation.

It does in the share dialog.

image

It doesn't in the organization role configuration. Adding the descriptions there sounds like a better solution, but also more work, so keeping the page makes sense :smile:

TC-MO:
@tobice I think I need your approval since you requested changes to merge that PR :) (in addition to resolving conflicts)

tobice (approved):

PR #2764, opened 2026-07-13
docs: add missing `TEST`, `CI`, and `MCP` run origins to origin table

barjin:
Updates the origin table with the new origin types (see [apify-shared-js](https://github.com/apify/apify-shared-js/blob/84114f76085c30ae780798f23cd9c8f9e3e6a8f2/packages/consts/src/consts.ts#L67) for source).

[Discussed on Slack](https://apify.slack.com/archives/CGZSN9DQC/p1783938506578589)


TC-MO (approved):
LGTM

PR #2765, opened 2026-07-13
feat: point navbar dropdown to API, SDK, and CLI landing pages

TC-MO:
Replace the eleven items in the "APIs, SDKs & CLI" dropdown with three links to the section landing pages (/api, /sdk, /cli); the detailed links now live on those pages. Extend activeBaseRegex with (/|$) so the dropdown also highlights as active on the bare landing-page URLs, not only their sub-pages.

@barjin this will supersede #2761


apify-docs-theme/src/config.js
```diff
@@ -86,72 +86,24 @@ const themeConfig = {
             {
                 label: 'APIs, SDKs & CLI',
                 type: 'dropdown',
-                activeBaseRegex: '^/(api|sdk|cli)/',
+                activeBaseRegex: '^/(api|sdk|cli)(/|$)',
```
barjin, line 89: This doesn't seem to work. See workarounds under https://github.com/facebook/docusaurus/issues/5954 (setting `activeClassName: 'navbar__link--active'`, not sure if it still works).

barjin (approved):
Thank you @TC-MO !

I'll be honest - I don't really love the UI 😅 

Image]

The navbar also worked as a language switcher between JS / Python SDKs or Clients. Now, if I accidentally land on a JS SDK page (when looking for Python), I'll have to make a loop through the landing page, which adds friction.

Anyway, the code looks fine (aside from the comment), I'll leave the UI to you guys. Approving.

TC-MO:
Thanks for comments @barjin. Good points, I think we are now trying to work within constraints of current UI, and we will definitely be making bigger redesigns and account for language switching.

TC-MO:
BTW @barjin is there a way then where we could preserve the highlight on the dropdown if on `/sdk` /`cli` or /`api` ?

barjin:
> is there a way then where we could preserve the highlight

I believe that's what I mentioned here, no?

https://github.com/apify/apify-docs/pull/2765#discussion_r3593974688

I'm not sure the ways described there do still work, but the thread seems authoritative enough. 

Lmk if that doesn't work, we can definitely try hacking it somehow

TC-MO:
oh right 🤦 my bad still hazy after catching up after a hiatus :D will try this, if it doesn't work I'll let you know

PR #2770, opened 2026-07-15
docs: Update best practices for PPE Actors and mention negative profit

### LINKED ISSUE #2634: Clarify how to deal with negative profit
Update https://docs.apify.com/platform/actors/publishing/monetize 
- Add more thorough explanations to Actor events, memory limits, respecting user spending limits
- Add some kind of troubleshooting section that explains possible reasons for negative profit and how to deal with it
- Split the topic into more manageable chunks

szaganek:
- Adds a section on events
- Adds a section on graceful abort
- Adds a section on avoiding negative profit
- Extends the section on respecting user limits 
- Removes fluff and repeated info 

Preview: https://pr-2770.preview.docs.apify.com/actors/publishing/monetize/pay-per-event

Closes #2634.

### DIFF
--- a/sources/platform/actors/publishing/monetize/pay_per_event.mdx
+++ b/sources/platform/actors/publishing/monetize/pay_per_event.mdx
@@ -129,6 +129,14 @@ The Apify platform enforces this limit. Once it's reached, `Actor.charge()` and 
 
 To protect your profits, finish the Actor run once charges reach the user's limit.
 
+:::tip Set a minimum for the user limit
+
+To prevent users from setting the maximum cost for the run too low, set the minimum value during the [monetization setup](/actors/publishing/monetize#set-up-monetization). For example, you can use it to cover the price of starting the Actor and producing one result.
+
+The minimum value is stored by the `minimalMaxTotalChargeUsd` property.
+
+:::
+
 For both custom and synthetic events, Apify SDKs return a `ChargeResult` from every `Actor.charge()` and `Actor.pushData()` call. The user spending limit is available in your Actor code as the `ACTOR_MAX_TOTAL_CHARGE_USD` [environment variable](/actors/development/programming-interface/environment-variables), and `ChargeResult` already accounts for it. Inspect its properties to decide when to stop the run:
 
 | Property | Description |
@@ -139,14 +147,14 @@ For both custom and synthetic events, Apify SDKs return a `ChargeResult` from ev
 
 :::caution The limit is per run, not per user
 
-`ACTOR_MAX_TOTAL_CHARGE_USD` and the `ChargeResult` limit apply to a single run. If a user starts many runs at once, your aggregate platform costs can grow beyond what a single run's limit suggests. To protect your profit, [set memory limits](#memory-limits) and keep per-event platform costs low.
+`ACTOR_MAX_TOTAL_CHARGE_USD` and the `ChargeResult` limit apply to a single run, not to a user's total spending. If a user starts many runs at once, your combined platform costs across all runs can grow beyond what a single run's limit suggests. To protect your profit, [set memory limits](#memory-limits) and keep per-event platform costs low.
 
 :::
 
 Make sure to:
 
-- Charge for work as soon as it's done, not in a large batch at the end. If the run is stops early, unbilled work is lost.
-- Prefer charging one unit at a time over batching with `count`, so a partial charge can't leave produced results unpaid.
+- Charge for work as soon as it's done, not in a large batch at the end. If the run stops early, produced results don't get charged.
+- Charge for results after they're saved and available for users. Don't bill users for what they can't access, for example if a run gets aborted.
 - Finish the run once `eventChargeLimitReached` is `true`.
 
 :::tip Consider using `ChargingManager`
@@ -289,7 +297,7 @@ Apify recommends using the synthetic `apify-actor-start` event in all Actors. It
 
 The `apify-actor-start` synthetic event works as follows:
 
-- It's enabled by default for all new PPE Actors. For existing Actors, you can enable it in Apify Console.
+- It's available by default for all new PPE Actors. For existing Actors, you can enable it in Apify Console.
 - Apify automatically charges the event. Don't manually charge the event in your Actor code, as it will fail.
 - The default price of the event is $0.00005, which equals to $0.05 per 1,000 starts. To keep your Actor competitive, Apify recommends keeping the default price.
 - The number of events charged depends on the amount of memory allocated to the run. Up to and including 1 GB of RAM, the event is charged once. Then, it's charged once for each extra GB of memory. For example:
--- a/sources/platform/actors/publishing/monetize/pay_per_event.mdx
+++ b/sources/platform/actors/publishing/monetize/pay_per_event.mdx
@@ -125,7 +125,7 @@ You can also call the [Charge events in run](/api/v2/post-charge-run) API endpoi
 
 Your Actor users can set the maximum cost for the run in Apify Console or with the API. It helps users control their spending, as they won't be billed beyond the limit.
 
-The Apify platform enforces this limit. Once it's reached, `Actor.charge()` and `Actor.pushData()` stop charging and pushing data, but your Actor keeps running, consuming platform resources, and producing results that are no longer billed.
+The Apify platform enforces this limit. Once it's reached, `Actor.charge()` stop charging and `Actor.pushData()` stops pushing data over the limit. The platform then aborts the run automatically. However, the run keeps consuming platform resources for a short time before it stops. For details, see [Handle graceful shutdown](#handle-graceful-shutdown).
 
 To protect your profits, finish the Actor run once charges reach the user's limit.
 
@@ -137,7 +137,12 @@ The minimum value is stored by the `minimalMaxTotalChargeUsd` property.
 
 :::
 
-For both custom and synthetic events, Apify SDKs return a `ChargeResult` from every `Actor.charge()` and `Actor.pushData()` call. The user spending limit is available in your Actor code as the `ACTOR_MAX_TOTAL_CHARGE_USD` [environment variable](/actors/development/programming-interface/environment-variables), and `ChargeResult` already accounts for it. Inspect its properties to decide when to stop the run:
+The user spending limit is available in your Actor code as the `ACTOR_MAX_TOTAL_CHARGE_USD` [environment variable](/actors/development/programming-interface/environment-variables), and `ChargeResult` already accounts for it. Returning a `ChargeResult` depends on the event type:
+
+- For custom events, `Actor.charge()` and `Actor.pushData()` called with an event name always return a `ChargeResult`.
+- For synthetic events, pushing datasets items is charged automatically. In the Python SDK, `Actor.pushData()` returns a `ChargeResult`. In the JavaScript SDK, `Actor.pushData()` called without an event name returns `void`.
+
+Inspect the `ChargeResult` properties to decide when to stop the run:
 
 | Property | Description |
 | --- | --- |
@@ -226,7 +231,7 @@ When using [Crawlee](https://crawlee.dev/), use `crawler.autoscaledPool.abort()`
 
 ### Handle graceful shutdown
 
-When a user gracefully aborts a run, the platform sends the `aborting` event and force-stops the run 30 seconds later. Use this window to bill for work already done, so an abort doesn't cost you the revenue.
+When a platform or user gracefully aborts a run, the platform sends the `aborting` event and force-stops the run 30 seconds later. Use this window to bill for work already done, so an abort doesn't cost you the revenue.
 
 Apify SDKs already handle the graceful shutdown for you. They listen for `aborting` and call `Actor.exit()` automatically. Register your own `aborting` handler to:
 
--- a/sources/platform/actors/publishing/monetize/pay_per_event.mdx
+++ b/sources/platform/actors/publishing/monetize/pay_per_event.mdx
@@ -38,7 +38,7 @@ You define the events and can set different pricing for each of them.
 
 ### Synthetic events
 
-Sythetic events cover the most common billing patterns. They're predefined by Apify and charged automatically:
+Synthetic events cover the most common billing patterns. They're predefined by Apify and charged automatically:
 
 - [`apify-actor-start`](#synthetic-start-event) covers the costs spent during the first 5 seconds of starting the Actor.
 - [`apify-default-dataset-item`](#synthetic-default-dataset-item-event) charges for each item written to the run's default dataset.
@@ -71,6 +71,7 @@ Only revenue and cost for Apify customers on paid plans are taken into considera
 To make your Actor profitable, price your events so that the revenue covers the platform costs. For details, see [Compute costs for PPE Actors](/actors/publishing/monetize/pricing-and-costs#compute-costs).
 
 If your Actor produces negative profit, make sure to:
+
 - [Set memory limits](#memory-limits).
 - [Respect user spending limits](#respect-user-spending-limits).
 - Enable the [`apify-actor-start` synthetic event](#synthetic-start-event).
@@ -131,7 +132,7 @@ To protect your profits, finish the Actor run once charges reach the user's limi
 For both custom and synthetic events, Apify SDKs return a `ChargeResult` from every `Actor.charge()` and `Actor.pushData()` call. The user spending limit is available in your Actor code as the `ACTOR_MAX_TOTAL_CHARGE_USD` [environment variable](/actors/development/programming-interface/environment-variables), and `ChargeResult` already accounts for it. Inspect its properties to decide when to stop the run:
 
 | Property | Description |
-|----------|-------------|
+| --- | --- |
 | `eventChargeLimitReached` | Checks if the user's limit allows for another charge of the event. |
 | `chargeableWithinLimit` | Indicates how many more of each event you can still charge within the remaining budget. Use it if you have multiple events. |
 | `chargedCount` | Indicates how many events were billed by the call. For batch charges (`count`), it can be lower than you requested. Charges above `chargedCount` are not billed. |
--- a/sources/platform/actors/publishing/monetize/pay_per_event.mdx
+++ b/sources/platform/actors/publishing/monetize/pay_per_event.mdx
@@ -1,31 +1,24 @@
 ---
-title: Pay per event
+title: Pay per event pricing model
 description: Learn how to monetize your Actor with pay-per-event (PPE) pricing, charging users for specific actions like Actor starts, dataset items, or API calls.
 slug: /actors/publishing/monetize/pay-per-event
 sidebar_position: 1.1
+sidebar_label: Pay per event
 ---
 
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-The pay per event (PPE) pricing model offers a flexible monetization option for Actors on Apify Store:
 
-- You can charge users based on specific events triggered programmatically by your Actor's code. Common events include starting an Actor, creating a dataset item, or calling an external API.
-- You can define different pricing for individual events.
-- You can charge for specific events directly from your Actor using the [JS](/sdk/js/reference/class/Actor#charge) or [Python](/sdk/python/reference/class/Actor#charge) SDK, or by calling the [PPE charging API](/api/v2/post-charge-run) directly.
+The pay per event (PPE) pricing model offers a flexible monetization option for Actors on Apify Store. The key benefits are:
 
-See the key benefits of the PPE pricing model:
-
-| Category | Pay per event (PPE) |
-| --- | --- |
-| Revenue scalability | Unlimited, scales with usage |
-| AI/MCP compatibility | ✅ Fully compatible |
-| User cost predictability | Predictable |
-| Store discounts | ✅ Store discounts available |
-| Marketing boost | Priority store placement |
-| Commission opportunities | Standard 20% |
-| Custom event billing | ✅ Charge for any event |
-| Per-result billing | Optional (via event; automatic via `apify-default-dataset-item`) |
+- Unlimited revenue scalability
+- AI/MCP compatibility
+- Predictable user costs
+- Store discounts
+- Standard 20% commission
+- Custom event billing
+- Per-result billing (optional)
 
 :::tip Additional benefits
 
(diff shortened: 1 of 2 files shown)


lisvrijsen (commented):
Looks and reads really good. Huge improvement!

What do you think about explicitly stating that the developer should test their Actor with an invalid input? Just charging for an invalid input doesn't necessarily cover the entire platform cost. If the Actor takes 10 min to handle the invalid input, and it charges for 1 event only, then the cost is still a lot higher.

novotnyj (approved):

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
-| Store discounts | ✅ Store discounts available |
-| Marketing boost | Priority store placement |
-| Commission opportunities | Standard 20% |
-| Custom event billing | ✅ Charge for any event |
-| Per-result billing | Optional (via event; automatic via `apify-default-dataset-item`) |
+- Unlimited revenue scalability
+- AI/MCP compatibility
+- Predictable user costs
+- Store discounts
+- Standard 20% commission
```
stetizu1: Not sure if this is actually the benefit, but I see it was presented like that 😅

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
+- Custom event billing
+- Per-result billing (optional)
```
stetizu1: I would skip this / maybe replace with something like `2 auto-charged events available (optional)` - this was probably used as a comparison with PPR and now it does not mean much standalone

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
-| User cost predictability | Predictable |
-| Store discounts | ✅ Store discounts available |
-| Marketing boost | Priority store placement |
-| Commission opportunities | Standard 20% |
-| Custom event billing | ✅ Charge for any event |
-| Per-result billing | Optional (via event; automatic via `apify-default-dataset-item`) |
+- Unlimited revenue scalability
+- AI/MCP compatibility
+- Predictable user costs
+- Store discounts
```
stetizu1: Maybe `Store discounts available`? They are still optional. but it is more important to communicate why are store discounts a benefit

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
 
 :::
 
-## How is profit computed
+## About events
+
+An event is a specific, measurable action that your Actor triggers programmatically and charges the user for. The most common events include:
+
+- Starting an Actor.
+- Creating a dataset item.
```
stetizu1: Both events above are actually default synthetic events - so maybe chose different exaples also?, maybe `add result to key-value store`, `scrape additional data` - you may search store for inspiration, but I would not use the default events.

I personally drop both and add something like
- Create one result
if you want them semi-represented here (but no tso specific

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
+- Calling an external API.
+
+You define the events and can set different pricing for each of them.
```
stetizu1: `set different pricing for each of them` is a bit obvious part - but user can also set different prices for different `tiers` - user plans. Maybe this is the place to briefly mention it in the sentence?

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
+
+### Synthetic events
+
+Synthetic events cover the most common billing patterns. They're predefined by Apify and charged automatically:
```
stetizu1: `most common` does not align that much with the move to standby and non-scraping use-cases. I would more focus on - predefined events that do not require any code for charging / charge automatically.

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
+
+- [`apify-actor-start`](#synthetic-start-event) covers the costs spent during the first 5 seconds of starting the Actor.
+- [`apify-default-dataset-item`](#synthetic-default-dataset-item-event) charges for each item written to the run's default dataset.
+
+You can adjust or remove synthetic events in Apify Console.
```
stetizu1: `adjust` is very flexible here, I would say 'the thought of'
These events are optional and if you use them, you can set their prices

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
+- Creating a dataset item.
+- Calling an external API.
+
+You define the events and can set different pricing for each of them.
+
+### Synthetic events
+
+Synthetic events cover the most common billing patterns. They're predefined by Apify and charged automatically:
+
+- [`apify-actor-start`](#synthetic-start-event) covers the costs spent during the first 5 seconds of starting the Actor.
```
stetizu1: I am missing info on when it is charged. Covering cost is nice, but the main point is that 1 `apify-actor-start` is charged per 1 GB of RAM used to run the Actor. The 5second cover is actually 'tiny' and just a very small info/advantage of using this event

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
+- [`apify-default-dataset-item`](#synthetic-default-dataset-item-event) charges for each item written to the run's default dataset.
```
stetizu1: I would maybe stress that it is only default dataset in some () or so - devs get confused here sometimes. You need custom events for different datasets

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
+
+You can adjust or remove synthetic events in Apify Console.
+
+### Custom events
+
+To charge users for the unique value that your Actor delivers, define custom events. You name each event, describe it, set its price, and charge it from your Actor's code.
```
stetizu1, line 43: Not sure where to mention tiered pricing, but I would hide it in all these sentences with more explanation below

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
 
 :::note Negative profit isolation
 
-An Actor's negative net profit does not affect the positive profit of another Actor. For aggregation purposes, any Actor with a negative net profit is considered to have a profit of $0.
+An Actor's negative net profit doesn't affect the positive profit of another Actor. For aggregation purposes, an Actor with a negative net profit is considered to have a profit of $0.
 
-- _Previously:_ `Total Profit = (-$90) + $100 = $10`
-- _Now:_ `Total Profit = $0 + $100 = $100`
+- Before: `Total profit = (-$90) + $100 = $10`
+- Now: `Total profit = $0 + $100 = $100`
```
stetizu1: I would be scared of "before" here - before when? What information this gives in past one year?

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -81,16 +113,45 @@ To pass platform usage costs on to users, in the **Actor pricing** step of the [
...
 
-The `eventChargeLimitReached` property checks if the user's limit allows for another charge of this event. If you have multiple events, analyze the `chargeableWithinLimit` property to see if other events can still be charged before stopping the Actor.
+### Respect user spending limits
 
-:::info ACTOR_MAX_TOTAL_CHARGE_USD environment variable
+Your Actor users can set the maximum cost for the run in Apify Console or with the API. It helps users control their spending, as they won't be billed beyond the limit.
 
-All pricing models support a spending limit set through the Apify Console or API. This limit is available in your Actor code as the `ACTOR_MAX_TOTAL_CHARGE_USD` [environment variable](/actors/development/programming-interface/environment-variables), which contains the user's maximum total cost for the run.
-The Apify SDK's `ChargeResult` respects the user set limit already.
+The Apify platform enforces this limit. Once it's reached, `Actor.charge()` and `Actor.pushData()` stop charging and pushing data, but your Actor keeps running, consuming platform resources, and producing results that are no longer billed.
```
stetizu1: It does stop charging, but I think it won't stop pushing data. Also Actor will be "gracefully aborted", it does not continue to run indefinitely, it just take some time when it is aborting gracefully (it has time to not loose data) - so it is still better handled when user handle it, not platform. But it creates just a slight excess.

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -81,16 +113,45 @@ To pass platform usage costs on to users, in the **Actor pricing** step of the [
...
+
+To protect your profits, finish the Actor run once charges reach the user's limit.
+
```
stetizu1, line 124: I would also mention here that you can set "minimum " for this value - minimal maxCostPerRun that will set the minimal price where run results in any meaningful output. It can be for example set to "1 start + 1 result".

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -81,16 +113,45 @@ To pass platform usage costs on to users, in the **Actor pricing** step of the [
...
+For both custom and synthetic events, Apify SDKs return a `ChargeResult` from every `Actor.charge()` and `Actor.pushData()` call. The user spending limit is available in your Actor code as the `ACTOR_MAX_TOTAL_CHARGE_USD` [environment variable](/actors/development/programming-interface/environment-variables), and `ChargeResult` already accounts for it. Inspect its properties to decide when to stop the run:
```
stetizu1: I am not sure Apify SDK Actor.pushData() returns anything. Please double-check with tooling on this.

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -81,16 +113,45 @@ To pass platform usage costs on to users, in the **Actor pricing** step of the [
...
+
+| Property | Description |
+| --- | --- |
+| `eventChargeLimitReached` | Checks if the user's limit allows for another charge of the event. |
+| `chargeableWithinLimit` | Indicates how many more of each event you can still charge within the remaining budget. Use it if you have multiple events. |
+| `chargedCount` | Indicates how many events were billed by the call. For batch charges (`count`), it can be lower than you requested. Charges above `chargedCount` are not billed. |
+
+:::caution The limit is per run, not per user
+
+`ACTOR_MAX_TOTAL_CHARGE_USD` and the `ChargeResult` limit apply to a single run. If a user starts many runs at once, your aggregate platform costs can grow beyond what a single run's limit suggests. To protect your profit, [set memory limits](#memory-limits) and keep per-event platform costs low.
```
stetizu1: I am not even sure I fully understand this 😅

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -81,16 +113,45 @@ To pass platform usage costs on to users, in the **Actor pricing** step of the [
...
+
+:::
+
+Make sure to:
+
+- Charge for work as soon as it's done, not in a large batch at the end. If the run is stops early, unbilled work is lost.
```
stetizu1: > If the run is stops early, unbilled work is lost.

Probably missclick 😅 Also 'work is not lost' - charging is. Things are not charged

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -81,16 +113,45 @@ To pass platform usage costs on to users, in the **Actor pricing** step of the [
...
+- Prefer charging one unit at a time over batching with `count`, so a partial charge can't leave produced results unpaid.
```
stetizu1: This says a bit the same thing as previous one. What am I missing there is description that charge should always be after result is somehow saved for users so you never charge for something that can be "not given" to user, for example by ending run early. This could end up in disputes where we would return money to the customer. Just a good practice.

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -81,16 +113,45 @@ To pass platform usage costs on to users, in the **Actor pricing** step of the [
...
+- Finish the run once `eventChargeLimitReached` is `true`.
+
+:::tip Consider using `ChargingManager`
+
+If charging happens in multiple places across your code, or when you use a crawler where you don't directly control the main loop, you can use the `ChargingManager` to check the remaining budget periodically instead of inspecting every `ChargeResult`.
```
stetizu1: `crawler` - I would say with move to agentic, I would just say `Actor`? We should focus on our docs to not be so scrape-focused

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -151,42 +212,92 @@ async def main():
 
 :::note Crawlee integration and spending limits
 
-When using [Crawlee](https://crawlee.dev/), use `crawler.autoscaledPool.abort()` instead of `Actor.exit()` to gracefully finish the crawler and allow the rest of your code to process normally.
+When using [Crawlee](https://crawlee.dev/), use `crawler.autoscaledPool.abort()` instead of `Actor.exit()` to gracefully finish the crawler and let the rest of your code run normally.
 
 :::
 
-## Best practices for PPE Actors
+### Handle graceful shutdown
```
stetizu1, line 225: Exactly here - this is also called when charged limit is reached. And one of the reasons for "work first, charge later" I just mentioned

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -151,42 +212,92 @@ async def main():
...
+:::caution Non-graceful aborts are default behavior
+
+A non-graceful abort stops the run immediately, without emitting `aborting`, so your handler never runs. Don't rely on it and charge for work as soon as it's done. For details, see [Respect user spending limits](#respect-user-spending-limits).
 
 :::
 
-Starting an Actor takes time, and creates additional cost for the Actor creator, because the profit equals revenue minus platform costs.
+### Use `apify-actor-start` {#synthetic-start-event}
+
+The `apify-actor-start` synthetic event reduces your startup costs and makes your Actor pricing more competitive. When enabled, Apify covers the compute cost of the first 5 seconds of every run.
```
stetizu1, line 288: yep, info is probably enough here, maybe no need to mention above that we are waiving the first 5 secs

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -151,42 +212,92 @@ async def main():
...
 
-#### How the synthetic start event works
+The `apify-actor-start` synthetic event works as follows:
 
-- The Apify Actor start event is _automatically enabled_ for all new PPE Actors. For existing Actors, you can enable it in Apify Console.
-- Apify _automatically charges_ the event.
-  - You must _not_ manually charge for the synthetic start event (`apify-actor-start`) in your Actor code. If you attempt to charge this event yourself, the operation will fail.
-- The default price of the event is _$0.00005_, which equals _$0.05 per 1,000 starts_. We recommend keeping the default price to keep your Actors competitive.
-- The number of events charged _depends on the memory_ of the Actor run. Up to and including 1 GB of RAM, the event is charged once. Then it's charged once for each extra GB of memory. For example:
+- It's enabled by default for all new PPE Actors. For existing Actors, you can enable it in Apify Console.
```
stetizu1: I am not sure what "enabled by default" means here. I would say that you can see it when setting your actor info for the first time. If you already have pricing set without it, you can re-show it by filling `apify-actor-start` to the name field.

[future]: We will change this at some point to be always visible

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -198,33 +309,32 @@ For existing Actors, you can add this event manually:
...
 1. Complete the remaining event details: title, description, and price.
 
-#### Synthetic start event for Actors with start event
+#### Actors with a custom start event
 
-Your Actor might already have a start event defined, such as `actor-start` or another variant of the event name. In this case, you can choose whether to use the synthetic start event or keep the existing start event.
+Your Actor can't use both the synthetic and custom start events.
 
-If you want to use the synthetic start event, remove the existing start event from your Actor and add the synthetic start event in Apify Console in the **Publication** tab.
+To use the synthetic event, remove the existing custom event from your Actor in Apify Console.
```
stetizu1, line 322: Well, I guess, but there is no thing implemented to really stop you. But it is definitely bad practice

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -319,32 +433,30 @@ async def main():
...
-### Keep pricing simple with fewer events
-
-Try to limit the number of events. Fewer events make it easier for users to understand your pricing and predict their costs.
+### Keep pricing simple
 
-### Make events produce visible results {#actor-events}
+Limit the number of events. Fewer events make it easier for users to understand your pricing and predict the costs.
 
-For Actors that produce data, events should map to something concrete in the user's dataset or storage.
+Avoid cases where producing a single result triggers too many events. It makes your Actor more expensive.
```
stetizu1, line 446: Hmm, this goes against our google maps scraper 😅 Do we want to present this? It seems like common knowledge.

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -319,32 +433,30 @@ async def main():
...
+You can still implement events that don't produce tangible results, such as running AI workflows or processing external API calls. In such cases, inform users about actions performed by your Actor by displaying a status message or pushing a record to the dataset. It helps users understand what actions they were charged for.
 
-:::note Additional context
+The following examples show clear mapping of events to results:
 
-You can display a status message or push a record to the dataset to inform users about non-data actions performed by your Actor. This helps users understand what actions were charged for, even if those actions do not produce tangible output.
+- `post` event: Each charge adds one social media post to the dataset.
+- `profile` event: Each charge adds one user profile to the dataset.
+- `processed-image` event: Each charge adds one processed image to the dataset.
+- `ai-analysis` event: Each charge processes one document through an AI workflow. It doesn't produce an output, but processing is still valuable.
```
stetizu1: Well it produces the output - output is the analysis. We want to show everything to the user, but one line in the dataset can have different columns and charges based on columns - as per GMaps scraper.

stetizu1 (changes requested):
Sorry I did not say which comments are actually important and which are nits (the most important are the ones that I see something factual is confusing)
- and in some you do have the information that I was missing, it is just more down the document. In these cases, maybe add a link. Thanks.

sources/platform/actors/publishing/monetize/pay_per_event.mdx
szaganek: Since this just an intro, I made it more generic then, all details are in the linked section.

sources/platform/actors/publishing/monetize/pay_per_event.mdx
szaganek: It's there :) But I changed to order, hopefully it's more visible now.

sources/platform/actors/publishing/monetize/pay_per_event.mdx
szaganek: Yeah, so we have this info on another page: https://docs.apify.com/actors/publishing/monetize/pricing-and-costs

Let me tackle it as a separate PR (https://github.com/apify/apify-docs/issues/2776). I'll reorganize and combine the two better.

sources/platform/actors/publishing/monetize/pay_per_event.mdx
szaganek: Hm, I based it on this: https://docs.apify.com/sdk/js/docs/concepts/pay-per-event#handling-the-charge-limit 

@michael-apify did I misunderstand that it's up to dev to stop the run?

sources/platform/actors/publishing/monetize/pay_per_event.mdx
szaganek: Don't underestimate people 🥲

sources/platform/actors/publishing/monetize/pay_per_event.mdx
szaganek: I rephrased it 🤔

sources/platform/actors/publishing/monetize/pay_per_event.mdx
szaganek: ok, I think I got it

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
+- Processing an item.
+- Calling an external API.
+
+You define the events and can set different pricing for each event type and [discount tier](/actors/publishing/monetize/pricing-and-costs#discount-tiers).
+
+### Synthetic events
+
+Synthetic events are predefined by Apify and charged automatically:
+
+- [`apify-actor-start`](#synthetic-start-event) reduces the costs of starting the Actor.
```
stetizu1, line 36: What I meant originally was that I would also expect info in this one sentence what it actually charges for (gb of memory) - to be consistent with result  - but if you think this makes more sense, keep it.

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
+You can remove synthetic events or change their price in Apify Console.
+
+### Custom events
+
+To charge users for the unique value that your Actor delivers, define custom events. You name each event, describe it, set its price, and charge it from your Actor's code.
 
-Your profit is calculated from the mentioned formula:
+### Get event names
+
+To retrieve the list of all chargeable event names for your Actor, use the [Get Actor](/api/v2/act-get) API endpoint.
```
stetizu1, line 47: I didn't even know that 😅

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -151,42 +225,92 @@ async def main():
...
+    await Actor.exit()
+```
+
+</TabItem>
+</Tabs>
+
+
+:::caution Non-graceful aborts are default behavior
+
+A non-graceful abort stops the run immediately, without emitting `aborting`, so your handler never runs. Don't rely on it and charge for work as soon as it's done. For details, see [Respect user spending limits](#respect-user-spending-limits).
```
stetizu1, line 282: Does user-spending-limits have anything in common with non-graceful aborts? They are done manualy

stetizu1 (approved):
👍

sources/platform/actors/publishing/monetize/pay_per_event.mdx
TC-MO: I think it's a remnant of when we had more models, and they had different commission. Might be okay to remove at this point?

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
+- Store discounts available
+- Standard 20% commission
+- Custom event billing
+- Two automatically charged events available
 
 :::tip Additional benefits
 
 Actors that implement PPE pricing receive additional benefits, including increased visibility in Apify Store and enhanced discoverability.
 
 :::
```
TC-MO: Same might be true here. Since PPR is getting deprecated is it still worth mentioning additional benefits of PPE?

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -1,61 +1,93 @@
...
+- Without isolation: `Total profit = (-$90) + $100 = $10`
+- With isolation: `Total profit = $0 + $100 = $100`
 
 :::
 
-## How to set pricing for PPE Actors
+## Set pricing for PPE Actors
 
 1. _Understand your costs_: Analyze resource usage (e.g CPU, memory, proxies, external APIs) and identify cost drivers
 1. _Define clear events_: break your Actor's functionality into measurable, chargeable events.
```
TC-MO: Fix opportunity, unify whether lists ends with periods or not.
```suggestion
1. _Understand your costs_: Analyze resource usage (e.g CPU, memory, proxies, external APIs) and identify cost drivers.
1. _Define clear events_: break your Actor's functionality into measurable, chargeable events.
```

sources/platform/actors/publishing/monetize/pay_per_event.mdx
```diff
@@ -319,32 +446,30 @@ async def main():
...
+You can still implement events that don't produce tangible results, such as running AI workflows or processing external API calls. In such cases, inform users about actions performed by your Actor by displaying a status message or pushing a record to the dataset. It helps users understand what actions they were charged for.
 
-:::note Additional context
+The following examples show clear mapping of events to results:
 
-You can display a status message or push a record to the dataset to inform users about non-data actions performed by your Actor. This helps users understand what actions were charged for, even if those actions do not produce tangible output.
+- `post` event: Each charge adds one social media post to the dataset.
+- `profile` event: Each charge adds one user profile to the dataset.
+- `processed-image` event: Each charge adds one processed image to the dataset.
+- `ai-analysis` event: Each charge processes one document through an AI workflow.
```
TC-MO, line 459: perhaps a table instead of bullet list?

TC-MO (approved):
Few thoughts and visual nits, otherwise LGTM

sources/platform/actors/publishing/monetize/pay_per_event.mdx
szaganek: It's linked since that section focuses on charging work as soon as it's done 🤔 Well, it's definitely not the last rewrite for this article.

PR #2774, opened 2026-07-16
fix(openapi): Preserve RequestWithoutId and its required fields when bundling

vdusek:
`RequestWithoutId` was a bare `$ref` with sibling keys, so the bundler inlined it: the published `openapi.json` has no such component, the add-request / batch-add-requests / update-request bodies point straight at `RequestBase`, and `required: [uniqueKey, url]` disappears. Generated clients therefore get no schema for those bodies — `apify-client-python` fell back to a response shape and sent `userData` and friends snake_cased, which the API rejects with HTTP 400 (apify/apify-client-python#1026).

Wrapping `RequestWithoutId` in a single-member `allOf` keeps it a named component through bundling. `required: [uniqueKey, url]` moves to `RequestBase`, and `Request` drops the now-redundant entries — net constraints unchanged.

`redocly lint` passes, and the bundle keeps the component with all 11 properties and the required constraint.

*✍️ Drafted by Claude Code*

### DIFF
--- a/apify-api/openapi/components/schemas/request-queues/RequestWithoutId.yaml
+++ b/apify-api/openapi/components/schemas/request-queues/RequestWithoutId.yaml
@@ -1,6 +1,4 @@
 title: RequestWithoutId
 description: A request stored in the request queue, including its metadata and processing state, without the assigned ID.
-$ref: ./RequestBase.yaml
-required:
-  - uniqueKey
-  - url
+allOf:
+  - $ref: ./RequestBase.yaml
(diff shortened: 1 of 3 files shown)


fnesveda (approved):

apify-api/openapi/components/schemas/request-queues/RequestWithoutId.yaml
```diff
@@ -1,6 +1,7 @@
 title: RequestWithoutId
 description: A request stored in the request queue, including its metadata and processing state, without the assigned ID.
-$ref: ./RequestBase.yaml
-required:
-  - uniqueKey
-  - url
+# Wrapped in a single-member `allOf` so that bundling keeps this component. A bare `$ref` with sibling keys is a
```
Pijukatel, line 3: Maybe the comment can be just in the description of this PR

Pijukatel (approved):

PR #2778, opened 2026-07-17
fix: Remove unnecessary whitespace on mobile

szaganek:
On mobile, the second navbar still shows up. It's empty, because all links moved into the dropdown. That creates unnecessary whitespace.

Before:
Screenshot 2026-07-17 at 13 40 04

After:
Screenshot 2026-07-17 at 13 40 11

### DIFF
--- a/apify-docs-theme/src/theme/Navbar/Content/index.jsx
+++ b/apify-docs-theme/src/theme/Navbar/Content/index.jsx
@@ -23,9 +23,9 @@ function NavbarItems({ items }) {
     );
 }
 
-function NavbarContentLayout({ left, center, right }) {
+function NavbarContentLayout({ left, center, right, className }) {
     return (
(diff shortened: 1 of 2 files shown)


apify-docs-theme/src/theme/Navbar/Content/index.jsx
```diff
@@ -115,6 +115,7 @@ export default function NavbarContent() {
                 }
             />
             <NavbarContentLayout
+                className="navbar__inner--mobile"
```
barjin: I wouldn't call this `--mobile` - this component is rendered everywhere. Maybe something like `subnavbar` (with the same CSS rule) would work better?

barjin (approved):
Lgtm, just one naming nit. 

Thank you @szaganek !

PR #2779, opened 2026-07-17
fix: Content on mobile rendered too narrow

szaganek:
Content from apify-docs was rendered with extra padding compared to other repos. Extra margins now apply to desktop only.

Before:
Screenshot 2026-07-17 at 14 50 18

After:
Screenshot 2026-07-17 at 14 50 14


barjin (approved):
lgtm, thank you @szaganek !

PR #2788, opened 2026-07-20
docs: Update procedure on passing platform usage costs on to users

szaganek:
We are moving PPE+U under **Additional options**.


stetizu1 (approved):
It is ok, though I don't like the heading of the section too much, sounds like a command :D "Pass it!"

PR #2791, opened 2026-07-21
fix(docs): Duplicated routes in Academy

szaganek:
Due to the recent navigation changes, slugs got duplicated and some content disappeared from Academy.


TC-MO (approved):
LGTM

PR #2795, opened 2026-07-22
docs: Remove add-actor mentions from MCP integration docs

MQ37:
add-actor is no longer selectable for new MCP connections (call-actor is substituted instead, apify/apify-mcp-server#1127). Removes its table row and the now-inaccurate dynamic-tool-discovery note.


MQ37:
FYI @jirispilka

marcel-rbro (approved):
Fixed small not, looks good otherwise 👍

PR #2798, opened 2026-07-23
fix: removed contact card from Hubspot setup instructions

JanHranicky:
- As of now the card only operates on companies, removed Contact reference that were not true


marcel-rbro (approved):
Fixed minor styling issue, LGTM 👍

PR #2819, opened 2026-07-27
docs: Add experimental API clients section to API page

jancurn:
Adds a new "Experimental API clients" section to the API documentation page showcasing official Apify client libraries for Rust, Go, PHP, Java, and .NET.

**Key changes:**

- Added `ExperimentalClient` interface and `experimentalClients` array with complete metadata (language, description, repository, installation, and code examples) for 5 experimental client libraries
- Created new "Experimental API clients" section on the API page using tabbed interface to display each client with installation instructions and usage examples
- Updated navigation configuration to simplify the "APIs, SDKs & CLI" dropdown into separate "API", "SDK", and "CLI" menu items
- Added Rust syntax highlighting support to Docusaurus code block configuration

The experimental clients section includes links to GitHub repositories, "Get started" buttons, and reference documentation links for each language.

https://claude.ai/code/session_016as7phod6GN5z2g7x6ng1V


TC-MO (approved):
LGTM, few non-blocking thoughts

I think it would be better to preserve dropdowns. Clicking on API/SDK/CLI would still put you on the general landing page, but we would preserve a one-click way to get to API Reference & specific languages. Until we redesign docs these dropdowns can act as language switchers for docs.

There seems to be inconsistency with how we apply line numbers to codeblocks on the landing page. Python & JS has them while any other codeblock does not.

B4nan:
If we now have separate items in the navbar for api/sdk/cli, I would like to see those as dropdowns with the links to the different language variants, personally.

Also, I'd change the order of the languages in the experimental clients section, I feel like rust should go to the end of the list. Plus the GH stars buttons are not working, they don't show the numbers (and even if they would, we have little to no stars on those repositories).

jancurn:
Thanks @B4nan - all three points addressed.

**Dropdowns:** done in 78f0428 / 86cfab8, pushed after your comment. API, SDK and CLI are dropdowns again, opening on hover as before, and their labels are now real links too, so clicking `API` opens `/api` instead of only toggling the menu. `MCP` was added next to them as a plain link.

```
API   -> /api    API reference, Client for JavaScript / Python / Rust / Go / PHP / Java / .NET
SDK   -> /sdk    SDK for JavaScript, SDK for Python
CLI   -> /cli    Installation, Quick start, Command reference
MCP   -> /integrations/mcp
```

**Language order:** Rust moved to the end, so the tabs read Go, PHP, Java, .NET, Rust.

**Star buttons:** removed from the experimental clients - agreed they earn nothing there. Each tab now has just Get started (repo) and View reference (`docs/README.md`). I left the buttons on the JavaScript and Python tabs alone since they predate this PR; say the word if you want those gone too. I could not reproduce the missing-count part from my environment (api.github.com is blocked there), so if the count is also missing for `apify-client-js` on the preview, that is a separate `GitButton` bug worth its own issue.

One note on CI: `Check PR title` is red because of the `[DRAFT]` prefix in the title, which breaks the conventional-commits format. Left as is on purpose - it goes green once the prefix comes off.

---
_Generated by [Claude Code](https://claude.ai/code)_

jancurn:
@B4nan this should be good to go (human here)

B4nan (approved):
thanks, fine by me

B4nan:
maybe one small nit, rust is still on the first position in the navbar dropdown

image

B4nan:
and another visual nit, when the viewport is around 1000px, the navbar items collide

image

jancurn:
@TC-MO both of your notes are in.

**Dropdowns:** restored in 78f0428 / 86cfab8, which landed after your review. They work as you described - hover opens the menu, clicking `API` / `SDK` / `CLI` goes to the landing page, and the menu is a one-click path to the reference and each language.

**Line numbers:** the theme numbers code lines through a CSS counter with a hardcoded language allowlist (`ts`, `typescript`, `javascript`, `json`, `json5`, `py`, `python`, `dockerfile`, `XML`, `yaml`, `js`) in `apify-docs-theme/src/theme/custom.css`, which is why only the JavaScript and Python blocks were numbered - the cURL block was already unnumbered before this PR. Turned them off for every block on this page in 8f59f64 rather than extending the allowlist, since that rule ships to all six docs sites and the one-line install commands would look odd numbered. Docs pages keep their line numbers.

Unrelated find while in there: the allowlist entry is `.language-XML`, but Docusaurus emits `language-xml`, so that one has never matched. Worth a separate fix if XML blocks are meant to be numbered.

---
_Generated by [Claude Code](https://claude.ai/code)_

jancurn:
@B4nan both nits are fixed in 851785a.

**Rust in the dropdown:** moved to the end, so it matches the tab order on the page - API reference, JavaScript, Python, Go, PHP, Java, .NET, Rust.

**Collision around 1000px:** the two halves of the link row ran out of room between 997px and about 1010px. I did not move the mobile breakpoint, because Docusaurus pins it to 996px in both Infima and `useWindowSize`, and `useNavbarMobileSidebar` calls that hook without an override - revealing the hamburger above 996px through CSS alone gives a button that does nothing, since the drawer never renders and the context force-closes it on desktop (checked at 1050px: no `.navbar-sidebar` in the DOM at all). So the row now fits down to 996px instead: the gap between the right-hand links tightens below 1100px, which turns an 11px overlap at 997px into 37px of clearance. Nothing changes at 1101px and up, and the hamburger still works at 996px and below.

---
_Generated by [Claude Code](https://claude.ai/code)_

jancurn:
Should be good now, merging

PR #2826, opened 2026-07-28
docs: enrich Actor development pages from help-center articles

marcel-rbro:
Salvages still-useful content from four retiring help-center articles into the platform docs, so the information survives the help-center wind-down and lives in one place.

- **Continuous integration**: replaces the outbound "Bitbucket CI guide" link with an in-page Bitbucket Pipelines section (modernized YAML on `node:22`, secured `APIFY_TOKEN`, tests before push, branch-based build tags).
- **Usage and resources**: adds reference compute-unit rates and a test-run method for estimating consumption, and repoints the two former help-center links in-page.
- **Builds**: adds a versioning best-practices subsection for public Actors.
- **Monetize**: adds a section on compensating users for failed runs with platform credits.

Part of the help-center audit; a companion PR covers the Integrations articles.


sources/platform/actors/publishing/monetize/index.mdx
```diff
@@ -106,6 +106,24 @@ When monetizing your Actor, you might want to limit features or usage for users
 - Graceful exits: If a free user hits a limit, don't crash the Actor or return a system error. Instead, exit gracefully with a clear [status message](/actors/development/programming-interface/status-messages#communicating-limitations) explaining the limit (e.g., "Free tier limit reached").
 - Avoid confusion: Never make a policy restriction look like a bug or platform error.
 
+## Compensate users for failed runs
+
+If a run fails or returns bad results because of a problem with your Actor, you can compensate the affected user with Apify platform credits. The amount is deducted from your payout.
```
tomnosek, line 111: I think the content team said that using "credits" is a no-no, and that we should use "prepaid usage". Verify.

sources/platform/actors/publishing/monetize/index.mdx
```diff
@@ -106,6 +106,24 @@ When monetizing your Actor, you might want to limit features or usage for users
...
+
+1. The user reports the problem on your Actor's **Issues** tab.
+1. Review the issue and decide whether it's valid. Ask the user for the run details, such as the run ID or a shared run link, so you can verify what happened before you approve anything.
+1. If the compensation is warranted, open the Apify support chat, select **Request user credit compensation**, and fill in the requested details: the user ID, the Actor ID, the amount, and the reason (or a link to the issue). Apify then credits the user's account.
+
+<img src={require('./images/request-user-credit-compensation.webp').default} alt="The Request user credit compensation option in the Apify support chat" width="320" />
+
+The **Request user credit compensation** option appears in the Apify support chat only for Actor developers. If you're contacting support as a user, you won't see it.
```
tomnosek, line 119: > If you're contacting support as a user
This feels a bit awkward - aren't they all users? How do you distinguish them?
Technically, it works based on whether they currently have any public Actor in Store.

sources/platform/actors/publishing/monetize/index.mdx
```diff
@@ -106,6 +106,24 @@ When monetizing your Actor, you might want to limit features or usage for users
...
+
+Keep a few things in mind when handling compensations:
```
tomnosek, line 121: I do think that this section is quite simplified and the "How to talk about compensation" section in Help Center is more descriptive for the publisher. Since they'll be scanning this info, chances are that they'll skip these three bullet points altogether. Since we've had quite a lot of problems with this, it deserves much bigger space and attention.

tomnosek:
I've left a couple of comments (not a proper review though, just a random stumble), but my overall impression is that it expands a subpar page, creating even more subpar page 😄 The page is sort of like an incoherent reference jumping from topic to topic.

A couple of points:
- It's in the Actors > Running Actors section, so the target audience is also common users. Yet it starts by saying Actors run in Docker containers. How is that important to this topic?
- After the leading paragraph, we get the heading "Resources"... in an article already named "Usage and resources". I don't even know what you mean by it, and I've already read it twice as headings in 5 rows.
- It says about the memory that "It must be a power of 2." but it was recently changed to a drop-down menu, so it's not an important info anymore (and for dynamic memory, it's written in the dynamic memory page).
- In the middle, there's a section about Crawlee... If I'm a user, I'm like "What the F is Crawlee?" There could be a callout, but it's quite irrelevant for the broad audience, so it shouldn't be part of the main content.
- The "What is a compute unit" is in the middle, yet half of the page above talks about "CUs" and often doesn't even reference this part.
- The "What determines consumption" section suddenly talks about Playwright and Cheerio. Again, as a user, I'm completely confused.

This page jumps from topic to topic, talks about terms that I don't understand as a reader, and in the end, I feel like I haven't learned anything. With a certain level of skimming, I certainly can't configure the Run options part properly. So, I think the page should be just completely rewritten.

(Btw. I know that this commit is not adding the whole page and that it was created in the past likely in many iterations, but let's end up with a better page, not just mixing more confusion in this confusion pot.)

TC-MO:
Hey there, I also took a look, and building on what Tom said, I would say the best course of action would be to close this PR for a few different reasons.

1. The screenshot added goes  against our [guidance](https://github.com/apify/apify-docs/blob/master/CONTRIBUTING.md#formatting-conventions), wrong highlight color, wrong style. It's better not to add screenshot and make sure the prose stands on its own than to add poor quality screenshots, especially since we are already running a clean-up of such screenshots from docs. Also it adds back the old logo to docs which is less than ideal.
2. It predates our reconciliation. And thus is assigned to someone else.
3. As Tom mentioned, it takes sub-par pages and makes them even more sub-par (sub-sub-par if you pardon my poor joke :D)  which was not the task at hand. We decided that in addition of porting the content we'll take a look at the target pages and uplift them.

PR #2834, opened 2026-07-29
docs: rename APIFY_MCP_PROXY_URL to ACTOR_MCP_CONNECTOR_BASE_URL

valekjo:
The MCP connector base URL environment variable is being renamed on the platform, so the docs need to match - otherwise anyone following the MCP connectors guide reads a variable name that no longer resolves.

Since the reference tables in `environment_variables.md` are grouped by name prefix, the variable also moves from the Apify platform (`APIFY_`) table to the standard Actor (`ACTOR_`) table.

TODO: Release this once worker release is out.


TC-MO (approved):
LGTM

marcel-rbro (approved):

PR #2835, opened 2026-07-29
ci: Drop cross-repo model regeneration dispatch to apify-client-python

vdusek:
Removes the cross-repo dispatch that opened a companion model-regeneration PR in `apify-client-python` for every spec-changing docs PR, and rewrites the `AGENTS.md` section describing it.

`apify-client-python` now pulls the spec on its own nightly schedule instead, so nothing on this side needs to push to it or be merged in lockstep with a client PR.

Why the dispatch goes away:

- The client can only generate models from the **published** spec, so it can never be genuinely ahead of a deployed docs change - the companion PR was opened against an unpublished bundle it would have to regenerate from anyway.
- Two mechanisms writing the same generated files made them diverge. The dispatch reused one long-lived branch per docs PR and regenerated with the codegen tooling from that stale branch, which silently reverted fixes that had landed on the client's master in the meantime.
- Every such PR opened with a placeholder `TODO` title that a human had to replace before it could be merged, so unmerged PRs and their stale branches piled up.

Also drops what only the removed jobs needed: the `closed` pull request trigger, the `if:` guard skipping lint on close events, and the `pull-requests: read` permission.

**Merge this before** apify/apify-client-python#983 - that PR deletes `manual_regenerate_models.yaml`, which the dispatch removed here still calls.

*✍️ Drafted by Claude Code*

### DIFF
--- a/.github/workflows/openapi-ci.yaml
+++ b/.github/workflows/openapi-ci.yaml
@@ -2,21 +2,16 @@ name: OpenAPI checks
 
 on:
     pull_request:
-        # The 'closed' type is needed to trigger cleanup of the corresponding apify-client-python PR.
-        types: [opened, synchronize, reopened, closed]
     push:
         branches:
             - master
 
 permissions:
     contents: read
-    pull-requests: read
 
 jobs:
     lint:
         name: Lint specification
-        # Skip lint/build/validate on PR close events - those only need the cleanup job.
-        if: github.event_name != 'pull_request' || github.event.action != 'closed'
         runs-on: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


TC-MO (approved):
LGTM, so no longer companions PR's and no longer thinking about stacking merges so everything generates correctly :D?

vdusek:
> LGTM, so no longer companions PR's and no longer thinking about stacking merges so everything generates correctly :D?

Yes, because of these and other issues, we rethought this. There will be only nightly syncs from now, let's see how that will work.

PR #2837, opened 2026-07-31
chore: bump the `@docusaurus/*` packages to 3.10.2

barjin:
Bumps all `@docusaurus/*` packages to 3.10.2. Needed three fixes to build: the renamed `future.faster` config key, a patch making the openapi theme's tabs use 3.10's context-based Tabs internals (backported from docusaurus-theme-openapi-docs 5.x, which we can't upgrade to yet - 13 of our 21 patch hunks don't apply to it), and bumping `@docusaurus/theme-common` in `apify-docs-theme` so it no longer resolves a second copy with its own React contexts. Verified with a full `pnpm build`.


TC-MO (approved):
LGTM

PR #2536, opened 2026-05-14
docs: misc academy docs fixes

artogahr:
Hi folks, I'm going through _some_ of the academy docs, I'll try to fix any inconsistencies I find here

PR #2838, opened 2026-07-31
fix(academy): avoid the Christmas tree

honzajavorek:
At this stage of the course, students are not yet introduced to data cleaning, so I'd rather not put .replace() in their way. Choosing a slightly different category on the IKEA website avoids the Christmas tree, which means the price is just a three-digit number, which can be scraped as-is.

This is a follow-up to the hotfix introduced in https://github.com/apify/apify-docs/pull/2536

_I never thought I'd find myself in a phase of life where I'm creating a git branch called "avoid christmas tree", but 2026 is exactly this eventful and here we are..._


szaganek (approved):
PR description on point, as always 👌

PR #2840, opened 2026-08-02
feat: add Actor task publication endpoints

Janjiran:
Part of https://github.com/apify/apify-core/issues/29471

Adding new public endpoint for Actor Task publication API.

---
Blocked by https://github.com/apify/apify-core/pull/29623


janbuchar (approved):
This looks legit as far as openapi semantics are considered

fnesveda (changes requested):
Is the API final? Why is it so different from the Actor publication API?
For example:
- Actors are published through the "update Actor" endpoint, tasks have special `/publish` and `/unpublish` endpoints
- Actors have `isPublic`, tasks have `publishedAt`
- Actors have a `categories` array, tasks have `categorization` string
- Actors have all the details (`seoTitle` etc.) on the top level, tasks have it in `publicConfig`

Can we unify the API so that it's the same as for Actors?

Also, property names like `inputSchemaFields`, `datasetName` and `datasetView` somehow don't capture that they're being used only when showing the task, but they sound like they're relevant to the whole task (e.g. that they limit what input fields can the task process). How about putting them under some `displayOptions` object?

Also CC @gippy - would you look at this through the lens of the rules for API endpoints you wrote? 🙏

gippy:
We already discussed this.

The 3 points:
* Actors have isPublic, tasks have publishedAt
* Actors have a categories array, tasks have categorization string
* Actors have all the details (seoTitle etc.) on the top level, tasks have it in publicConfig

Already exist in the current console backend API, so changing them would require migration and complete rewrite.

I already shared my opinion on this when it was going to be impelemented in console backend, but now that train has left the station.

So for API I proposed the option to have the `/publish` and `/unpublish` endpoints, because public tasks do not have `isPublic` field and sending `publishedAt: <some date>` / `publishedAt: null` to publish/unpublish is weird. Same sending `isPublic` field which is just virtual and does not really exist. It matches what the guidelines say and it's not that uncommon.

Since publishing does multiple verification steps, not just for the task detail but also for Actor permission and the Actor "readyness" for task publication it kinda makes sense.

Alternatives are either:
1. Have a single PUT Actor task endpoint which takes a different shape of data and then "transforms" it into real data in the database. Which is hacky
2. Completely redoing the public tasks schema to match Actors, but then the task would have many optional fields which 99% of tasks do not have, which is also is not great.

From my point of view I think it's ok to not have it match, the logic on Actor is not great either with the Actor update endpoint having 9 dependencies and over 400+ lines of code in the `updateActor` function alone. This way at least the logic is pretty simple. Update task endpoint just checks if the task is valid against schema and all logic related to publication/unpublication is separate. It also makes it much easier to document the errors the endpoints returns.

fnesveda:
> Already exist in the current console backend API, so changing them would require migration and complete rewrite.
>
> I already shared my opinion on this when it was going to be impelemented in console backend, but now that train has left the station.

I really don't agree with this. We can't shape the public API around our internal implementation, we need to make a good public API and then make the implementation work with it.

Writing a migration is a one time pain (and I believe this one can be done through Claude pretty well). A bad API stays with us forever.

> So for API I proposed the option to have the `/publish` and `/unpublish` endpoints

On their own, these are good, and I agree we should have done the Actors API this way, but _that_ train has surely left the station. So now that we have the Actors API done like this, let's try to have the tasks API consistent with it.

Janjiran:
> > Already exist in the current console backend API, so changing them would require migration and complete rewrite.
> 
> >
> 
> > I already shared my opinion on this when it was going to be impelemented in console backend, but now that train has left the station.
> 
> 
> 
> I really don't agree with this. We can't shape the public API around our internal implementation, we need to make a good public API and then make the implementation work with it.
> 
> 
> 
> Writing a migration is a one time pain (and I believe this one can be done through Claude pretty well). A bad API stays with us forever.
> 
> 
> 
> > So for API I proposed the option to have the `/publish` and `/unpublish` endpoints
> 
> 
> 
> On their own, these are good, and I agree we should have done the Actors API this way, but _that_ train has surely left the station. So now that we have the Actors API done like this, let's try to have the tasks API consistent with it.


Isn't that exactly what we're trying to do here though? To make a good public API and then make the implementation work with it, instead of shaping the new API around the Actor publication design, which you agree is the worse one. So we are designing the tasks API the way we think a good API should look.

I get the consistency point, but if a _bad API stays with us forever_, then requiring every new API to copy the old pattern is exactly how it stays forever. By following the latest API conventions @gippy put together, the `/publish` and `/unpublish` endpoints should be the correct way and later we can add the same endpoints to Actors. That's a non-breaking change and the `isPublic` flow can be deprecated over time. Copying the Actor design into the tasks API now closes that door.

The schema differences (`categories` vs. `categorization`, `publicConfig` etc) are a fair point but I'd separate it from the publish endpoints discussion.

fnesveda:
> Isn't that exactly what we're trying to do here though? To make a good public API and then make the implementation work with it, instead of shaping the new API around the Actor publication design, which you agree is the worse one. So we are designing the tasks API the way we think a good API should look.

I get your point, and in isolation, the `/publish` + `/unpublish` flow is better, but the `isPublic` flow isn't so bad that it would warrant breaking the consistency across endpoints. This is how I sort the options according to how nice they are:
1. `/publish` + `/unpublish` flow on both Actors and tasks
2. `isPublic` flow on both Actors and tasks
3. `/publish` + `/unpublish` on one, `isPublic` on the other

> I get the consistency point, but if a _bad API stays with us forever_, then requiring every new API to copy the old pattern is exactly how it stays forever. By following the latest API conventions @gippy put together, the `/publish` and `/unpublish` endpoints should be the correct way and later we can add the same endpoints to Actors. That's a non-breaking change and the `isPublic` flow can be deprecated over time. Copying the Actor design into the tasks API now closes that door.

Requiring every new API to copy the old pattern is how we prevent a bad API in the first place - an inconsistent API is a bad API.

If we decide "later we can add the same endpoints to Actors", realistically, we'll never do it, because we'll always have better things to do than changing an ancient, heavily integrated, slightly bad API flow. (I've had my technical purist heart broken many times by the reality of prioritization with limited capacity 🥲)

Janjiran:
Based on the discussion above, we've agreed to change the API to match the Actors:

- publication status is changed via virtual `isPublic` property that end-user can send with the PUT request. The `isPublic` is also returned from GET and LIST endpoints and dynamically computed via `publishedAt` field. Other tasks I've put in a ticket https://github.com/apify/apify-core/issues/29803 and will be discussing later

fnesveda:
I like the `isPublic` changes, thanks. I'll review this fully once the API is finalized. Until then, maybe it's worth converting this PR to draft so that we don't merge it accidentally?

apify-api/openapi/components/schemas/actor-tasks/Task.yaml
```diff
@@ -55,3 +55,14 @@ properties:
   standbyUrl:
     type: [string, "null"]
     format: uri
+  isPublic:
+    type: boolean
+    description: |
+      Whether the task is published on its public landing page. Derived from
+      `publicConfig.publishedAt`. Set it via the [Update task](/api/v2/actor-task-put)
+      endpoint to publish or unpublish the task.
```
katzino: Suggestion: This might be a bit misleading, no? The landing page will render if both `publishedAt` and `renderableAt` are set, right? `publishedAt` on its own allows task to be copied, not necessarily the landing page to be shown. WDYT?

apify-api/openapi/components/schemas/actor-tasks/TaskPublicConfig.yaml
```diff
@@ -0,0 +1,37 @@
...
+    type: [string, "null"]
+    description: Use-case category of the public task.
+  inputSchemaFields:
+    type: [array, "null"]
+    description: Names of the task input fields displayed on the public task page.
+    items:
+      type: string
+  datasetName:
+    type: [string, "null"]
+    description: Name of the dataset from the Actor's dataset schema whose results are displayed.
```
katzino: Nit: Could be beneficial to mention that load the default dataset if `null`

apify-api/openapi/components/schemas/actor-tasks/TaskPublicConfig.yaml
```diff
@@ -0,0 +1,37 @@
...
+  datasetView:
+    type: [string, "null"]
+    description: Key of the dataset view from the Actor's dataset schema used to display results.
```
katzino: Nit: Same here, we fallback to `overview` view if I remember correctly.

apify-api/openapi/components/schemas/actor-tasks/UpdateTaskRequest.yaml
```diff
@@ -18,3 +18,18 @@ properties:
...
+      cannot be cleared this way. Set `isPublic` to change the publication state.
+      Updating `publicConfig` requires write permission to the task's Actor.
+    $ref: ./TaskPublicConfig.yaml
+  isPublic:
+    type: boolean
+    description: |
+      Set to `true` to publish the task on its public landing page, or `false` to unpublish it.
+      Sending the value the task already has does nothing. Publishing requires the task's
+      `publicConfig` to be filled in and write permission to the task's Actor; it fails if the
+      task is not ready to be published, leaving the rest of the update unapplied.
```
katzino: Suggestion: Similar to my first comment, I would maybe state that `publishedAt` itself allows task to be copied while for the landing we run the validation (it shouldn't differ that often but there is still a chance the landing page stops working with new build while the task can still be copied).

katzino (approved):
LGTM ✅ 
Had some factual comments, feel free to skip those that doesn't make sense to you. Maybe someone from the content team could help us to formulate it properly.

apify-api/openapi/components/schemas/actor-tasks/CreateTaskRequest.yaml
```diff
@@ -23,3 +23,9 @@ properties:
     anyOf:
       - $ref: ../actors/ActorStandby.yaml
       - type: "null"
+  publicConfig:
+    description: |
```
szaganek, line 27: I'd go with something like this:
`Configuration that controls how the published task appears on its public landing page. Editing this object requires write permission to the Actor that the task belongs to.`

We don't need to include info on what not to do here, the publication process should be explained in the endpoint description and the relevant field.

apify-api/openapi/components/schemas/actor-tasks/Task.yaml
```diff
@@ -55,3 +55,18 @@ properties:
   standbyUrl:
     type: [string, "null"]
     format: uri
+  isPublic:
+    type: boolean
+    description: |
```
szaganek, line 60: Fluffy :D I'd go with:

`Whether the task is published. Based on the 'publicConfig.publishedAt' field.`

The rest of the description seems to belong to the endpoint description, not the field.

apify-api/openapi/components/schemas/actor-tasks/TaskPublicConfig.yaml
```diff
@@ -0,0 +1,42 @@
...
+  Public-facing configuration of a published task, used by the task's public landing page.
+  The task's publication state is determined by `publishedAt` - a task is published when
+  `publishedAt` is set and unpublished when it is `null`.
+properties:
+  publishedAt:
+    type: [string, "null"]
+    format: date-time
+    readOnly: true
+    description: |
+      Time when the task was published, or `null` if the task is not published.
```
szaganek:
```suggestion
      Time when the task was published, or `null` if the task isn't published.
```

apify-api/openapi/components/schemas/actor-tasks/TaskPublicConfig.yaml
```diff
@@ -0,0 +1,42 @@
...
+      This field is server-controlled - to publish or unpublish a task, set `isPublic`
```
szaganek: Shorter sentences improve readability, I'd go with:
`This field is server-controlled. To publish or unpublish a task, use the [Update task](/api/v2/actor-task-put) endpoint and set 'isPublic'.`

apify-api/openapi/components/schemas/actor-tasks/TaskPublicConfig.yaml
```diff
@@ -0,0 +1,42 @@
...
+      via the [Update task](/api/v2/actor-task-put) endpoint.
+    examples: ["2025-06-16T09:20:45.777Z"]
+  seoTitle:
+    type: [string, "null"]
+    description: SEO title of the public task page. Defaults to the task title when not set.
```
szaganek:
```suggestion
    description: Name of the Actor task to display by search engines such as Google. Defaults to the task title.
```
Worth adding character limit or similar things.

apify-api/openapi/components/schemas/actor-tasks/TaskPublicConfig.yaml
```diff
@@ -0,0 +1,42 @@
...
+    examples: [Scrape data from a website]
+  seoDescription:
+    type: [string, "null"]
+    description: SEO description of the public task page. Defaults to the task description when not set.
```
szaganek:
```suggestion
    description: Description of the Actor task to display by search engines such as Google. Defaults to the task description.
```

apify-api/openapi/components/schemas/actor-tasks/TaskPublicConfig.yaml
```diff
@@ -0,0 +1,42 @@
...
+  categorization:
+    type: [string, "null"]
+    description: Use-case category of the public task.
```
szaganek:
```suggestion
    description: Category that best defines the use case of the task.
```

apify-api/openapi/components/schemas/actor-tasks/TaskPublicConfig.yaml
```diff
@@ -0,0 +1,42 @@
...
+  inputSchemaFields:
+    type: [array, "null"]
+    description: Names of the task input fields displayed on the public task page.
+    items:
+      type: string
+  datasetName:
+    type: [string, "null"]
+    description: |
+      Name of the dataset from the Actor's dataset schema whose results are displayed. When
+      `null`, the Actor's default dataset is used - the only dataset it declares, or the one
```
szaganek:
```suggestion
      `null`, the Actor's default dataset is used. That is, the only dataset the Actor declares, or the one
```

apify-api/openapi/components/schemas/actor-tasks/TaskPublicConfig.yaml
```diff
@@ -0,0 +1,42 @@
...
+      named `default` when it declares several.
+  datasetView:
+    type: [string, "null"]
+    description: |
+      Key of the dataset view from the Actor's dataset schema used to display results. Must be
+      one of the views declared on the resolved dataset; a task cannot be published without it.
```
szaganek:
```suggestion
      one of the views declared on the resolved dataset. You can't publish a task without it.
```

apify-api/openapi/components/schemas/actor-tasks/UpdateTaskRequest.yaml
```diff
@@ -18,3 +18,22 @@ properties:
     anyOf:
       - $ref: ../actors/ActorStandby.yaml
       - type: "null"
+  publicConfig:
+    description: |
+      Public-facing display configuration of the task's public landing page. The provided
+      fields are merged into the stored configuration and validated. The stored configuration
+      cannot be cleared this way. Set `isPublic` to change the publication state.
```
szaganek: How to clear the stored configuration then?

fnesveda (changes requested):
I see some of the things I've raised in the previous review are still here, and still different from Actors:

> - Actors have a `categories` array, tasks have `categorization` string
> - Actors have all the details (`seoTitle` etc.) on the top level, tasks have it in `publicConfig`
> 
> Also, property names like `inputSchemaFields`, `datasetName` and `datasetView` somehow don't capture that they're being used only when showing the task, but they sound like they're relevant to the whole task (e.g. that they limit what input fields can the task process). How about putting them under some `displayOptions` object?

Have you discussed them already, and if so, what was the verdict? The response shape is also a part of the public API, so we need to have it finalized before merging this.

Janjiran:
- `categorization` is not used anymore, thanks for noticing - I'm removing it from the docs.
- `inputSchemaFields`, `datasetName` and `datasetView` - not it `displayOptions `, but they are inside `publicConfig` which should signalize the users that they only important for the public landing page. Same goes for the `seoTitle` and description - they are currently only used for the public landing pages whereas the Actors use it for the console store

fnesveda (approved):
Alright then 👍

apify-api/openapi/components/schemas/actor-tasks/UpdateTaskRequest.yaml
Janjiran: By sending null to the fields, I've added that to the text

szaganek (approved):

PR #2841, opened 2026-08-03
docs(openapi): Add missing descriptions to Get Actor endpoint

szaganek:
- Adds missing descriptions.
- Cleans up the intro topic.

Part of #2684.

### DIFF
--- a/apify-api/openapi/components/schemas/actors/Actor.yaml
+++ b/apify-api/openapi/components/schemas/actors/Actor.yaml
@@ -114,12 +114,7 @@ properties:
       URL for sending requests to the Actor in Standby mode.
       `null` if the Standby mode isn't enabled.
   notice:
-    type: [string, "null"]
-    enum: [NONE, RESIDENTIAL_PROXY_REQUIRED, UNDER_MAINTENANCE]
-    default: NONE
-    examples: [NONE]
-    description: A warning displayed on the Actor's page in Apify Store and Console.
-      Can be set by the Actor's developer or automatically by Apify's quality checks.
+    $ref: ./ActorNotice.yaml
   categories:
     type: array
     items:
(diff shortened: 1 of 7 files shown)


apify-api/openapi/components/schemas/actors/Actor.yaml
```diff
@@ -15,42 +15,51 @@ properties:
...
     type: boolean
     examples: [false]
+    description: Whether the Actor is available to users in Apify Store.
   actorPermissionLevel:
     $ref: ./ActorPermissionLevel.yaml
   createdAt:
     type: string
     format: date-time
     examples: ["2019-07-08T11:27:57.401Z"]
+    description: The date and time the Actor was created.
```
mixalturek: Nitpick: Do we want to mention that the format is ISO 8601 with milliseconds precision, or is it obvious from the example? The same for `modifiedAt`.

apify-api/openapi/components/schemas/actors/Actor.yaml
```diff
@@ -80,36 +92,54 @@ properties:
...
   categories:
     type: array
     items:
       type: string
     examples: [["DEVELOPER_TOOLS", "OPEN_SOURCE"]]
+    description: A list of categories that best define the Actor. Reflected in Apify Store's search and filtering options.
   isCritical:
     type: boolean
     examples: [false]
+    description: Whether the Actor is maintained by Apify.
```
mixalturek, line 127: Nitpick: Is this correct? The description doesn't match the field name for me. Critical for what? ;-)

mixalturek (approved):
Thanks a lot!

apify-api/openapi/components/schemas/actors/Actor.yaml
szaganek: Yeah, the name threw me off as well 😅 But the explanation stands (for example, [helpText](https://github.com/apify/apify-core/blob/develop/src/packages/intl/src/en/en.actor.ts#L340))

mixalturek (approved):

PR #2842, opened 2026-08-03
fix: disable the default blog plugin

barjin:
The classic preset enables the blog plugin by default, producing an empty `/blog` page whose markdown link points to a nonexistent `/blog.md`. The link checker fails on that 404.


TC-MO (approved):
LGTM

PR #2846, opened 2026-08-04
docs: Unify navigation under Actors category

### LINKED ISSUE #2802: Split Publication and Monetization into two categories
It's way too crowded. 

Move duplicated content from Academy: https://docs.apify.com/academy/actor-marketing-playbook/store-basics/how-actor-monetization-works

szaganek:
- Splits navigation under `Actors` into four steps.
- Unifies titles: Running, Developing, Publishing, Monetizing.
- Keeps the old slugs.

Closes #2802.


sources/platform/actors/development/index.md
```diff
@@ -1,8 +1,8 @@
 ---
-title: Actor development
+title: Developing Actors
```
TC-MO: I'd rather we not use gerunds in titles/headings/sidebar labels.

TC-MO (commented):
I like the approach, only thing that bother me is gerunds in titles/sidebar labels. I think it would be better to stick with nouns. (also actively remove any other gerunds that we may have leftover)

TC-MO (approved):
(it was supposed to be approve, not comment) but still worth discussing the gerunds :D

sources/platform/actors/development/index.md
szaganek: `Running Actors` is a challenge then 😅 How about bare infinitives, they are more action-oriented.

TC-MO: As long as we don't use gerund I think it will be fine. Maybe we should also codify it in contributing? I thought it already was but that is not the case

PR #2849, opened 2026-08-04
feat(academy): lesson about tests driven development

honzajavorek:
This PR adds a new lesson to the AI-first course. This lesson was forged in summer Prague at a temperature of 4000°C. The manufacturing process involved regular cooling in a dark basement office, resulting in a unique material structure.

honest-work


sources/academy/platform/scraping_with_apify_and_ai/04_tests_driven_prompting.md
```diff
@@ -1,33 +1,159 @@
...
-Adding fixtures, expectations. Setting up tests and teaching the agent to run tests. Dealing with corner cases by pointing the agent to the fixtures.
+**In this lesson, we'll keep developing our app for tracking prices on an e-commerce website. We'll describe edge cases with real-world examples, and Cursor will use them not only to get things right, but also to check that nothing breaks next time we change something.**
 
-In lesson 3, students would try to make changes via ChatGPT and see that it gets tedious, which leads to introducing an agent-based IDE to work inside the template more comfortably.
--->
+---
+
+The README as a source of truth for the AI agent gets us far, but it has limits:
+
+- Describing a large set of edge cases is tedious. "If this tiny detail is a certain way, process it as X, otherwise Y" for each situation is possible, but messy.
```
szaganek, line 14:
```suggestion
- Describing a large set of edge cases is tedious. Adding "If this tiny detail is a certain way, process it as X, otherwise as Y" for each situation is possible, but messy.
```

sources/academy/platform/scraping_with_apify_and_ai/04_tests_driven_prompting.md
```diff
@@ -1,33 +1,159 @@
...
+- Sometimes the edge case lies in the page's HTML, the text format that describes its content and structure. We'd have to say, "if you encounter exactly this HTML code, process it like this." Pasting long snippets of HTML into a README isn't great.
+- After each change, we have to trust that the agent didn't break what already worked. We can prompt it to "go through the whole README and verify all the behavior," but that's slow and unreliable.
```
szaganek, line 16:
```suggestion
- After each change, we have to trust that the agent didn't break what already worked. We can prompt it to go through the whole README and verify the behavior, but that's slow and unreliable.
```

sources/academy/platform/scraping_with_apify_and_ai/04_tests_driven_prompting.md
```diff
@@ -1,33 +1,159 @@
...
+_Red-green test-driven development_, often shortened to TDD, means that whenever the AI agent modifies our project, it starts with the expectations. It runs the tests and watches them fail, which proves that they actually check something. Only then does it change the code to make them pass. This technique makes development more reliable.
+
+Now let's send this prompt to the AI agent:
+
+```text
+Read the Testing section in README and set up a test suite
+covering the behavior we already have.
+```
+
+When the AI agent gets to work, you should see it create a new directory called `tests`. It'll probably use `curl`, a command-line program for downloading files, to create an HTML snapshot of the Sales listing.
```
szaganek, line 49:
```suggestion
When the AI agent gets to work, you should see it create a new directory called `tests`. To create an HTML snapshot of the Sales listing, it'll probably use `curl`, a command-line program for downloading files.
```

sources/academy/platform/scraping_with_apify_and_ai/04_tests_driven_prompting.md
```diff
@@ -1,33 +1,159 @@
...
+ℹ skipped 0
+ℹ todo 0
+ℹ duration_ms 165.677875
+```
+
+Depending on what the agent created, your output may look different, but it should follow a similar pattern. You don't need to understand every line. The important parts are `pass 6` and `fail 0`, which confirm that all tests passed.
+
+## Handling product variants
+
+Some prices in our data are "from" values because many items in the listing represent several product variants. Let's scrape each variant as a separate product with its actual price.
```
szaganek, line 77: I'm reading the whole thing in the morning on purpose to catch things that make my sleepy brain confused. It had to read "from values" three times. Can we make the explanation more... hm, wordy?

szaganek (commented):
Huge respect for producing all that during the heatwave 🥲 Just a few suggestions from me. We also should use curly quotes for things other then the code, but I didn't mark them.

sources/academy/platform/scraping_with_apify_and_ai/04_tests_driven_prompting.md
honzajavorek: If I can draw any conclusions from the years of my schooling, any course lessons should be optimized for sleepy brains – with utmost priority. I'm on it.

honzajavorek:
@szaganek Thanks for the review! I checked the repository guidelines and couldn't find a requirement for curly quotes in prose. The standards use straight quotes themselves, and `Microsoft.Quotes` is disabled and only checks punctuation placement. I'm fine using curly quotes, I even like them, but we should document the convention and its exceptions somewhere.

sources/academy/platform/scraping_with_apify_and_ai/04_tests_driven_prompting.md
honzajavorek: Hopefully fixed in [e7ef2a6](https://github.com/apify/apify-docs/pull/2849/commits/e7ef2a674269c39b3eb4286b2d574ceca9b0cc6f)

szaganek (approved):

PR #2850, opened 2026-08-05
docs: fix Apify MCP server page and its generated .md output

### LINKED ISSUE #2805: Rebuild MCP documentation from structured runtime and client data
> [!NOTE]
> This issue was AI-generated based on an AI analysis of Apify’s agent-facing contracts.

## Current problem

The generated [machine-readable MCP documentation](https://docs.apify.com/integrations/mcp.md) repeats authentication and client sections, misplaces expanded tab content, introduces account requirements before anonymous discovery, and retains an expired SSE migration notice. It also describes tool behavior that no longer matches the [released MCP server](https://github.com/apify/apify-mcp-server/blob/master/CHANGELOG.md). An agent reading the Markdown can follow a longer or incorrect setup path even though the human page appears usable.

## Why this reduces adoption and usage

- The repeated and misplaced content consumes model context before an agent reaches a working configuration.
- Stale tool and transport guidance causes failed setup attempts and unnecessary recovery steps.
- Contradictory anonymous-versus-authenticated ordering increases the largest existing Configurator funnel drop before users copy a configuration.
- Agents are less likely to treat Apify documentation as authoritative when the machine-readable version differs from the human page.

## Desired state

The HTML and machine-readable MCP pages are anonymous-first, OAuth-first for execution, structurally valid, and generated from current runtime/client metadata.

## Acceptance criteria

- [ ] Order the journey: anonymous discovery, hosted OAuth, client-specific setup, bearer fallback, local stdio, advanced selection.
- [ ] Remove the expired SSE notice and retired tool behavior.
- [ ] Define shared authentication prose once and generate only client-specific deltas inside tabs.
- [ ] Generate tool names/defaults from the released MCP package.
- [ ] Snapshot-test HTML and `.md` output for headings, tab expansion, lists, fences, and duplicate sections.

Priority: P0. Estimated effort: M.

birosrichard:
closes: https://github.com/apify/apify-docs/issues/2805
Fixes the MCP server integration page's generated `.md` output: nested Tabs inside a TabItem broke the llms-txt plugin's flattening, duplicating the auth explanation five times and misattaching tab content to the wrong client. Also corrects tool names, defaults, and the SSE transport note against the released `@apify/actors-mcp-server` package, and reorders the page anonymous-first.

Also updates `manus.md`, which repeated the same stale `get-actor-output` claim in its default-tools list and example walkthrough. Out of the issue's original scope, but the same fix applies.

### DIFF
--- a/sources/platform/integrations/ai/manus.md
+++ b/sources/platform/integrations/ai/manus.md
@@ -80,11 +80,11 @@ Try asking Manus something like:
 
 > "Search for 'best project management tools' on Google and summarize the top 10 results"
 
-Manus will call `search-actors` to find [Google Search Scraper](https://apify.com/apify/google-search-scraper), use `call-actor` to run it, and then `get-actor-output` to retrieve and summarize the results.
+Manus will call `search-actors` to find [Google Search Scraper](https://apify.com/apify/google-search-scraper), use `call-actor` to run it, and then `get-dataset-items` to retrieve and summarize the results.
 
 ## Configure tools
 
-After connecting, the Apify MCP server exposes a default set of tools for Actor discovery (`search-actors`, `fetch-actor-details`, `call-actor`, `get-actor-output`), web browsing (`apify/rag-web-browser`), and documentation search (`search-apify-docs`, `fetch-apify-docs`). See the [full tool reference](/integrations/mcp#available-tools) for the complete list.
+After connecting, the Apify MCP server exposes a default set of tools for Actor discovery (`search-actors`, `fetch-actor-details`, `call-actor`) and documentation search (`search-apify-docs`, `fetch-apify-docs`), plus the run and storage tools that are auto-injected alongside `call-actor`. See the [full tool reference](/integrations/mcp#available-tools) for the complete list.
 
 To control which tools are available, append a `tools=` query parameter to the server URL:
 
--- a/sources/platform/integrations/ai/manus.md
+++ b/sources/platform/integrations/ai/manus.md
@@ -84,7 +84,7 @@ Manus will call `search-actors` to find [Google Search Scraper](https://apify.co
 
 ## Configure tools
 
-After connecting, the Apify MCP server exposes a default set of tools for Actor discovery (`search-actors`, `fetch-actor-details`, `call-actor`) and documentation search (`search-apify-docs`, `fetch-apify-docs`), plus the run and storage tools that are auto-injected alongside `call-actor`. See the [full tool reference](/integrations/mcp#available-tools) for the complete list.
+After connecting, the Apify MCP server exposes a default set of tools for Actor discovery (`search-actors`, `fetch-actor-details`, `call-actor`), web browsing (`apify/rag-web-browser`), and documentation search (`search-apify-docs`, `fetch-apify-docs`), plus the run and storage tools that are auto-injected alongside `call-actor`. See the [full tool reference](/integrations/mcp#available-tools) for the complete list.
 
 To control which tools are available, append a `tools=` query parameter to the server URL:
 
--- a/sources/platform/integrations/ai/mcp.md
+++ b/sources/platform/integrations/ai/mcp.md
@@ -6,8 +6,6 @@ slug: /integrations/mcp
 toc_max_heading_level: 4
 ---
 
-<!-- markdownlint-disable MD024 -->
-
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 import ThirdPartyDisclaimer from '@site/sources/_partials/_third-party-integration.mdx';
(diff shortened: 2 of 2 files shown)


marekh19 (changes requested):
👨 Marek
 
I've let my clanker do the review, see below. The writing style notes are more on the docs team decision I would say. 

I've validated the point with removing `apify--rag-web-browser` myself and it is really included if you don't select any specific actors in `tools`.

This is what I got by default:

```
abort-actor-run
apify--rag-web-browser
call-actor
fetch-actor-details
fetch-apify-docs
get-actor-run
get-dataset-items
get-key-value-store-record
report-problem
search-actors
search-apify-docs
```

So this content change might be misleading.

Also checked the published package:

`package/dist/utils/tools_loader.js:108-109` - when no tools selectors are supplied at all:
```js
// No selectors supplied: use defaults unless actors were explicitly empty
return actorsExplicitlyEmpty ? [] : defaults.actors;   // = ['apify/rag-web-browser']
```

and `tools_loader.js:204-212` pushes the default categories, then `report-problem`, then those Actor tools:

```js
for (const cat of toolCategoriesEnabledByDefault) {   // = ['actors', 'docs']
    result.push(...categories[cat]);
}
result.push(reportProblem);
...
if (actorTools.length > 0) result.push(...actorTools);   // ← rag-web-browser lands here
```

> [!note]
> `defaults.actors` applies only when you don't pass any selectors. `?tools=actors,doc`s gives you no RAG browser. Which is the "replaces, doesn't add" behavior the PR already documents. So the correct fix isn't just re-adding the tool to the list; it's saying the default set includes it and that naming `tools` at all drops it.

I think it's worth it to address these points.

---

🤖 Clanker

### Writing style

1. **Concept used before it's explained** — `content-standards.md` "Information ordering" (line 67: "no concept is used before it's explained"). `mcp.md:60` opens the first post-intro section with "when the `tools` query parameter contains only tools enabled for unauthenticated use", and `mcp.md:69` hands the reader `?tools=search-actors,...`. The `tools` parameter isn't explained until `## Tool selection` (`mcp.md:245`), and the load-bearing gotcha — "The `tools` parameter does not add to the default selection - it replaces it" (`mcp.md:271`) — sits 200 lines further on. Move the `tools` explanation above `### Anonymous discovery`, or link forward from line 60.

2. **Passive voice** — `writing-style.md` "Active & inclusive voice"; `quality-standards.md:35`. `mcp.md:89` "Two methods are available."; `mcp.md:267` "Tools are grouped into five categories"; `mcp.md:275` "Tools marked _auto-injected_ are added automatically".

### Judgement calls

- **Mysterious Name** — `mcp.md:277` renames the column to `Loaded by default`, then fills it with three values: ✅, blank, and `Auto-injected`. A yes/no header now carries a third state. Worse, `mcp.md:284` marks `get-actor-run` as `Auto-injected` while `mcp.md:275` names `get-actor-run` as one of the *triggers* for auto-injection — it injects itself. Split into two columns, or drop `get-actor-run` from the trigger list.
- **Mysterious Name** — `mcp.md:128` "The configuration below uses the OAuth form." "Form" reads as a web form. Say "uses OAuth".
- **Duplicated Code** — `mcp.md:50` "If your client doesn't support remote MCP servers, run the server locally over stdio instead" is restated verbatim at `mcp.md:225`. The "authenticate before running Actors" fact appears four times in 40 lines (`mcp.md:48, 81, 83, 87`). The anonymous URL appears twice back to back (`mcp.md:69` inline, `mcp.md:75` in JSON) — drop the inline one.
- **Shotgun Surgery** — the default-tool list is asserted in `mcp.md:247` and again in `manus.md:87`, forcing both to be edited here. A third copy sits untouched at `sources/platform/get-started/agent-onboarding.md:116`; it happens to still agree, but it will drift.
- **Residual `.md` flattening risk** — the nested `<Tabs>` are gone (good), but the outer `<Tabs>` at `mcp.md:130` wraps four unlabeled client panels. If the PR's goal is clean llms-txt output, confirm those four don't concatenate into one unmarked run.

### Spec

`mcp.md:247` now claims *"the MCP server loads the `actors` and `docs` tool categories, plus `report-problem`."* The released package disagrees. Its README (§Tools configuration) lists defaults as `actors`, `docs`, **and `apify/rag-web-browser`**, and `src/const.ts` confirms it: `export const defaults = { actors: [RAG_WEB_BROWSER] }`.

The PR compounds this by deleting the `apify--rag-web-browser` row from the tools table (old `mcp.md:331`, `✅` enabled by default). The README still marks that tool `✅`. So the PR removed a correct row and replaced a correct default list with an incomplete one — the exact class of error AC4 exists to prevent. It also leaves the very next line, the example `?tools=actors,docs,apify/rag-web-browser`, looking like an arbitrary illustration when it is in fact the default set written out.

Same root cause in `manus.md:87`: the new text drops web browsing from the default set entirely.

birosrichard:
You were right about `apify/rag-web-browser`; it's missing from the defaults list and the table. Fixed both in `mcp.md`, and `manus.md` had the same problem, so I fixed that too.

Also added the point you raised: re-adding the tool to the list isn't enough. It's not part of any category, so naming `tools=` at all (even `?tools=actors,docs`) drops it unless you list `apify/rag-web-browser` explicitly. That's in the caution block now.

Tested this against the live server both ways, anonymous discovery via curl and an authenticated MCP connection, and both tool lists match what the docs say now, including report-problem getting withheld on this client.

Left a few of the style suggestions alone: the duplicate-URL pattern and the header split are consistent with the rest of the page, not actual bugs, and the "unlabeled tabs" flattening concern doesn't apply; every `TabItem` here already has a label.

marekh19 (approved):

PR #2852, opened 2026-08-05
docs: add sunset note to Flowise integration page

marcel-rbro:
The Flowise team has [discontinued the project](https://flowiseai.com/sunset). The code stays open source and existing deployments keep working, so the walkthrough stays for now - this just flags the status and links to their announcement, which carries the timeline. The page can be removed once usage dies out.

### DIFF
--- a/sources/platform/integrations/ai/flowise.md
+++ b/sources/platform/integrations/ai/flowise.md
@@ -11,6 +11,12 @@ import ThirdPartyDisclaimer from '@site/sources/_partials/_third-party-integrati
 
 <ThirdPartyDisclaimer />
 
+:::caution Flowise is sunset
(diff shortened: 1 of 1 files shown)


TC-MO (commented):
Shouldn't this be absolutely upfront even before explanation of what Flowise is? I'd rather know that project is semi-dead before I start to invest time. Furthermore with this setup we have admonition next to admonition which visually is not great

marcel-rbro:
I _assumed_ (👀) that anyone coming to this page would already know that since they'd probably be coming from a Flowise resource. Anyways, I've fixed it, now it's unmissable.

TC-MO (approved):

PR #2857, opened 2026-08-07
feat: factory ai docs

Jakub-Vacek:
This PR adds docs to Factory.ai AI plugin.


sources/platform/integrations/ai/factory.md
```diff
@@ -0,0 +1,147 @@
...
+
+import ThirdPartyDisclaimer from '@site/sources/_partials/_third-party-integration.mdx';
+
+[Factory](https://factory.ai) is an AI software engineering platform. Its CLI agent, Droid, runs in your terminal, reads and edits your codebase, runs commands, and completes multi-step development tasks from natural-language prompts.
+
+The [Apify plugin for Factory](https://github.com/apify/apify-factory-plugin) connects Droid to Apify's library of [Actors](https://apify.com/store) and bundles:
+
+- The [Apify MCP server](/integrations/mcp) for searching Apify Store, running Actors, and retrieving datasets through the [Model Context Protocol (MCP)](https://modelcontextprotocol.io/docs/getting-started/intro).
+- An `apify` routing droid that picks the right tool or skill from a natural-language request.
+- Five built-in skills for common workflows (see [Bundled skills](#bundled-skills) below).
```
marcel-rbro, line 16: The plugin also ships `/apify` and `/create-actor` slash commands - was leaving them out deliberate? `/apify` is shorthand for the `@apify` prefix you teach below so omitting it makes sense, but `/create-actor` is a separate guided workflow.

sources/platform/integrations/ai/factory.md
```diff
@@ -0,0 +1,147 @@
...
+
+This guide covers installation from the Factory plugin marketplace.
+
+<ThirdPartyDisclaimer />
+
+## Prerequisites
+
+- [An Apify account](https://console.apify.com/sign-up) - sign up for free if you don't have one.
+- [Factory](https://factory.ai) - the `droid` CLI installed and authenticated locally.
```
marcel-rbro, line 25: Three of the five bundled skills (`apify-actor-development`, `apify-actorization`, `apify-ultimate-scraper`) need the local `apify` CLI, which isn't listed here - should it be a prerequisite? `github-copilot.md` handles this with an "Authentication paths" section.

```suggestion
- [Factory](https://factory.ai) - the `droid` CLI installed and authenticated locally.
- [Apify CLI](https://docs.apify.com/cli/docs/installation) - needed by the Actor development, actorization, and ultimate scraper skills.
```

sources/platform/integrations/ai/factory.md
```diff
@@ -0,0 +1,147 @@
...
+
+Install the Apify plugin with the CLI:
+
+```bash
+droid plugin install apify@factory-plugins --scope user
+```
+
+Use `--scope project` instead to enable the plugin only in the current project.
+
+You can also install interactively. Run `/plugins`, open the **Browse** tab, select **Apify**, and choose an install scope.
```
marcel-rbro, line 43: I couldn't find an `apify` plugin in `Factory-AI/factory-plugins` - its `marketplace.json` lists eight plugins and none is Apify, so `apify@factory-plugins` looks like it won't resolve. Is a listing there pending, or should this point at Apify's own marketplace the way the plugin README does?

````suggestion
The Apify plugin is distributed from Apify's own Factory marketplace, so register it first:

```bash
droid plugin marketplace add https://github.com/apify/apify-factory-plugin
```

Install the Apify plugin with the CLI:

```bash
droid plugin install apify@apify --scope user
```

Use `--scope project` instead to enable the plugin only in the current project.

You can also install interactively. After you add the marketplace, run `/plugins`, open the **Browse** tab, select **Apify**, and choose an install scope.
````

File: sources/platform/integrations/ai/factory.md [hunk 4]
```diff
@@ -0,0 +1,147 @@
...
+
+[Factory](https://factory.ai) is an AI software engineering platform. Its CLI agent, Droid, runs in your terminal, reads and edits your codebase, runs commands, and completes multi-step development tasks from natural-language prompts.
+
+The [Apify plugin for Factory](https://github.com/apify/apify-factory-plugin) connects Droid to Apify's library of [Actors](https://apify.com/store) and bundles:
+
+- The [Apify MCP server](/integrations/mcp) for searching Apify Store, running Actors, and retrieving datasets through the [Model Context Protocol (MCP)](https://modelcontextprotocol.io/docs/getting-started/intro).
+- An `apify` routing droid that picks the right tool or skill from a natural-language request.
+- Five built-in skills for common workflows (see [Bundled skills](#bundled-skills) below).
+
+This guide covers installation from the Factory plugin marketplace.
```
marcel-rbro, line 18: Same question here - and adding the marketplace is always required rather than a fallback, since Droid never ships it.

```suggestion
This guide covers installation from the Apify plugin marketplace.
```

sources/platform/integrations/ai/factory.md
```diff
@@ -0,0 +1,147 @@
...
+
+```bash
+droid plugin marketplace add https://github.com/Factory-AI/factory-plugins
+```
+
+Then run `/plugins`, open the **Browse** tab, and search for **Apify**. If the plugin still doesn't appear, update the marketplace:
+
+```bash
+droid plugin marketplace update factory-plugins
+```
```
marcel-rbro, line 111: Same marketplace question here, and `marketplace update` takes the marketplace name rather than the repo name.

````suggestion
Re-register the marketplace and try again:

```bash
droid plugin marketplace add https://github.com/apify/apify-factory-plugin
```

Then run `/plugins`, open the **Browse** tab, and search for **Apify**. If the plugin still doesn't appear, update the marketplace:

```bash
droid plugin marketplace update apify
```
````

File: sources/platform/integrations/ai/factory.md [hunk 6]
```diff
@@ -0,0 +1,147 @@
...
+
+### The Apify MCP server shows as unauthenticated
+
+Run `/mcp`, select the **apify** server, and choose **Authenticate** to re-trigger the OAuth flow. See [Authenticate to Apify](#authenticate-to-apify).
+
+### Browser doesn't open, or OAuth fails
+
+If the browser doesn't open automatically, copy the OAuth URL shown in the terminal and paste it into your browser manually.
+
+If you're running Droid in a headless environment (SSH, remote container) or the OAuth flow still fails, authenticate with an API token instead. Copy your token from [Apify Console > Settings > Integrations](https://console.apify.com/settings/integrations) and set it before starting Droid:
```
marcel-rbro, line 125: Does `export APIFY_TOKEN` actually authenticate the MCP server here? The plugin's `mcp.json` is a plain HTTP entry with no token config, and the README's headless guidance is to authenticate locally first or switch to the CLI/SDK skills instead.

marcel-rbro (changes requested):
The install commands point at a marketplace that doesn't appear to contain the plugin - flagging that as a blocker, with the rest as questions.

sources/platform/integrations/ai/factory.md
Jakub-Vacek: Plan is to be in Factory marketplace, our plugin repo structure has it's own marketplace.json so suggested way also works. 

we can:
A) Go with factory marketplace - open PR in that marketplace and if it is not merged we can switch to our marketplace - optimistic approach
B) Go with Apify marketplace in docs, if submission is successful we can update the docs.

sources/platform/integrations/ai/factory.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Browser doesn't open, or OAuth fails
+
+If the browser doesn't open automatically, copy the OAuth URL shown in the terminal and paste it into your browser manually.
+
+If you're running Droid in a headless environment (SSH, remote container), the MCP OAuth flow can't complete. Authenticate locally first so the OAuth token is stored in your system keyring, then reconnect remotely. To keep working without MCP, use the CLI and SDK skills instead. Run `apify login`, or export an API token from [Apify Console > Settings > Integrations](https://console.apify.com/settings/integrations):
+
+```bash
+export APIFY_TOKEN=<YOUR_API_TOKEN>
+```
```
marcel-rbro, line 133: This paragraph gives two different fixes at once, and `apify login` is mentioned but never shown as a command. Splitting them into two options:

````suggestion
If you're running Droid in a headless environment (SSH, remote container), the MCP OAuth flow can't complete. You have two options:

- Keep using MCP: authenticate locally first so the OAuth token is stored in your system keyring, then reconnect remotely.
- Work without MCP: use the CLI and SDK skills instead, which authenticate through the Apify CLI:

  ```bash
  apify login
  ```

  In a non-interactive environment, export an API token from [Apify Console > Settings > Integrations](https://console.apify.com/settings/integrations) instead:

  ```bash
  export APIFY_TOKEN=<YOUR_API_TOKEN>
  ```
````

Jakub-Vacek: I aggre with this suggestion but I don't think that  ```export APIFY_TOKEN=<YOUR_API_TOKEN>``` replaces ```apify login```. Instead it should be  ```apify login -t <YOUR_API_TOKEN>````

PR #2858, opened 2026-08-10
docs: Delete outdated screenshot and fix UI copy

szaganek:
We introduced a new wizard, so the dialog no longer exists.


marcel-rbro (approved):
Looks good 👍

PR #2871, opened 2026-08-12
feat: vs code docs

Jakub-Vacek:
This PR adds docs for using Apify plugin directly in VS Code. VS Code can out of the box use Plugins for GitHub Copilot released in this repo: https://github.com/github/awesome-copilot/blob/main/plugins/external.json#L57 - this works now.


static/img/platform/integrations/vscode-white.svg
daveomri, line 1: Just a nit:
I think that blue logo works for both dark & light modes

We could drop this one and reuse blue on both versions

daveomri (commented):
One thing, otherwise LGTM 👍🏼

static/img/platform/integrations/vscode-white.svg
Jakub-Vacek: Yeah, I think so too. In VS Code brand guide they are pointing devs to use the white logo for any darker backgrounds so 🤷

marcel-rbro:
Will be resolved in #2889

Issue #2866, opened 2026-08-11
Pricing and billing topics


szaganek:
- https://help.apify.com/en/articles/13671754-navigating-apify-billing 🔄 
- https://help.apify.com/en/articles/11666209-applying-vat-exemption
- https://help.apify.com/en/articles/10774732-google-maps-scraper-is-going-to-pay-per-event-pricing
- https://help.apify.com/en/articles/10727112-how-to-apply-a-coupon ✅ 
- https://help.apify.com/en/articles/5136728-subscribing-to-the-apify-platform ✅ 
- https://help.apify.com/en/articles/7959177-subscription-plan-changes-and-proration 🔄
- https://help.apify.com/en/articles/3490384-what-is-a-compute-unit

PR #2882, opened 2026-08-18
docs: Add topic on applying promo codes

szaganek:
Part of #2866.

Might get merged into a bigger topic once I'm done rewriting billing. But for now, it can sit here.


TC-MO (approved):
Do hope it will end up in some larger topic later, feels wrong to have such small stub 🤔 OTOH it is pretty task-based ¯\_(ツ)_/¯ 

either way LGTM

PR #2886, opened 2026-08-18
docs: mark Google SERP proxy numPages parameter as deprecated

VojtaM39:
We have deprecated the `numPages` parameter on our Google SERP proxy. The only way to get further results is now to use the `start` parameter.

### DIFF
--- a/sources/platform/proxy/google_serp_proxy.md
+++ b/sources/platform/proxy/google_serp_proxy.md
@@ -60,35 +60,24 @@ See a [full list](https://ipfs.io/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXW
 
 ## Fetch more than one page of results
 
-By default, Google returns up to _10 search results per page_. To retrieve additional results beyond the first page, use the `numPages` query parameter.
-
-When you specify `numPages`, the proxy makes multiple requests in the background and merges them into a single HTML response.
-
-### `numPages` parameter
-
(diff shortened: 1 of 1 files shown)


bliuchak (approved):

daniil-poletaev (approved):
Looks good 👍

sources/platform/proxy/google_serp_proxy.md
```diff
@@ -60,36 +60,25 @@ See a [full list](https://ipfs.io/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXW
...
 
 ```text
-http://www.google.com/search?q=wikipedia&numPages=3
+http://www.google.com/search?q=wikipedia            # results 1-10
+http://www.google.com/search?q=wikipedia&start=10   # results 11-20
+http://www.google.com/search?q=wikipedia&start=20   # results 21-30
 ```
 
-This example fetches and combines the first 3 pages of results (up to 30 results total) into one HTML response.
+Each request is priced as a single request.
```
TC-MO: This is a bit circular, a request being prices as a request as well as it just restates LOC25

```suggestion
Each page you fetch counts as one request.
```

TC-MO (approved):
One suggestion, otherwise LGTM

sources/platform/proxy/google_serp_proxy.md
VojtaM39: 0bce1443f38a6d433b1ff194918d13874c762d07

PR #2890, opened 2026-08-18
docs: improve documentation titles and add proxy type links

jancurn:
Improved consistency and clarity across documentation by updating page titles to use gerund forms and adding internal links to related proxy documentation.

Key changes:
- Updated Actor section titles to use gerund forms ("Developing Actors", "Publishing Actors", "Monetizing Actors", "Running Actors") for consistency with documentation conventions
- Added internal links to proxy type documentation in the proxy parameters table (datacenter, residential, and Google SERP proxies) to improve navigation and discoverability

https://claude.ai/code/session_01WMamf9TfuBnQiKFwxXaLVM


TC-MO:
Hey @jancurn thanks for the update to links, valid change. Regarding the gerunds the PR description mentions that this is for consistency with documentation conventions but our convention is actually opposite. As you can see in [CONTRIBUTING.md](https://github.com/apify/apify-docs/blob/master/CONTRIBUTING.md#formatting-conventions), Quality standards for AI ([here](https://github.com/apify/apify-docs/blob/master/standards/writing-style.md#form) & [here](https://github.com/apify/apify-docs/blob/master/standards/quality-standards.md#formatting-and-structure)) & Msoft style guide that we also follow.

jancurn:
@TC-MO Oh right, I didn't know about the gerunds. Perhaps these rules should be copied in AGENTS.md too, so the agents pick them up?

The reason why I changed it is that the imperative form (Run, Develop, Publish, Monetize) is suitable for landing pages but doesn't feel right for docs:

Screenshot 2026-08-19 at 9 48 06

How about using "Actor running", "Actor development", "Actor publishing", "Actor monetization" instead of gerunds?

jancurn:
@TC-MO here is another PR to ensure agents respect CONTIBUTING.md - https://github.com/apify/apify-docs/pull/2893

TC-MO:
I'll take a separate look on #2893 to keep this on topic of PR :) 

Let's keep imperatives here. One of our goals is more task-based docs and these four sections are exactly that. The `Run` section is a landing page but also includes what is necessary to run Actors. Publish, Monetize & Develop work in the same way. 

On adding `Actor` to each of those, we are already in `Actors` section I see little if not none value in that addition. It would only serve to be more visually cluttered.

jancurn:
Do you have some examples of top docs using the imperatives? I don't recall seeing this anywhere, except for marketing pages. I didn't mean to add "Actor" to menu, but have it in the page title.

TC-MO:
Ah okay then, sorry for confusion! The larger pages that were recently unplifted like `Run` already include `Actors` in the main title but omit it from sidebar. We'll make sure that all pages work similarly.

as for examples there are a lot:

- Stripe: Build a Stripe-hosted checkout page. Across the payments docs you also get "Accept online payments", "Create a subscription", "Add a payment method", "Build a marketplace", "Fund and send payouts". Even the in-page headings are imperative, "Set up your server", "Create a Checkout Session", "Choose a mode", "Automate tax collection".
- Google Cloud: Create and start a Compute Engine instance, plus "Create an instance", "Manage VMs", "Stop or suspend a VM", "Rename a VM", "Update VM properties". Their style guide actually mandates it, task headings use the bare infinitive & they explicitly say avoid -ing as the first word because it translates inconsistently.
- Kubernetes: the whole Tasks section, "Install Tools", "Administer a Cluster", "Configure Pods and Containers", "Run Applications", "Run Jobs", "Access Applications in a Cluster", "Schedule GPUs".
- GitLab: their style guide requires active verb + noun, e.g. "Create an issue".
- Android & Twilio: "Build AI experiences", "Get started", "Optimize by form factor" and "Send your first SMS using code", "Send an RCS message".

TBF for example GitHub uses gerunds for their titles/headings, and also Red Hat I believe in their style guide goes with gerunds. But fact of the matter is that imperatives are also inline with user search pattern so it works better. (as in you search `how to run an Actor`, not `running an Actor`)

TC-MO (approved):
LGTM

TC-MO:
In the meantime I see conflict happened due to updates to Unblocker docs, I'll resolve it and merge the PR. @jancurn

PR #2895, opened 2026-08-19
docs: Actor titles

TC-MO:
Improve titles of the pages so OpenGraph renders with better context

### DIFF
--- a/sources/platform/actors/development/permissions/index.md
+++ b/sources/platform/actors/development/permissions/index.md
@@ -1,5 +1,5 @@
 ---
-title: Actor permissions
+title: Declare Actor permissions
 sidebar_label: Permissions
 description: Learn how to declare and manage permissions for your Actor, what access levels mean, and how to build secure, trusted Actors for Apify users.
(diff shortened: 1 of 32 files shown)


sources/platform/actors/development/permissions/index.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Actor permissions
+title: Declare Actor permissions
```
jancurn: I'd keep "Actor permissions" here it's more matching

jancurn (approved):
Thanks

sources/platform/actors/development/permissions/index.md
TC-MO: Yeah I was on the fence, it goes with our push for more task based documentation but for now I'll leave it as it was. Maybe next week split the docs into overview & a separate more task-based page.

jancurn: I have reservations to that push, let's discuss separately

PR #2896, opened 2026-08-19
docs: rename Unblocker proxy to Unblocker

TC-MO:
Drop the "proxy" qualifier, matching the Slack decision. Capitalization stays. Renames the page slug to /proxy/unblocker and adds a redirect from the old URL.

### DIFF
--- a/sources/platform/proxy/unblocker.md
+++ b/sources/platform/proxy/unblocker.md
@@ -8,7 +8,7 @@ slug: /proxy/unblocker
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-Unblocker is a proxy service designed for scraping websites protected by anti-bot and anti-captcha systems. Instead of building and maintaining your own detection and bypass logic, you can route requests through Unblocker, which automatically handles many common protection mechanisms.
+Unblocker is a proxy service designed to scrape any website - from pages without protections to those heavily protected by bot detection and CAPTCHA challenges. Instead of building and maintaining your own detection-bypass logic, you can route requests through Unblocker, which automatically handles many common protection mechanisms.
 
 This makes it a good choice for accessing websites with advanced bot protection while keeping your scraping code simple.
 
--- a/sources/platform/proxy/unblocker.md
+++ b/sources/platform/proxy/unblocker.md
@@ -1,27 +1,27 @@
 ---
-title: Unblocker proxy
+title: Unblocker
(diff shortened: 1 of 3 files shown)


sources/platform/proxy/unblocker.md
```diff
@@ -1,27 +1,27 @@
...
 sidebar_position: 10.5
-slug: /proxy/unblocker-proxy
+slug: /proxy/unblocker
 ---
 
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-Unblocker proxy is designed for scraping websites protected by anti-bot and anti-captcha systems. Instead of building and maintaining your own detection and bypass logic, you can route requests through Unblocker proxy, which automatically handles many common protection mechanisms.
+Unblocker is a proxy service designed for scraping websites protected by anti-bot and anti-captcha systems. Instead of building and maintaining your own detection and bypass logic, you can route requests through Unblocker, which automatically handles many common protection mechanisms.
```
daniil-poletaev: (nit): don't really like those anti-bot, anti-captcha systems especially along with word protected by. I know it was written by me, but since we're touching it maybe slight update? 😅 


```suggestion
Unblocker is designed to scrape any website - from simple pages to those heavily protected by bot detection and CAPTCHA challenges. Instead of building and maintaining your own detection-bypass logic, you can route requests through Unblocker, which automatically handles many common protection mechanisms.
```

daniil-poletaev (approved):
Looks good, thank you ❤️ 

Just one tech debt that was introduced by me 😄

sources/platform/proxy/unblocker.md
TC-MO: do we want to remove mention that this a a proxy service under the hood? This was I think the only place I left it deliberately so users can connect the dots on this (unless we don't want that?) . One nit from me I hate using the word `simple` in docs :D I'll propose maybe slight adjustment once I get confirmation about the removal of `proxy service` wording

daniil-poletaev: I don't have strong opinion about wording proxy service, so we can keep it ;)

TC-MO:
```suggestion
Unblocker is a proxy service designed to scrape any website - from pages without protections to those heavily protected by bot detection and CAPTCHA challenges. Instead of building and maintaining your own detection-bypass logic, you can route requests through Unblocker, which automatically handles many common protection mechanisms.
```

PR #2902, opened 2026-08-20
docs(openapi): Update Create Actor reference

szaganek:
- Deleted `pricingInfos` from the request body, because you can only set up monetization with the Update Actor endpoint.
- Updated the endpoint description, the older version made it sound as if you could create public Actors.
- Updated the response example to show what is actually returned when you create an Actor.
- Added a missing description to `apifyMarginPercentage` because my plan was to update the whole `pricingInfos`. But then it all went south. 

Part of #2684.

### DIFF
--- a/apify-api/openapi/components/schemas/actor-pricing-info/CommonActorPricingInfo.yaml
+++ b/apify-api/openapi/components/schemas/actor-pricing-info/CommonActorPricingInfo.yaml
@@ -10,8 +10,7 @@ properties:
     examples: [0.2]
     description: >
         Apify's share of the revenue generated under this pricing info record,
-        as a fraction between 0 and 1. Set by the Apify platform: 0.2 for pricing models that generate
-        revenue, 0 for pay-per-usage.
+        as a fraction between 0 and 1. Set by the Apify platform.
   createdAt:
     type: string
     format: date-time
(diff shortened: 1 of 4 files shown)


apify-api/openapi/components/schemas/actor-pricing-info/CommonActorPricingInfo.yaml
```diff
@@ -7,7 +7,11 @@ required:
 properties:
   apifyMarginPercentage:
     type: number
-    description: "In [0, 1], fraction of pricePerUnitUsd that goes to Apify"
+    examples: [0.2]
+    description: >
+        Apify's share of the revenue generated under this pricing info record,
+        as a fraction between 0 and 1. Set by the Apify platform: 0.2 for pricing models that generate
```
tobice: - Is the `0.2` margin correct under all circumstances?
- Will it stay correct? Does it make sense to hard-code it in the API documentation?

tobice (changes requested):

apify-api/openapi/components/schemas/actor-pricing-info/CommonActorPricingInfo.yaml
szaganek: Hmm, right, since it's always documented properly for each model anyway, it makes no sense to get into details here.

tobice (approved):

janbuchar (approved):
LGTM (as far as OpenAPI semantics are concerned)

PR #2904, opened 2026-08-21
docs: enforce no stacked headings in writing style guide

jancurn:
Adds a style guide rule that a heading must be followed by body text before the next heading (with a matching quality-checklist item) and brings the Run Actors page into compliance: an intro paragraph under the page title, the tutorial section retitled to "Run Actors manually in Apify Console" with a lead sentence, and a link on the Apify API mention. Also adds a brief "Run Actors from AI agents" section pointing to the MCP, API, and Agent Skills docs.

https://claude.ai/code/session_01PYBfJuQtU7wyP511g6k9vk

### DIFF
--- a/sources/platform/actors/running/index.md
+++ b/sources/platform/actors/running/index.md
@@ -9,13 +9,17 @@ slug: /actors/running
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-To run your first Apify Actor, we recommend trying one of the existing Actors from [Apify Store](https://apify.com/store). For details on building your own, see [Actor development](./development).
+You can run Actors manually in Apify Console, with the Apify API, or programmatically from your own applications. To run your first Apify Actor, we recommend trying one of the existing Actors from [Apify Store](https://apify.com/store). For details on building your own, see [Actor development](./development).
 
-## Prerequisites
+## Run Actors manually in Apify Console
+
+The easiest way to run an Actor is directly in [Apify Console](https://console.apify.com). The following tutorial takes you through your first run, from choosing an Actor to exporting its results.
+
+### Prerequisites
 
 To complete this tutorial, you need an Apify account. If you don't have it yet, [sign up for free](https://console.apify.com/sign-up).
 
-## 1. Choose your Actor
+### 1. Choose your Actor
 
 To find an Actor in Apify Store:
 
@@ -25,7 +29,7 @@ To find an Actor in Apify Store:
 
 For this tutorial, let's choose [Website Content Crawler](https://console.apify.com/actors/aYG0l9s7dbB7j3gbS/information/version-0/readme).
 
-## 2. Configure and run your Actor
+### 2. Configure and run your Actor
 
 Once you select the Actor, you will be taken to the Actor's detail page.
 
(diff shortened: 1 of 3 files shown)


TC-MO:
Hey @jancurn thanks for the rule, it's a valid addition. I'm less sure about the change to `actors/running` though.

Looking at the page, the only stacked heading there was the front matter title followed straight by `## Run your first Apify Actor`. The `H2 → H3` pair below it  had text in between, so it was compliant already. It's a one line fix, just an intro paragraph under the imports.

Flattening the H3s into H2s has a few side effects I'd rather avoid:

- `Prerequisites` becomes page scoped, so it now reads as prerequisites for the API and programmatic sections too, while the text still says "to   complete this tutorial"
- Steps 1-4 end up as peers of `Run Actors with the Apify API`, so in the ToC you get four numbered sections sitting next to two unnumbered ones with nothing marking where the tutorial ends
- The page lead is now the tutorial's opening line, on a page that covers three different ways to run an Actor

If you want I can revert the heading changes and just add the intro paragraph instead? Keeping your `[Apify API](/api)` link, that is a good catch.

One thing on the rule itself, I think "if you have nothing to say between two headings, the first heading is redundant" is what led here. A parent heading that groups its children isn't really redundant, grouping is what it's for.
Could we point at adding an orienting sentence or merging the sections instead of removing the heading?

jancurn:
@TC-MO Fair enough, just the title "Run your first Apify Actor" read bit weird and the section looked bad. I suggest we keep the title but change it to "Manually in Console", so that it's consistent with other sections. We should still add some text between the titles to match the new rule

jancurn:
Addressed both comments in b8970ea:

- The tutorial is grouped under its own H2 again, with its H3 subsections restored, so Prerequisites and steps 1-4 are scoped to it as before.
- Per Jan's suggestion the section is retitled by method — I went with "Run Actors manually in Apify Console" rather than the literal "Manually in Console", to stay parallel with "Run Actors with the Apify API" / "Run Actors programmatically" and because the terminology guide reserves bare "console". Easy to shorten if you prefer.
- Both stacked spots now have lead text: an intro paragraph under the page title listing the three ways to run an Actor, and an orienting sentence under the new H2.
- The style guide rule no longer calls a grouping heading redundant; it now says to add an orienting sentence, or merge sections only when there is genuinely nothing to say.

---
_Generated by [Claude Code](https://claude.ai/code)_

sources/platform/actors/running/index.md
```diff
@@ -9,9 +9,11 @@ slug: /actors/running
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-## Run your first Apify Actor
+You can run Actors manually in Apify Console, with the Apify API, or programmatically from your own applications. To run your first Apify Actor, we recommend trying one of the existing Actors from [Apify Store](https://apify.com/store). For details on building your own, see [Actor development](./development).
 
-To get started, we recommend trying one of the existing Actors from [Apify Store](https://apify.com/store). For details on building your own, see [Actor development](./development).
+## Run Actors manually in Apify Console
+
+The easiest way to run an Actor is directly in [Apify Console](https://console.apify.com). The following tutorial takes you through your first run, from choosing an Actor to exporting its results.
```
TC-MO: We really should avoid statements like "xyz is easiest/simplest", and link to Console is not necessary, since prerequisites section already has one for sign-up and then step 1 has Console link as well.

```suggestion
This tutorial covers a complete first run without any code, by choosing an Actor in [Apify Store](https://apify.com/store) and running it.
```

TC-MO (approved):
Just one nit otherwise LGTM. In a follow up PR I'll go over and deal with other stacked headings I noticed

PR #2910, opened 2026-08-24
ci: retry header assertion requests to survive upstream hiccups

barjin:
The header assertions proxy through to the live GitHub Pages sites of the sibling repos, so a transient upstream 5xx or timeout fails the whole job; these retries make it ride out the blip.

See CI error, e.g., [here](https://github.com/apify/apify-docs/actions/runs/32340620232/job/96338818326#step:10:215)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -97,7 +97,7 @@ jobs:
                         extra_args=("$@")
                         response=$(curl -s -D - -o "$body_sink" -w "\n%{http_code}" "${retry_opts[@]}" "${extra_args[@]}" "$url" 2>/dev/null)
                         status=$(echo "$response" | tail -1)
-                        location=$(echo "$response" | grep -i "^location:" | tr -d '\r' || true)
+                        location=$(echo "$response" | grep -i "^location:" | tr -d '\r' | tail -1 || true)
                         echo "→ $url → HTTP $status ${location:+(${location})}"
                         if [ "$status" = "301" ] || [ "$status" = "302" ]; then
                             echo "❌ Got redirect for $url: $location" && exit 1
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -121,7 +128,8 @@ jobs:
                     function assert_html_md_alternate() {
                         url=$1
                         expected_href=$2
-                        matched=$(curl -s "$url" | grep -oE '<link [^>]*type="?text/markdown[^>]*>' | grep -F "$expected_href" || true)
+                        curl -s -o "$body_sink" "${retry_opts[@]}" "$url"
```
B4nan: When retries are exhausted, this bare `curl` aborts the whole step under `set -euo pipefail`, and because of `-s` it prints nothing: no `→ $url` line, no ❌ message. Before this PR the failure was swallowed by the `|| true` pipeline and the function reported the failing URL. Curl also opens the output file lazily, so a failed fetch leaves the previous call's body in `$body_sink`; truncating it first makes sure we only ever grep the response we just fetched.

```suggestion
                        : > "$body_sink"
                        curl -s -o "$body_sink" "${retry_opts[@]}" "$url" || true
```

.github/workflows/test.yaml
```diff
@@ -88,7 +95,7 @@ jobs:
                         url=$1
                         shift
                         extra_args=("$@")
-                        response=$(curl -s -D - -o /dev/null -w "\n%{http_code}" "${extra_args[@]}" "$url" 2>/dev/null)
+                        response=$(curl -s -D - -o "$body_sink" -w "\n%{http_code}" "${retry_opts[@]}" "${extra_args[@]}" "$url" 2>/dev/null)
                         status=$(echo "$response" | tail -1)
                         location=$(echo "$response" | grep -i "^location:" | tr -d '\r' || true)
```
B4nan: This is missing the `tail -1` guard that `assert_header` got: a retried attempt carrying a `Location` header would print several lines into the log message. Cosmetic only, since the pass/fail check uses `$status`, which `-w` prints once per operation.

```suggestion
                        location=$(echo "$response" | grep -i "^location:" | tr -d '\r' | tail -1 || true)
```

B4nan (commented):
Two small follow-ups on the retry handling. Both affect diagnostics and log output only; the assertions themselves stay correct.

B4nan (approved):

PR #2916, opened 2026-08-25
feat: replace home breadcrumb icon with active navbar item label

jancurn:
Changed the first breadcrumb from displaying a home icon to showing the label of the currently active top navigation item. This provides better context about which section of the documentation the user is viewing, as there is no single home no more. The implementation now matches navbar items against the current pathname using their `activeBasePath` or `activeBaseRegex` patterns, with the most specific match winning when multiple items apply.

**From**

Screenshot 2026-08-25 at 14 10 39

**To**

Screenshot 2026-08-25 at 14 10 44

https://claude.ai/code/session_01DEB4iSj3QjDoVmCApoe9G6


TC-MO (approved):
LGTM, clicked through preview seems to be working fine

PR #2920, opened 2026-08-25
docs: add /add-plugin install path to Cursor page

lukas-bekr:
Cursor supports installing plugins from the chat with `/add-plugin apify`, but the Cursor integration page only documented the marketplace UI. This adds the slash command as the fast route and keeps the existing walkthrough as the alternative.


marcel-rbro (approved):
LGTM 👍

PR #2923, opened 2026-08-27
docs: Add a multiple proxy groups separator example

MatousMarik:
Added this from the edit page - sorry for the wrong commit name.


TC-MO (approved):
LGTM

PR #2924, opened 2026-08-27
docs: Update the quickstart with web IDE

szaganek:
The old version mixed the web IDE and the CLI. 

- Now it focuses solely on the web IDE. 
- Has screenshots that help understand the UI.
- Doesn't go into details. It's about opening Console and getting a working Actor as quickly as possible.

I'll add separate topics on hosting Actors in git repos.


TC-MO (approved):
LGTM

PR #2935, opened 2026-09-01
fix: hubspot redirect config

protoss70:
HubSpot requires the url to include capitalized name of `HubSpot` instead of `hubspot`, we need to add redirect in order to submit our integration.

More info in the image

image (8)

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -801,6 +801,8 @@ server {
 
     # Misc
     rewrite ^/platform/integrations/llama$ /platform/integrations/llama-index permanent;
+    # HubSpot marketplace publishing requires the capitalized path; redirect to canonical lowercase page
+    rewrite ^/integrations/HubSpot$ /integrations/hubspot permanent;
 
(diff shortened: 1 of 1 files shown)


nginx.conf
```diff
@@ -801,6 +801,8 @@ server {
 
     # Misc
     rewrite ^/platform/integrations/llama$ /platform/integrations/llama-index permanent;
+    # HubSpot marketplace publishing requires the capitalized path; redirect to canonical lowercase page
+    rewrite ^/integrations/HubSpot$ /integrations/hubspot permanent;
```
drobnikj: Suggestion: this exists to satisfy a pending marketplace submission whose requirements may still move, and a `permanent` 301 is cached hard by browsers. The file already uses `redirect` for exactly that situation ("302 while the IA settles; promote to permanent once stable") — worth following here.

marcel-rbro (approved):
LGTM 👍  can't test redirects on preview anyway, will check once live.

PR #2947, opened 2026-09-03
docs: Add info on consequences of increasing Actor price

szaganek:
Extends the section on monetization changes with some info and tips.


sources/platform/actors/monetizing/set-up-monetization.mdx
```diff
@@ -35,6 +35,10 @@ The monetization setup consists of three steps:
 
 To update the monetization of your Actor, follow the same steps as in the [Monetize your Actor](#monetize-your-actor) section.
 
+Be careful about increasing prices. If you make your Actor too expensive, users might find an alternative solution in Apify Store. Avoid setting high prices exclusively for users on a free plan. Such users often become paying users once they see the value of your Actor, but they need to test it on enough data first.
```
TC-MO, line 38: Seems like great candidate for an admonition no?

szaganek: If anything, I think only the sentence on negative profit makes sense as an admonition (tip).

TC-MO: That actually does work better

TC-MO (approved):
LGTM

PR #2954, opened 2026-09-07
docs: document slim Docker images and package manager setup

vladfrangu:
Mirrors apify/crawlee#4082 for the Actor Dockerfile page:

- Document the `-slim` tag variants (no preinstalled `apify`/`crawlee`/`typescript`) in the Node.js base images section and the tag naming convention, and recommend them by default.
- New *Package managers* section: Corepack enabled (no package manager besides npm preinstalled), `packageManager` support, flat `node_modules` linker config, `/pkg-cache`, and how to override the linker via `ENV`.
- Supported Node.js versions are now 22, 24 and 26. Example Dockerfiles on the Dockerfile, source code, performance and academy pages bumped from `:16`/`:20` to `:24`.

[🤖] Claude Code using Fable 5

### DIFF
--- a/sources/platform/actors/development/actor_definition/docker.md
+++ b/sources/platform/actors/development/actor_definition/docker.md
@@ -23,7 +23,15 @@ All Apify Docker images are pre-cached on Apify servers to speed up Actor builds
 
 ### Node.js base images
 
-These images come with Node.js (versions `22`, `24`, or `26`), the [Apify SDK for JavaScript](/sdk/js), and [Crawlee](https://crawlee.dev/) preinstalled. The `latest` tag corresponds to the latest LTS version of Node.js.
+These images come with Node.js (versions `22`, `24`, or `26`). The `latest` tag corresponds to the latest LTS version of Node.js.
+
+Every image is published in two flavours. The full image preinstalls the [Apify SDK for JavaScript](/sdk/js), [Crawlee](https://crawlee.dev/) and TypeScript. The `-slim` variant (e.g. `24-slim`, `24-1.60.0-slim`) only ships the browser automation library the image is built around (for example `puppeteer` or `playwright`), and `actor-node:24-slim` ships no npm packages at all. Slim images are smaller and faster to pull, and your `package.json` is the single source of truth for dependency versions.
+
+Use the slim variant unless you have a reason not to. Reach for the full image when you want to run something quickly without maintaining a `package.json`, or when you rely on the exact preinstalled versions of `apify` and `crawlee`.
+
+```dockerfile
+FROM apify/actor-node-playwright-chrome:24-1.60.0-slim
+```
 
 | Image | Description |
 | ----- | ----------- |
@@ -35,16 +43,6 @@ These images come with Node.js (versions `22`, `24`, or `26`), the [Apify SDK fo
 | [`actor-node-playwright`](https://hub.docker.com/r/apify/actor-node-playwright/) | Ubuntu image with [`playwright`](https://github.com/microsoft/playwright) and all its browsers (Chromium, Google Chrome, Firefox, WebKit). |
 
 See the [Docker image guide](/sdk/js/docs/guides/docker-images) for more details.
-
-#### Slim images
-
-Every Node.js image is also published as a slim variant, with a `-slim` suffix appended to the tag (e.g. `24-slim`, `24-1.60.0-slim`). Slim images do not preinstall `apify`, `crawlee` or `typescript`. They only ship the browser automation library the image is built around (for example `puppeteer` or `playwright`), and `actor-node:24-slim` ships no npm packages at all. This makes them smaller and faster to pull, and your `package.json` is the single source of truth for dependency versions.
-
-Use the slim variant unless you have a reason not to. Reach for the full image when you want to run something quickly without maintaining a `package.json`, or when you rely on the exact preinstalled versions of `apify` and `crawlee`.
-
-```dockerfile
-FROM apify/actor-node-playwright-chrome:24-1.60.0-slim
-```
 
 ### Python base images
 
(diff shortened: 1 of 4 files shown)


sources/platform/actors/development/actor_definition/docker.md
```diff
@@ -36,6 +36,16 @@ These images come with Node.js (versions `20`, `22`, or `24`) the [Apify SDK for
 
 See the [Docker image guide](/sdk/js/docs/guides/docker-images) for more details.
 
+#### Slim images
```
honzajavorek:
```suggestion
### Node.js slim images
```

Oftentimes, flat is better than nested. I wouldn't nest this section as H4, I think it's okay if we use just `###` here and a descriptive heading.

sources/platform/actors/development/actor_definition/docker.md
```diff
@@ -36,6 +36,16 @@ These images come with Node.js (versions `20`, `22`, or `24`) the [Apify SDK for
...
+
+Every Node.js image is also published as a slim variant, with a `-slim` suffix appended to the tag (e.g. `24-slim`, `24-1.60.0-slim`). Slim images do not preinstall `apify`, `crawlee` or `typescript`. They only ship the browser automation library the image is built around (for example `puppeteer` or `playwright`), and `actor-node:24-slim` ships no npm packages at all. This makes them smaller and faster to pull, and your `package.json` is the single source of truth for dependency versions.
+
+Use the slim variant unless you have a reason not to. Reach for the full image when you want to run something quickly without maintaining a `package.json`, or when you rely on the exact preinstalled versions of `apify` and `crawlee`.
```
honzajavorek: > Use the slim variant unless you have a reason not to.

This makes me think that this section should be listed first? So that anyone looking for Node.js images finds this one first and only when they have a reason not to use them, they scroll down to find out more about the base images. Does it make sense? 🤔 And in such case, `Reach for the full image` could become `Reach for the [full image](#anchor-to-the-base-images-section)`

sources/platform/actors/development/actor_definition/docker.md
```diff
@@ -141,6 +153,34 @@ The asterisk (`*`) tells npm to use whatever version is already installed, which
 1. Predictability - You know exactly which version you're running
 1. Debugging - Version-specific issues are easier to track down
 
+## Package managers
```
honzajavorek:
```suggestion
## Node.js package managers
```

I don't see this addressing any other managers than those specific to the Node.js ecosystem, so as a reader, I want to know before I start to read this that there isn't anything for me if I use pip or uv.

sources/platform/actors/development/actor_definition/docker.md
```diff
@@ -141,6 +153,34 @@ The asterisk (`*`) tells npm to use whatever version is already installed, which
...
+```
+
+The images preconfigure the package managers so that:
+
+- pnpm and yarn install a flat, npm-style `node_modules` (`node-linker=hoisted` for pnpm, `nodeLinker: node-modules` for yarn) instead of a symlinked store or Plug'n'Play, so dependencies resolve without extra loaders.
+- All caches (`NPM_CONFIG_CACHE`, `YARN_CACHE_FOLDER`, pnpm store and cache, `COREPACK_HOME`) live under `/pkg-cache` instead of `$HOME`. The directory only holds throwaway data, so you can `rm -rf /pkg-cache/*` at the end of your `Dockerfile` to reclaim space without touching installed dependencies.
+
+:::note Overriding the linker
+
+These settings are applied through environment variables (`PNPM_CONFIG_NODE_LINKER`, `YARN_NODE_LINKER`), and both pnpm and yarn give environment variables precedence over `.npmrc` / `.yarnrc.yml`. To use a different linker, override the variable in your `Dockerfile` instead of the config file:
```
honzajavorek: This reads in kind of confusing way. I'd streamline the sentences so that the message reads more "from A to B" and not jumping by random:

```md
To use a different linker, use environment variables, such as `PNPM_CONFIG_NODE_LINKER`
or `YARN_NODE_LINKER`. Both pnpm and yarn give environment variables precedence over
`.npmrc` or `.yarnrc.yml`, so you can override the variable in your `Dockerfile` instead
of in the config file:

... code block ...
```

honzajavorek (changes requested):
Thanks for visiting our docs monastery with the gift of new content! It looks good, but I dropped a few comments so that the resulting document is more coherent, at least according to my humble opinion.

I also feel that even if all my comments are addressed, the document as a whole kind of jumps from general advice to Node.js advice, then to general advice, then to Python advice, then back to Node.js advice, and so on. But I don't know if I'm to poke into that right now. Leaving it up to you, if you feel like the sections could use a better sorting, when you glance over their order.

sources/platform/actors/development/actor_definition/docker.md
vladfrangu: Folded into the intro of *Node.js base images* instead, so there is no separate heading at all. The two flavours and the recommendation now come before the table.

[🤖] Claude Code using Fable 5

sources/platform/actors/development/actor_definition/docker.md
vladfrangu: Done. The recommendation is now the first thing under *Node.js base images*, before the table. Since full and slim are two flavours of every image in the table, I described both in one paragraph rather than a separate section, so no anchor was needed.

[🤖] Claude Code using Fable 5

sources/platform/actors/development/actor_definition/docker.md
vladfrangu: Done.

[🤖] Claude Code using Fable 5

sources/platform/actors/development/actor_definition/docker.md
vladfrangu: Reworded, keeping one detail your draft dropped: the images already set the linker via these variables, which is why a config file alone silently does nothing. That is the reason the override has to go into the Dockerfile.

[🤖] Claude Code using Fable 5

vladfrangu:
Regrouping the page by scope is in #2964, stacked on this one, so it can be reviewed separately.

[🤖] Claude Code using Fable 5

honzajavorek (approved):
Awesome, thank you!

PR #2955, opened 2026-09-07
fix: use single adaptive svg for academy homepage illustrations

### LINKED ISSUE #2830: Use dynamic svg's for dark mode
@birosrichard Hi! The existing images for Academy courses, such as https://docs.apify.com/img/academy/api-scraping.svg, are crafted as adaptive to dark/light mode. If you open the image with light mode, it displays in light colors, if you open it with dark mode, it displays in dark colors. (You can try it yourself with the one I linked.)

You added a set of dark images, so now we have one adaptive set and one dark-only set. I think we should decide what's the convention after this change. Either we should strip the adaptive images to be light-only, so that their adaptive nature doesn't surprise someone who isn't prepared for it, or we should use the adaptive images for dark mode as well.

What do you think?

_Originally posted by @honzajavorek in https://github.com/apify/apify-docs/issues/2704#issuecomment-5092672082_

birosrichard:
Closes https://github.com/apify/apify-docs/issues/2830

Academy homepage cards had two conflicting dark-mode mechanisms: adaptive SVGs reacting to OS `prefers-color-scheme`, and a separate `imageUrlDarkTheme` swap to hand-drawn dark-only SVGs. This gave inconsistent dark-mode behavior depending on whether the OS theme matched the site's manual toggle.

Consolidates each illustration to a single adaptive SVG (`@media (prefers-color-scheme: dark)`, no swap), removes the now-redundant `*-dark.svg` files, and corrects colors/geometry to match the previously deployed dark assets (verified element-by-element against the deleted files).


honzajavorek (approved):
Thanks! 🚀

yeinn1000 (approved):
👍

PR #2984, opened 2026-09-16
feat: better component for prompts

### LINKED ISSUE #2979: Adjust the prompt component to show the prompt by default
The prompt component, as used in https://docs.apify.com/actors/development/quick-start/build-with-ai, always hides the prompt. As we're adding prompt examples throughout the docs and in courses, it would be nice if it showed the prompt by default and had an explicit flag for hiding the prompt in case it's long and we just want a prominent copy button. How it looks currently:

Image

When this is done, @marcel-rbro can change his prompts to use this component, and I can change all prompts in my AI course (to be published) to use this as well.

honzajavorek:
Closes https://github.com/apify/apify-docs/issues/2979

- Component now renders to llm.txt (bug of current PromptButton, does not render to llm.txt at all)
- Using blockquote instead of pre, so that llm.txt reads more nicely, also semantics, also word wrapping works better
- Can be collapsed or not
- Can be used as a code block language! Just write ```prompt and that's it!
- Of course, ```prompt title="Blah blah" is also supported to specify custom title
- Renamed from PromptButton to PromptBlock so that it better describes what it's good for

### Preview

To see it in action, check out preview of the following pages:

- https://pr-2984.preview.docs.apify.com/academy/actor-marketing-playbook/promote-your-actor/checklist#share-on-twitterx-with-a-demo
- https://pr-2984.preview.docs.apify.com/actors/development/quick-start/build-with-ai#ai-coding-assistant-instructions
- https://pr-2984.preview.docs.apify.com/academy/scraping-with-apify-and-ai/developing-scraper-with-ai-agent#scraping-prices


src/components/PromptBlock.jsx
```diff
@@ -48,28 +48,31 @@ export default function PromptButton({
...
-                <div className={styles['full-prompt-container']}>
-                    <div className={styles['full-prompt']}>
-                        <pre>{prompt}</pre>
-                    </div>
+            {/* Always rendered (not gated on showPrompt) so the prompt text is present in the
+                static HTML and gets picked up by the llms.txt generator; visibility for human
+                readers is controlled purely by CSS via the "visible" class below. */}
+            <div className={`${styles['full-prompt-container']} ${showPrompt ? styles.visible : ''}`}>
+                <div className={styles['full-prompt']}>
+                    <blockquote className={styles['full-prompt-blockquote']}>{prompt}</blockquote>
```
barjin, line 73: > Using `blockquote` instead of `pre`, so that `llm.txt` reads more nicely, also semantics

Can you please elaborate? It seems `<blockquote>` is made into a single-line Markdown`> quote`.

Image

Not that I expect the LLMs to care much, but it's not much nicer to read for humans. 

I'm not sure about the semantics either, W3C describes blockquote as _a section that is quoted from another source_ ([source](https://www.w3.org/TR/2011/WD-html-markup-20110525/blockquote.html)). `pre` imo fits here better.

src/components/PromptBlock.module.css
```diff
@@ -158,23 +170,35 @@ html[data-theme='dark'] .toggle-button:hover {
...
     border: 1px solid var(--ifm-color-emphasis-200);
-    border-radius: 8px;
-    box-shadow: 0 1.5px 3px 0 rgb(0 0 0 / 15%);
+    border-top: none;
+    border-radius: 0 0 8px 8px;
+    box-shadow: none;
 }
 
-.full-prompt pre {
+.full-prompt-blockquote.full-prompt-blockquote {
```
barjin, line 191: this feels hacky (battling CSS specificity?), maybe another reason why we should keep `pre`?

barjin (approved):
I've got a few talking points (⬇️ ), but generally lgtm. Thanks @honzajavorek !

src/components/PromptBlock.jsx
honzajavorek: My sentence is wrong and it conflates few things together. The readability boost should happen for a person reading the docs, because it now word-wraps, which, I think, is desired for prompts (unlike code).

I feel like blockquote is better semantics than code, because we quote some prose we invented or tried with LLMs, and it's not technically code. In the end, perhaps https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/samp would be the closest? 😅

But tbh, I don't have a strong opinion about this. I have a strong opinion that the text of the prompt should wrap for those who read it. But I don't care that much what HTML element implements it, if it's blockquote or pre+css.

Thanks for checking the llm.txt, I didn't notice that it's a _single line_ now, that's not really nice.

PR #2988, opened 2026-09-17
docs: repoint docs-tests baseline at moved account settings doc

marcel-rbro:
Closes #2951.

#2919 moved `account/settings.md` to `account/settings/index.md`. The baseline still pointed at the old path, so all 16 of its doc-side assertions failed and the check has been red since 09-07.

Path fix only, in `pages.json` and `assertions/account-settings.json`. Fifteen assertions needed the new path plus a +6 line shift; `settings-tab-side-menu` lost its justifying sentence in that rewrite, so it's re-quoted against the new intro. No Console UI claim changes.

Verified against staging: **64 passed · 0 failed · 12 skipped**. No genuine UI drift behind #2951 — the reported "32 failing" is 16 doc-side failures double-counted by the CI retry (#2911).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FSj1mTUbMKgmu2JoGJLSKD

Issue #2671, opened 2026-06-22
Docs-as-tests: automated Console UI drift detection


marcel-rbro:
## What

Stand up automated detection of documentation drift between the platform docs and the Apify Console UI. UI claims in the docs — a route resolves, a tab/button is named X, a heading is visible on page Y — are extracted as testable assertions and verified against Console staging, so when the UI changes and the docs don't, we find out automatically.

**Approach:** an adjustable list of docs pages → assertions extracted with an LLM → stored as a committed, human-reviewed baseline → evaluated against staging with Playwright. Failures point back to `source_file:line`.

## Status

Initial harness and a baseline for a 4-page Console test set (index, settings, billing, store — 53 assertions: routes, tabs, buttons, headings) are up in #2669.

## Remaining work

- [ ] Scale page coverage beyond the initial test set
- [ ] Handle detail-page assertions that require seeded fixtures (Actor / task / schedule / storage detail)
- [ ] Global left-nav group check (not a per-page claim)
- [ ] Session-gated pages (e.g. re-auth-required settings)
- [ ] Multi-step flow assertions (deferred — atomic claims first)
- [ ] Wire a scheduled GitHub Action (modeled on `lychee.yml`), with all secrets in GitHub Secrets and nothing in `.env`

## Source

Internal plan "AI-based testing for docs"; implementation tracked in #2669.

PR #2975, opened 2026-09-14
docs: storage tasks

TC-MO:
Sharing and cross-run usage were documented three times over, once per storage type, and had drifted apart; this consolidates them into two shared pages (Share storage, Use storage from another run) that the type pages now link to. Along the way it corrects the storage rate limits against `apify-core`.

PR #2989, opened 2026-09-17
docs: extend docs-tests coverage to storage and the rest of the account section

marcel-rbro:
Part of #2671. **Stacked on #2988** — retargets to `master` once that merges.

Widens the baseline from **3 pages / 38 assertions** to **11 / 87**: storage (overview, datasets, key-value stores, request queues) and the remaining account pages (2FA, promo codes, general resource access, organization usage). Also folds in the abandoned `docs/docs-tests-detail-fixtures` branch, giving 7 previously-skipped assertions a landing route.

Two harness bugs this surfaced:

- **Slug derivation.** `.mdx` kept its extension, and an `index.md` page produced `account-settings-index` instead of overwriting `account-settings` — so re-extracting a page that became a directory would write a second baseline file beside the first, and both would be evaluated. Now strips `.mdx`/`.md` and a trailing `/index`, using two plain substitutions since `\.mdx\?$` matches nothing under BSD sed (macOS).
- **Stale storage routes** in the extractor's route table. `/storage?tab=keyValueStores` returns 200 but redirects to the datasets tab; Console serves `/storage/key-value-stores`. Every re-extraction would have regenerated a wrong `at`.

The same stale URLs appear in the storage docs themselves, where they send readers to the wrong tab. That's a prose fix — separate PR.

Verified against staging: **139 passed · 0 failed · 35 skipped**, 0 issues. The skips are element assertions with no landing route (storage/Actor detail pages, connector modals), which need the seeded fixtures from #2671.

https://claude.ai/code/session_01FSj1mTUbMKgmu2JoGJLSKD


marcel-rbro:
Folded in `docs/docs-tests-detail-fixtures` (pushed 2026-07-28, never opened as a PR). That branch predates both the IA v3 path refresh and this expansion, so it conflicted with all three account baselines — the at-routes are reapplied here rather than rebased, and the branch can be deleted.

Gives a landing route to the dashboard sections (`/`), the billing **Pricing** and **Invoices** tabs, and the Settings **Referrals** tab. All 7 verified against staging, including the billing routes, which were never checked when the original branch was written.

Updated totals: **139 passed · 0 failed · 35 skipped** (was 132 / 42).

The remaining 35 skips need the seeded-user fixtures from #2671, not an at-route — storage and Actor detail pages, connector modals, and the 2FA enabled-state controls.

TC-MO:
Here I'd wait, I have a big storage rewrite #2975 in the works that is to be merged soon

PR #2990, opened 2026-09-17
chore: update codex installation flow

Jakub-Vacek:
This PR updates Codex App installation flow from official marketplace - it is much simpler then original flow :)


Jakub-Vacek:
@marcel-rbro let's wait with merge of this - not sure If this is needed after all (sorry about this)

PR #2992, opened 2026-09-17
docs: match UI labels exactly in docs-tests, and fix two casing drifts

marcel-rbro:
Part of #2671. **Stacked on #2989.**

The runner matched labels with `exact: false`, so an assertion passed whenever the documented label was merely *contained* in what Console renders. That hides the most common kind of drift — a casing change.

Switching the four locators to exact matching found two real ones that had been passing silently:

| Doc said | Console renders |
| --- | --- |
| **Plan Consumption** | **Plan consumption** |
| **account-level integrations** | **Account-level integrations** |

Both are bolded as UI element names, so house style says they should match the screen. Fixed in the docs and in the baseline targets. ("Graph" lowercased with the first, since it isn't part of the label.)

Exact matching is still whitespace-trimmed, so only wording, casing and trailing decoration can break a target — all things the docs should be updated to match. Of 52 UI assertions these two were the only breakages; the rest already matched character for character.

Verified against staging: **139 passed · 0 failed · 35 skipped**.

https://claude.ai/code/session_01FSj1mTUbMKgmu2JoGJLSKD


TC-MO (approved):
LGTM

PR #2995, opened 2026-09-17
docs: recapture Actor quality score screenshot

TC-MO:
The old screenshot exposed the internal admin navigation bar. 
New capture comes from a regular publisher account, sidebar collapsed, converted to WebP.


marcel-rbro (approved):
LGTM 👍

PR #3001, opened 2026-09-20
fix: prevent line breaks in table identifier columns

jancurn:
Prevent inline code in the first column of tables from wrapping at hyphens. Identifiers like tool names, environment variables, and field names now stay on a single line, while prose columns remain free to wrap naturally.

image

https://claude.ai/code/session_01VkicoRz9Mx8ZxRxB14w3ye


TC-MO (approved):
LGTM

PR #3005, opened 2026-09-21
fix: collapse doubled section prefix in docs .md cross-links

marekh19:
The `.md` twins and `llms*.txt` under `/sdk/{js,python}`, `/api/client/{js,python}` and `/cli` carry same-section links with the section prefix repeated (`/sdk/python/sdk/python/docs/...`), so roughly 6,700 links across those sections 404. The cause is `@signalwire/docusaurus-plugin-llms-txt@1.2.2`: with `relativePaths: false` it prepends the site URL to hrefs that already contain the `baseUrl`, which only shows up on sites served under a sub-path.

Two changes, both at our edge:

- A 301 in `nginx.conf` that collapses the doubled prefix, so every one of those links resolves in one hop. Verified end to end against the real GitHub Pages origins with doubled links taken from today's production content: 1 redirect, HTTP 200, correct `text/markdown`.
- The same substitution in `joinLlmsFiles.mjs`, so the root `llms-full.txt` is correct in the text itself, independently of nginx. Verified against the real production file: 6,664 doubled links to 0, no other content touched. `checkLlmsSize.mjs` now fails the build if any come back.

The proxied bodies from the five sub-sites keep the doubled URLs in their text. They resolve through the redirect, so nothing is broken, but they are not canonical until the sub-sites themselves stop generating them. An earlier revision of this PR rewrote those bodies with `sub_filter`; that is dropped, see the comment below for why.

The redirect is worth keeping even after the sub-sites are fixed, because the doubled URLs stay in search indexes and agent caches. It only collapses a repeated same-section prefix; the plugin also prefixes cross-section links (`/sdk/js/cli/docs/...`), 5 of them in the root `llms-full.txt`, which this cannot detect and which the upstream fix does not address either.

Closes apify/apify-web#6665


marekh19:
For context I've checked #2477 and that was a one-off docs patch. We can solve this one with patch as well but it means doing that patch in multiple repos, not only here.

marekh19:
## Approaches considered

**Rewrite the proxied bodies with `sub_filter` (the earlier revision of this PR).** Makes the served Markdown canonical, but `sub_filter` cannot read a compressed body, so each location has to send `Accept-Encoding: ""` upstream and the response reaches the client uncompressed. Measured: `/sdk/js/llms-full.txt` 149,735 bytes gzipped today against 663,680 uncompressed, and `/api/client/python/llms-full.txt` 153,113 against 1,312,936, so 4.4x to 8.6x on exactly the files agents fetch.

**`sub_filter` plus re-compression in nginx.** `gzip on; gzip_types text/markdown;` fixes the `.md` routes. It does not fix the 10 `llms*.txt` routes: they call `proxy_hide_header Content-Type`, and the gzip filter decides by reading that header, so it sees nothing and skips the response. `default_type` does not help either. It needs a loopback stage per section, which works (181,931 bytes gzipped, correct type, 0 doubled links) but adds an internal path and an extra hop across five sections to save one redirect.

**Patch the plugin in the five sub-repos.** All five use pnpm, so `pnpm patch` needs no new dependency and no `postinstall`. But it pins an exact version, so a Renovate bump breaks install in five repos, the patch is hand-written against compiled output in five places, and it only fixes the same-section case. Cross-section links stay broken: `stripBaseUrl` cannot strip a prefix the path never had.

**Set `relativePaths: true` in the five sub-repos.** One line each, it is the plugin default, and it fixes both link classes because nothing is prepended at all. The cost is root-relative instead of absolute URLs in those sections, which resolve fine on `docs.apify.com` but lose the origin if a file is copied elsewhere.

**Wait for upstream 2.x.** Not a plan on its own: npm `latest` is still 1.2.2 and nothing has shipped since `2.0.0-alpha.7` in November 2025.

## Why this one

The redirect is needed whichever route we take, because links already in search indexes and agent caches will point at doubled URLs for a long time. It is 10 lines in one repo, it ships without coordinating five others, it costs no bandwidth, and it makes all ~6,700 links resolve immediately. Everything else on that list is either more nginx to own for a cosmetic gain, or a cross-repo change that this redirect does not block.

The follow-up is `relativePaths: true` in the five sub-repos, which fixes the cause and also covers the cross-section links. Both `apify-sdk-js` and `apify-cli` deploy docs on push to `master`, so each takes effect the same day it merges. Worth doing one repo first and checking its generated `llms-full.txt` before touching the rest. Once all five have redeployed, the substitution in `joinLlmsFiles.mjs` can go and the redirect stays as a plain legacy-URL redirect.

webrdaniel (approved):
nice workaround

scripts/checkLlmsSize.mjs
```diff
@@ -27,6 +27,20 @@ if (llmsChars === null || llmsFullChars === null) {
     process.exit(1);
 }
 
+// Guards the substitution in joinLlmsFiles.mjs. The sub-sites generate links with their
+// section prefix repeated; if the substitution stops matching, the joined files silently
+// fill up with URLs that only resolve through the redirect in nginx.conf.
```
B4nan: Both scripts use the same pattern, so if the substitution stops matching, this guard stops matching too. What it really catches is someone removing the substitution (or `sanitizeFile` not running) while a sub-site still ships doubled links.

```suggestion
// Guards the substitution in joinLlmsFiles.mjs: if it gets removed while a sub-site still
// generates links with its section prefix repeated, the joined files silently fill up
// with URLs that only resolve through the redirect in nginx.conf.
```

scripts/checkLlmsSize.mjs
```diff
@@ -27,6 +27,20 @@ if (llmsChars === null || llmsFullChars === null) {
...
+const DOUBLED_SECTION_PREFIX =
+    /https:\/\/docs\.apify\.com\/(sdk\/js|sdk\/python|api\/client\/js|api\/client\/python|cli)\/\1\//g;
+
+for (const filePath of [llmsPath, llmsFullPath]) {
+    const matches = (await fs.readFile(filePath, 'utf8')).match(DOUBLED_SECTION_PREFIX);
```
B4nan: nit: `checkFile` already reads both files and only returns `.length`. Returning the content would save the second read here. Build time only, so fine either way.

B4nan (approved):

